PR #64, opened 2021-05-11
Add Audio to the Catalog

obulat:
This is the first attempt to adding audio to the Catalog.

Currently, this PR extracts generic features from the `ImageStore` class to a generic `MediaStore` class. This is the parent class for the new `AudioStore` class.

Further steps would be:
- [ ] run tests to see that `ImageStore` still works
- [ ] add an audio ingestion dag using either [Jamendo API](https://developer.jamendo.com/v3.0/tracks) #51 or [Freesound API](https://freesound.org/docs/api/resources_apiv2.html#sound-list-response) #16
- [ ] analyze audio fields available and open a discussion as to which fields to add


obulat:
Closing this in favor of #69

PR #66, opened 2021-05-17
[Quality] Make provider scripts pass validated license_info to the storage module

obulat:
The `licenses.py` module (`src/cc_catalog_airflow/dags/provider_api_scripts/common/licenses/licenses.py`) contains functions for deriving valid license details from URL or license name and license version, and validating license URLs`src/cc_catalog_airflow/dags/provider_api_scripts/common/licenses/licenses.py`.

Currently, the provider scripts extract the **license URL** and/or **license name and license version**, and pass these data to the `storage/image.py`. This module, in turn, handles license details validation using additional `licenses.py`, then converts the image details to a tsv row, and if the tsv row is missing columns, no image is stored. The return from `image.add_item` is the number of images stored. 
 
 This PR moves the workflow for license validation to the provider scripts. This short-circuits the license validation, and  ensures that `image_store.add_item` is only called for images with valid license details.


obulat:
Validating licenses in the Provider API scripts should improve the performance because we would skip several steps for items with invalid licenses.
One thing I learned while looking through the project history is that the ImageStore module validates licenses for the Common Crawl ingestion. I'm converting this PR to a draft PR until I can test this change with the Common Crawl.

zackkrida:
Have you had a chance to re test this @obulat? I'd be curious for @mathemancer to take a peek too—some historical context could be useful here.

mathemancer:
The context is that we wanted to make Provider API scripts as easy as possible for outside contributors, and reserved the `ImageStore` class for the more experienced.  Containing everything common behind the one easy-to-find `add_item` method avoids some needed learning for a new contributor who may have knowledge of the 3rd party API, but not of the Openverse Catalog codebase.  This had the logistical side-benefit of making review easier and quicker.  There were fewer things to check for a Provider API script submission, since the `add_item` method pretty much guaranteed that no malformed data would end up in the database (or clogging up downstream parts of the pipeline).  I.e., if you move it up into each script, you'll have to check every future script submission to make sure they're properly calling the license-verification logic and then the "fixed" licenses are what's actually being passed down to the `add_item` method.

W.r.t. performance, the bottleneck is getting things into the DB with further downstream widgets, not pulling them from APIs (with a couple of exceptions:  Flickr and WMC).

obulat:
> The context is that we wanted to make Provider API scripts as easy as possible for outside contributors, and reserved working on the `ImageStore` class for the more experienced. Containing everything common behind the one easy-to-find `add_item` method avoids some needed learning for a new contributor who may have knowledge of the 3rd party API, but not of the Openverse Catalog codebase. This had the logistical side-benefit of making review easier and quicker. There were fewer things to check for a Provider API script submission, since the `add_item` method pretty much guaranteed that no malformed data would end up in the database (or clogging up downstream parts of the pipeline). I.e., if you move it up into each script, you'll have to check every future script submission to make sure they're properly calling the license-verification logic and then the "fixed" licenses are what's actually being passed down to the `add_item` method.
> 
> W.r.t. performance, the bottleneck is getting things into the DB with further downstream widgets, not pulling them from APIs (with a couple of exceptions: Flickr and WMC).

Thank you very much for your reply, @mathemancer  ! It is really helpful to learn the context.

I have read that provider API scripts were determined to be the best point of entrance for new community contributors in the docs, and I really love this idea!

On the other hand, `add_item` method does require the license information. If the API provides the license URL, great! But if it only gives the name and version, then  the  contributor has to have some inside knowledge about licenses. For example, if you give the license name as upper-cased 'BY', or add 'cc-by' (both are my own mistakes), the `add_item` method will have an invalid license, and will not save the item. In this case, the new contributor might not even understand that a valid item has been discarded due to the script mistake.  We could solve the problem of having to know license names by adding license name constants and documenting them better.

Moreover, after reading this comment, I thought about having a template file for API scripts, which would have main methods filled in, #TODOs for what needs to be done, and any documentation that might be necessary for contributor to create a new script. There would actually be two templates: one for the script, and one for the test file :)

It's great to know about the performance, as I haven't really had a chance to gauge real prod performance.

mathemancer:
As you mention it, I love the template idea.  It might also be that modifying relevant methods to give more useful feedback.  I remember specifically _not_ making them more forgiving, since our assumption (especially around licensing) was that avoiding incorrect data was more important than making sure every bit of possible data was saved.  It's possible that's a goal worth reconsidering in the new context though.

obulat:
>  I remember specifically _not_ making them more forgiving, since our assumption (especially around licensing) was that avoiding incorrect data was more important than making sure every bit of possible data was saved. It's possible that's a goal worth reconsidering in the new context though.

It's an interesting idea to discuss on our next Tuesday developer chat on [Openverse Slack ](https://wordpress.slack.com/messages/openverse/)

obulat:
Following our discussion here, I created a template for provider API scripts in #93 .

This PR now changes all the existing provider API scripts to validate the license information using the `get_license_info` method with one of the two possible parameter sets: license_url AND/OR license name and license version. The `ImageStore` class checks that the license is valid by checking if `license_info.license` is not None in `add_item` and `_get_item` methods. Normally, checking in the public `add_item` should be enough, but I also added the check in the `_get_item` because in tests we also use the private `_get_item` method directly. We could adjust the tests and remove the check from the private method.

zackkrida:
@obulat Revisiting this—I'm excited to see it merged along with #93 
Will the audio provider API scripts need the same adjustments here?

obulat:
> @obulat Revisiting this—I'm excited to see it merged along with #93
> Will the audio provider API scripts need the same adjustments here?

The audio provider API scripts (1 item) are fully compatible with these changes!

zackkrida (approved):
flickr run manually + tests are working for me

PR #67, opened 2021-05-17
[Quality] Code cleanup

obulat:
This PR fixes several types of code problems highlighted by Pycharm's code inspection. 

### Changes:
- Replace mutable default parameters
Python creates a default parameter once when the function is defined. The same value is used every time the function is called. *This causes a problem* if the value is mutable, such as a dictionary or a list. If this value is ever changed inside the function, the changed value is used for subsequent function calls, instead of the default value.

- Make function names lower case, separating words with underscores: `sanitize_string` instead of `sanitizeString`
- Refactor to remove variables that might be referenced before assignment
- Remove unused variables and imports
- Add some type hinting


src/cc_catalog_airflow/dags/provider_api_scripts/cleveland_museum_of_art.py
```diff
@@ -92,6 +94,7 @@ def _get_response(
 def _handle_response(
                     batch
                     ):
```
krysal, line 96: The format of this param looks a bit odd here.

krysal (approved):
These are a ton of improvements! Thanks for taking care of this.

I left comments with some minor questions but everything seems fine to move on. 🙌

src/cc_catalog_airflow/dags/provider_api_scripts/cleveland_museum_of_art.py
obulat: Definitely! Thank you for catching this!

src/cc_catalog_airflow/dags/provider_api_scripts/finnish_museums.py
```diff
@@ -137,9 +140,9 @@ def _get_raw_tags(obj):
 
 def _get_landing(obj, landing_url=LANDING_URL):
     l_url = None
-    id = obj.get("id")
-    if id:
-        l_url = landing_url + id
+    id_ = obj.get("id")
+    if id_:
+        l_url = landing_url + id_
```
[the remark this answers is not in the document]
obulat: Both [`id`](https://docs.python.org/3/library/functions.html#id) and [`license` ](https://docs.python.org/3/library/constants.html#license)are reserved words in Python: id is a built-in function, and license is a built-in constant. So it's a good practice not to use them as variable names. 

PEP8 suggests that to avoid name collision when you want to use the built in name for your variable, you can use a single trailing underscore (`id` -> `id_`):

https://pep8.org/#descriptive-naming-styles
> single_trailing_underscore_: used by convention to avoid conflicts with Python keyword, e.g.:
> 
> Tkinter.Toplevel(master, class_='ClassName')

Thank you for asking, I learned a bit more about it while researching the answer :)

obulat:
Apparently, the code problems fixed in this PR **do** cause errors when running the scripts using Airflow dags, like this:

```
[2021-05-29 05:05:05,083] {requester.py:44} INFO - Received response from https://api.finna.fi/api/v1/search?filter%5B%5D=format%3A%220%2FImage%2F%22&filter%5B%5D=building%3A%220%2FMuseovirasto%2F%22&limit=100&page=6
[2021-05-29 05:05:05,088] {taskinstance.py:1482} ERROR - Task failed with exception
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/airflow/models/taskinstance.py", line 1138, in _run_raw_task
    self._prepare_and_execute_task_with_callbacks(context, task)
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/airflow/models/taskinstance.py", line 1311, in _prepare_and_execute_task_with_callbacks
    result = self._execute_task(context, task_copy)
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/airflow/models/taskinstance.py", line 1341, in _execute_task
    result = task_copy.execute(context=context)
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/airflow/operators/python.py", line 117, in execute
    return_value = self.execute_callable()
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/airflow/operators/python.py", line 128, in execute_callable
    return self.python_callable(*self.op_args, **self.op_kwargs)
  File "/usr/local/airflow/dags/provider_api_scripts/finnish_museums.py", line 39, in main
    _get_object_list(building)
  File "/usr/local/airflow/dags/provider_api_scripts/finnish_museums.py", line 59, in _get_object_list
    total_images = _process_object_list(object_list)
  File "/usr/local/airflow/dags/provider_api_scripts/finnish_museums.py", line 96, in _process_object_list
    total_images = _process_object(obj)
  File "/usr/local/airflow/dags/provider_api_scripts/finnish_museums.py", line 117, in _process_object
    license_url=license_url,
UnboundLocalError: local variable 'license_url' referenced before assignment
```

So, it is important to fix them.

obulat:
I'm closing this PR because it was split into 2: #100 and #103 to make it more manageable.

PR #72, opened 2021-05-25
Add ingestion column to MediaStore when using provider API

### LINKED ISSUE #87: [Feature] Add ingestion_type column when saving API data
## Current Situation

The Catalog Database has `ingestion_type` column: if we got the item data from Common Crawl ingestion, the value is "CommonCrawl", and if we used provider APIs to get the data, the value is "Provider API". Currently, the API ingestion process writes the tsv file, and then in a separate step, checks if the tsv file has the `ingestion_type` column by checking the number of columns. If the number is fewer than 17, `ingestion_type` column is added to the tsv file. 
This was used to add the column to the files that were written before we had the `ingestion_type` column.

## Suggested Improvement

We should add the `ingestion_type` column when we write the tsv file, in the ImageStore/AudioStore `add_item` method. The test for `ingestion_type` column should still be done in order to fix existing files, if need be. 

## Benefit

This change will simplify the API ingestion process, and lower the chance for errors due to the hard-coded number of columns that we compare to.

## Alternatives

We could just leave it as is, or add the `ingestion_type` column to the CommonCrawl scripts as well, and remove the check for `ingestion_type` entirely. However, it should be done when we can test CommonCrawl processes, and at the moment, we can't.

obulat:
Fixes #87
The database media tables (image and audio) have ingestion_type column with values of 'commoncrawl' if the data was ingested using the common crawl pipeline, and 'provider_api' if the data was ingested through the provider API scripts.

Currently, it is being added after the tsv file has been created: a tsv line is checked for the correct number of columns. If the number is one less than necessary, the 'ingestion_type' column is added. This is a remnant of the previous process, and was supposed to be used only for the old data that did not have 'ingestion_type'. 

This PR makes the provider API scripts add the 'ingestion_type' column with 'provider_api' value by default. The functionality of checking the tsv for the correct column number is left in place because it might be necessary for commoncrawl ingestion/ older data.


src/cc_catalog_airflow/dags/provider_api_scripts/common/storage/audio.py
```diff
@@ -40,6 +40,7 @@
     columns.JSONColumn(name="tags", required=False),
     columns.StringColumn(name="provider", required=False, size=80, truncate=False),
     columns.StringColumn(name="source", required=False, size=80, truncate=False),
+    columns.StringColumn(name="ingestion_type", required=False, size=80, truncate=False)
```
zackkrida, line 43: 👍

src/cc_catalog_airflow/dags/provider_api_scripts/common/storage/image.py
```diff
@@ -61,7 +61,10 @@
     ),
     columns.StringColumn(
         name='source', required=False, size=80, truncate=False
-    )
+    ),
+    columns.StringColumn(
+        name="ingestion_type", required=False, size=80, truncate=False
```
zackkrida, line 66: 👍

src/cc_catalog_airflow/dags/util/loader/column_names.py
```diff
@@ -1,5 +1,5 @@
 """
-This file holds string constants for the column names in the image
+This file holds string constants for the column names in the media
```
zackkrida, line 2: Good.

src/cc_catalog_airflow/dags/util/loader/ingestion_column.py
```diff
@@ -22,14 +22,17 @@ def check_and_fix_tsv_file(tsv_file_name):
...
-        old_cols_number = len(IMAGE_TSV_COLUMNS)
+        old_cols_number = len(IMAGE_TSV_COLUMNS) - 1
     new_cols_number = old_cols_number + 1
     with open(tsv_file_name) as f:
         test_line = f.readline()
     line_list = [word.strip() for word in test_line.split('\t')]
     if len(line_list) == old_cols_number:
+        # Previously, the last column was 'source'
+        # If source is 'commoncrawl', we set the ingestion_type to
+        # 'commoncrawl', else to 'provider_api'
```
zackkrida, line 35: Nice to see the behavior documented

zackkrida (approved):
Nice addition, thank you!

zackkrida (approved):

zackkrida (approved):

PR #74, opened 2021-05-27
Audio storage

obulat:
This is a cleaned-up version of #64

This PR creates a new AudioStore class by extracting generic features from the ImageStore to a generic MediaStore class, and inheriting both ImageStore and AudioStore from it.

It also creates a new database for audio. The ingestion workflow now differs for audio and image tables, and the selection is based on the media_type, which is recorded in the tsv filename.

To finalize this PR, we need to agree on the audio metadata that we are going to save.


obulat:
Closing this in favor of #85

PR #76, opened 2021-05-31
Re-add the lint and test workflows

obulat:
This PR re-adds the workflows for linting with flake8 and testing on PR and push from the original repository. During the move of the repository, we disabled these workflows by moving them to `workflows-disabled` folders. The only change in the linting workflow is replacing python version (3.9 instead of 3.7).

The two remaining workflows in the `workflows-disabled` folder will need to be reviewed later.


zackkrida (approved):

.github/workflows/push_pull_request_test.yml
```diff
@@ -0,0 +1,16 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build the stack
+        run: |
+          cd ./src/cc_catalog_airflow
+          cp env.template .env
+          docker-compose up -d
+      - name: Test
+        run: |
+          sleep 10
```
dhruvkb, line 15: This might introduce some flakiness in the tests. A more resilient approach would probably loop on `sleep`, based on the `docker container ls` response.

.github/workflows/push_pull_request_test.yml
```diff
@@ -0,0 +1,16 @@
+name: Automated testing workflow
```
dhruvkb, line 1: Minor suggestion: adding newlines and between sections for readability (similar to the `lint.yml` file).

dhruvkb (approved):
Looks good to me! Always a fan of DX automations 👍

PR #81, opened 2021-06-03
Fix failing test, improve import readability

obulat:
Addition of the testing and linting workflows surfaced a couple of errors that were added in the last commit:

- The `test_raw_pixel.py` had some changes from a different PR, which do not pass. The changes were reverted, and will be added in a different PR.
- There was a duplicate import in `raw_pixel.py` script. I removed that, and added a blank line between the standard library and local package imports for readability.

### DIFF
--- a/src/cc_catalog_airflow/dags/provider_api_scripts/raw_pixel.py
+++ b/src/cc_catalog_airflow/dags/provider_api_scripts/raw_pixel.py
@@ -1,5 +1,3 @@
-from common.storage.image import ImageStore
-from common.requester import DelayedRequester
 import requests
 import logging
 from urllib.parse import urlparse, parse_qs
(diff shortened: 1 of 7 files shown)


src/cc_catalog_airflow/dags/provider_api_scripts/raw_pixel.py
```diff
@@ -1,5 +1,3 @@
-from common.storage.image import ImageStore
-from common.requester import DelayedRequester
```
obulat, line 2: These classes are imported below, so the duplicate imports are removed.

dhruvkb (approved):
Looks good to me!

zackkrida (approved):
The reversion of `test_raw_pixel.py` looks correct to me.

PR #83, opened 2021-06-03
Extract MediaStorage entity as parent to ImageStore

### LINKED ISSUE #82: [Feature] Create a parent MediaStorage entity with common metadata
## Problem Description

Currently, the catalog only has data for images. To add audio, we need to create the Storage class for audio. The first step for this is extracting common media metadata from ImageStore class to MediaStore class. This MediaStore class will be the base for other media types in the future.

## Solution Description

The solution would be to extract the common columns that would be used for any media type into an abstract base class, MediaStore. This class would handle data cleaning and saving the buffer to the disk. The `add_item` method for cleaning and storing the media items would be implemented in the child classes: ImageStore, AudioStore, etc, as they have some logic specific to the media type.

## Alternatives

An alternative could be to create a completely separate AudioStore class. This would be problematic as it would make us have duplicate code.

obulat:
Fixes #82 

This PR extracts the data and methods from the `ImageStore` class that are common to all media types, and creates an Abstract base class called `MediaStore`.

`MediaStore` has the methods for validating the tags, metadata, license information, the resulting TSV rows, and for writing the buffer to the disk. It also has an abstract method, `add_item`, which will be implemented by all child classes (`ImageStore` currently, but `AudioStore` will be added shortly).  This method will handle the validation of single item metadata.

This is the third iteration of adding audio :) This time, in parallel with the API, this PR only handles the `MediaStore` extraction, so it can be merged before we have a final decision on what Audio metadata we want to save.

All the data fields we currently collect for images from providers can be found in IMAGE_TSV_COLUMNS list. Here, they are listed in order they are written to TSV, and the fields that are common for all media are in bold:
- **foreign_identifier**
- **foreign_landing_url**
- **image_url / audio_url** (url in database)
- **thumbnail_url**
- width
- height
- **filesize**
- **license_**
- **license_version**
- **creator**
- **creator_url**
- **title**
- **meta_data**
- **tags**
- **provider**
- **source**
- **ingestion_type**

There are several ways this PR can be tested:
1. Build the docker containers and run all the tests on Docker:
`docker exec cc_catalog_airflow_webserver_1 /usr/local/airflow/.local/bin/pytest`

or

2. Choose one of the provider API scripts, preferably the ones that don't require authentication, such as Cleveland museum or Science museum, and run the script by itself. After it finishes running, you should see the result in a `.tsv` file inside your `/tmp` folder. Hopefully, all fields collected should be written in the `.tsv` file.

or

3. In Pycharm, I right-click on `src/cc_catalog_airflow/dags/common` folder, and then select `Run 'pytest in common'` to run the tests only for the common module that was changed in the PR.


zackkrida:
FYI, I'm hoping to review this tomorrow 😸

PR #85, opened 2021-06-03
Add AudioStorage entity

### LINKED ISSUE #84: [Feature] Add AudioStore entity as a child of MediaStore
## Problem Description

To ingest audio data, the catalog needs to have the `AudioStore` class that would validate the data and save it to the disk. 

## Solution Description

After extracting the common functionality for storing any media type into `MediaStore` class, we can add other types of media by creating a child class that implements `add_item` with cleaning and validation that is specific to audio type. This means that we will need to list columns for all the metadata that we will store for audios.

## Alternatives

We could add a separate `AudioStore` class that is not a child of `MediaStore`, but then we would lose the ability to reuse the common code, and would probably have a lot of code duplication.

## Additional context

This is a part of the [meta issue on adding audio](https://github.com/WordPress/openverse/issues/19)

obulat:
Fixes #84 

This PR adds the `AudioStore` class as a child of generic `MediaStore` class with columns and validation for metadata that is specific to audio type. This class is based on the `ImageStore` class.

This is the third attempt at creating the PR for audio storage in the catalog :)


src/cc_catalog_airflow/dags/common/storage/test_audio.py
```diff
@@ -0,0 +1,489 @@
...
+    'bit_rate': None,
+    'sample_rate': None,
+    'category': None,
+    'genre': None,
+    'audio_set': None,
+    'alt_audio_files': None,
+    'source': 'testing_source',
+    'ingestion_type': 'provider_api',
+
+}
```
zackkrida, line 46: Do you think it would be better to mock fields that are list/arrays as empty ones instead of `None`? I don't know `/shrug` so thought I'd ask anyway

zackkrida (approved):
Tests passed!

src/cc_catalog_airflow/dags/common/storage/test_audio.py
obulat: This will make the tests more robust, thank you!

PR #91, opened 2021-06-08
Fix the dependency version conflict

obulat:
The update of urllib3 in #86 has caused a dependency version mismatch. This is causing the CI build failures at now.

This PR updates the versions of boto3, botocore and ipython to fix the build.

The local_s3 Docker image was also updated from Python3.7 to the Python:3.9 version, too.


zackkrida (approved):

krysal (approved):
LGTM

PR #108, opened 2021-06-21
Remove logging of url rewriting when not rewritten

obulat:
`urls.py` module validates license URLs that we save to the database.

`rewrite_redirected_url` function makes sure that we record the final license URL after any redirects. Currently, when called for all license URLs, it logs `{url_string} was rewritten to {rewritten_url}` even if the URL did not redirect, and, therefore, did not need rewriting. This can be confusing and fills the logs unnecessarily. 

This PR only logs rewritten URL if it is different from the URL passed to the function.

### DIFF
--- a/src/cc_catalog_airflow/dags/common/urls.py
+++ b/src/cc_catalog_airflow/dags/common/urls.py
@@ -64,7 +64,8 @@ def rewrite_redirected_url(url_string):
         response = requests.get(url_string)
         if response.ok:
             rewritten_url = response.url
-            logger.info(f'{url_string} was rewritten to {rewritten_url}')
+            if rewritten_url != response.url:
+                logger.info(f'{url_string} was rewritten to {rewritten_url}')
         else:
             logger.warning(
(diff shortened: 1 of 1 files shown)


src/cc_catalog_airflow/dags/common/urls.py
```diff
@@ -64,7 +64,8 @@ def rewrite_redirected_url(url_string):
         response = requests.get(url_string)
         if response.ok:
             rewritten_url = response.url
-            logger.info(f'{url_string} was rewritten to {rewritten_url}')
+            if rewritten_url != response.url:
```
krysal, line 67: Perhaps the intention was to compare with the original `url_string`?
```suggestion
            if rewritten_url != url_string:
```

obulat: Thank you for catching this, @krysal !

krysal (approved):
Good!

zackkrida (approved):
Makes perfect sense, nice change.

PR #111, opened 2021-06-25
Add Audio to the database

obulat:
This is a cleaned-up version of #64 part related to saving the Audio data to the database. And #69 :)

This PR:

- creates a new Audio database. The `sql` scripts for creation of the Postgres tables were added to both [`local_postgres`](../tree/add_audio_db/src/cc_catalog_airflow/local_postgres) folder that is used for local testing, and the [`openledger_sql`](../tree/add_audio_db/src/openledger_sql) folder which is [probably] used in production. 

-  Reorders the AudioStore columns so that audio-specific columns are all at the end. This makes it easier to write and read the data, and to test. It would be nice to do the same for ImageStore, however we have many TSV files that use the columns in the current order, so we would need to detect the TSV file structure and process it accordingly.
- adds media type parameter to the loader functions so that they can select correct list of columns to write to the database.
- adds audio-specific SQL queries to `loader/sql.py` for creating audio tables, upserting and overwriting audio in the database.

Added on June 30:
- moves `watermarked` column from image-specific columns to a list of columns common for all media types, as per [API models](https://github.com/WordPress/openverse-api/blob/c19234fb641a277595f9bcfd04ad4ca4422feadc/openverse-api/catalog/api/models/media.py#L30), as suggested by @krysal .
- detects correct `ingestion_type` column value for all media based on `source` value.

There are three different places where data is added or overwritten in the database, and in each place the columns are listed separately because different columns are affected. I would like to have this process abstracted in the future, so that we don't have to update column lists in all three places.


src/cc_catalog_airflow/dags/util/loader/ingestion_column.py
```diff
@@ -6,7 +6,7 @@
 import logging
 import os
 
-from common.storage.audio import AUDIO_TSV_COLUMNS
+from common.storage.audio import AUDIO_TSV_COLUMNS Current Change
```
zackkrida: Looks like some git merge conflict issue

zackkrida (changes requested):
Small bug i think

zackkrida (approved):

zackkrida:
> I would like to have this process abstracted in the future, so that we don't have to update column lists in all three places.

👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍

dhruvkb (dismissed):
I don't understand it fully but from what I do, it seems like an extension of the existing code to accept a `media_type` and set the columns accordingly. If my understanding is correct, it looks good to me.

src/cc_catalog_airflow/dags/util/loader/sql.py
```diff
@@ -273,17 +337,15 @@ def _merge_jsonb_arrays(column: str) -> str:
...
           {_newest_non_null(col.THUMBNAIL)},
-          {_newest_non_null(col.WIDTH)},
-          {_newest_non_null(col.HEIGHT)},
           {_newest_non_null(col.FILESIZE)},
           {_newest_non_null(col.LICENSE)},
           {_newest_non_null(col.LICENSE_VERSION)},
           {_newest_non_null(col.CREATOR)},
           {_newest_non_null(col.CREATOR_URL)},
           {_newest_non_null(col.TITLE)},
-          {_newest_non_null(col.WATERMARKED)},
```
krysal, line 284: We left the `watermarked` field generalized in the [API models](https://github.com/WordPress/openverse-api/blob/c19234fb641a277595f9bcfd04ad4ca4422feadc/openverse-api/catalog/api/models/media.py#L30).

That is how I recently learned that audios can also be watermarked ℹ️

obulat: I didn't know that. Thank you for this comment! I've made 'watermarked' field a common field for all media types.

zackkrida (approved):
🎶 🎶 🎶

PR #123, opened 2021-07-07
Add audio db

obulat:
This code should already be merged into master by https://github.com/WordPress/openverse-catalog/pull/111.
I've had `audio_popularity_calculation` based on `add_audio_db`, which was already merged. However, for some reason, when I merged #111, the base of the latter didn't change?


obulat:
Closing this in favor of #124. The problem of base branch not changing after merge, and wrong diff calculation is really confusing.

PR #125, opened 2021-07-07
Add missing `watermarked` column to audio loading table

obulat:
Audio table did not have `watermarked` field at first. @krysal pointed out to me in a PR review that we have it in the API, and should have it in the catalog, as well. I added it to the `AudioStorage` and to loader sql scripts, but forgot to add it to the sql script that creates the intermediary loading table. This PR fixes this.

This PR is also another reason why we should abstract the database column creation and writing code so that we can add a column in one place, and it updates everywhere. Instead of having to add a column to more than three places.


zackkrida (approved):

krysal (approved):
Thanks for catching this!

PR #126, opened 2021-07-08
Delete duplicated CommonCrawl providers

### LINKED ISSUE #122: [Quality] Remove duplicate Common Crawl providers
## Current Situation

For Museum Victoria and Science Museum, we have both CommonCrawl ETL Provider module and a provider API script.  

## Suggested Improvement

We can delete the CommonCrawl provider modules: 
https://github.com/WordPress/openverse-catalog/blob/main/src/providers/commoncrawl/MuseumVictoria.py
https://github.com/WordPress/openverse-catalog/blob/main/src/providers/commoncrawl/ScienceMuseum.py

## Benefit

This will ensure that we have a single source of truth for these museums.

## Implementation

krysal:
Fixes #122 by @obulat.

This PR deletes old CommonCrawl ETL provider scripts for Museum Victoria and Science Museum, since we already have provider API scripts for these sources.


zackkrida (approved):

obulat (approved):
Thank you!

PR #129, opened 2021-07-12
Make wikimedia script pass license_info, not license_url

obulat:
In #66 we updated MediaStore class to accept an instance of 'LicenseInfo' class, which can derive all necessary license information, and validate it.

All Provider API scripts were updated to pass `license_info` to `MediaStore.add_item` method except for Wikimedia script. This PR replaces old `add_item` parameters(`license_url`) with the new `license_info`. It also updates the tests so that they would detect parameter mismatch in the future.


krysal (approved):
LGTM 👍

PR #113, opened 2021-06-25
[API integration] Add Jamendo provider API script

obulat:
Fixes #51

This PR adds the script to pull the audio data from Jamendo API. This is the first audio provider integration to Openverse.

There are two 'audio_url' fields returned by Jamendo API: one for streaming and one for downloading.
```
"audio": "https://mp3d.jamendo.com/?trackid=168&format=mp32&from=37t85X61BkFJQKB4JnaEcg%3D%3D%7CjxH6jKzRNIYQdqbHcm2mpg%3D%3D",
"audiodownload": "https://mp3d.jamendo.com/download/track/168/mp32/",
```
When the download for audio is allowed("audiodownload_allowed": true), the audiodownload field is saved as the audio url.

PR #130, opened 2021-07-20
Modify audio columns

krysal:
In line with WordPress/openverse-api#141, this PR:
- Add `watermarked` column to the schema of the audio table in SQL files
- Rename the `genre` column to `genres` and changes its type to a Postgres array field
- Rename the `alt_audio_files` column to `alt_files`
- Keep the old `standardized_popularity` column for `image` and `audio` tables.


src/cc_catalog_airflow/dags/common/storage/audio.py
```diff
@@ -73,7 +73,7 @@
         name='category', required=False, size=80, truncate=False,
     ),
     columns.JSONColumn(
```
dhruvkb, line 75: This should be a string array column. In PostgreSQL they are represented by `varchar(<len>)[]` and the data can be stringified as `{<item 1>,<item 2>...}`.

The current implementation of `genres` is `varchar(80)[]`.

dhruvkb (changes requested):
`view_count` is also a non-nullable field in media. The sample data for images (in `openverse-api`) contains a `view_count` column with all values set to zero, can this be reproduced for audio?

src/cc_catalog_airflow/dags/common/storage/audio.py
krysal: Working on this, it's needed to add the new [column][columns] type.

[columns]: https://github.com/WordPress/openverse-catalog/blob/2bdb6ec7b1cf44486c03dd139fb7f61a3a32e465/src/cc_catalog_airflow/dags/common/storage/columns.py

krysal:
I'm not sure how `view_count` will fit within the audio table, is **not** in the schema of the [image table][image_schema], only in the [old one][image_old_schema] and apparently it's not used.

[image_schema]: https://github.com/WordPress/openverse-catalog/blob/main/src/cc_catalog_airflow/local_postgres/0003_openledger_image_schema.sql
[image_old_schema]: https://github.com/WordPress/openverse-catalog/blob/main/src/cc_catalog_airflow/local_postgres/0005_openledger_old_image_schema.sql#L66

dhruvkb:
If it's no longer used, should/can we remove it from the API models and sample data as well? 

I think it's best to remove fields that are no longer used, but we should loop in @obulat and @zackkrida for their opinions on this.

zackkrida:
I agree that we should remove `view_count`, it can be safely removed from everywhere it appears.

CleanShot 2021-07-27 at 11 30 08@2x

krysal:
This is ready for review, can be tested with an audio API provider script, like [Jamendo](https://github.com/WordPress/openverse-catalog/pull/113), and should load the audio table with the columns changed.

For example, the `genres` column as an array of `character varying(80)`:

Screen Shot 2021-07-29 at 11 12 54 AM

src/cc_catalog_airflow/dags/common/storage/audio.py
```diff
@@ -72,16 +72,18 @@
     columns.StringColumn(
         name='category', required=False, size=80, truncate=False,
     ),
-    columns.JSONColumn(
-        name='genre', required=False,
+    columns.ArrayColumn(
```
dhruvkb, line 75: Amazing implementation!

dhruvkb (approved):
LGTM 👍

zackkrida (approved):

PR #131, opened 2021-08-02
Add a PR template to the repository

dhruvkb:
Targets WordPress/openverse#35

This PR adds a PR template to the repository.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,76 @@
+## Fixes
+<!-- If PR doesn't fully resolve the issue, replace 'Fixes' below with 'Related to'. -->
+<!-- If there is no issue being resolved, please consider opening one before creating this pull request. -->
+Fixes #[issue number] by @[issue author]
+
+## Description
+<!-- Concisely describe what the pull request does. -->
+
+## Technical details
+<!-- Add any other information or technical details about the implementation; or delete this section entirely. -->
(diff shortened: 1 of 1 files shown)


obulat (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,76 @@
...
+Fixes #[issue number] by @[issue author]
+
+## Description
+<!-- Concisely describe what the pull request does. -->
+
+## Technical details
+<!-- Add any other information or technical details about the implementation; or delete this section entirely. -->
+
+## Tests
+<!-- Give steps for the reviewer to verify that this PR fixes the problem; or delete this section entirely. -->
```
krysal, line 13: Thinking again about this section, I consider it important so I'd prefer that contributors don't skip it.
```suggestion
## Tests
<!-- Give steps for the reviewer to verify that this PR fixes the problem. -->
```

dhruvkb: It's generally much easier to deduce if a PR works by seeing the issue it addresses.
- If the PR fixes a bug, you can follow its steps to reproduce and they'll not reproduce the issue with the PR applied.
- If the PR adds a feature, we can check if the feature is added.

This section would only be useful in certain edge case scenarios or when an issue does not exist (which generally would be the case when a maintainer makes a PR).

zackkrida (approved):

PR #168, opened 2021-09-10
Refactor db loader to make it easier to update

### LINKED ISSUE #166: [Infrastructure] Simplify the TSV to Postgres loading process
## Current Situation

Currently, the TSV to Postgres loader has various SQL queries in several methods. This means that whenever a TSV or DB column is added, changed or removed, the changes need to be added in several places, and if one of them is missed, the loading process results in error.
Another, connected problem is that there are some legacy workflows left that were created to clean up old data. These are not necessary anymore as all of the data is cleaned up already, and the workflows and tests for them create a great complexity when adding/changing/removing columns.

## Suggested Improvement

We can remove unnecessary legacy workflows, and simplify the loader so that a change to a column can be added only twice: once to the TSV column list, and once to the DB column list.

## Benefit

This will drastically lower the difficulty of updating data model, and reduce the number of bugs created.

## Alternatives

We could leave the loader script as is, and add changes whenever necessary, testing along the way. This is time- and labor-consuming. Considering that we currently need to update audio columns, and add category column to images, this could potentially lead to a lot of errors.

## Implementation

obulat:
## Fixes

Fixes #166 by @obulat

## Description

Currently, changing the media data we collect from the providers is difficult and error-prone.
In addition to updating the corresponding `MediaStore` class's, we also need to update the DB loader workflow. This means updating sql queries in at least 4-5 places. 

To simplify, this PR creates a DbColumn class that defines a database column, and describes its behavior on conflict when updating data.  So, when changing a column, we only need to make two changes for the DB loader workflow:

Add the column (or changes to an existing column) to the:
- relevant `.sql` script in the [`local_postgres` folder](https://github.com/WordPress/openverse-catalog/tree/main/openverse_catalog/local_postgres)
 - dict with DbColumn instances - `util.loader.columns.DB_COLUMNS`
 - list of db columns, in the correct order:  `util.loader.columns.common_columns` or `util.loader.columns.media_columns[AUDIO]`

## Technical details

To enable this change, the columns that are specific to a media type (eg `width` and `height` for images) had to be moved to the end of the TSV file and the DB.
Also, since there are no old TSV files anymore, there is no need to check for `ingestion_type` column, and this functionality is removed.

## Tests

Run the container and the tests with `just up` and `just test`. Also, run one of the Provider API workflows followed by the `tsv_to_postgres_loader`.

## Checklist

- [x] My pull request targets the *default* branch of the repository (`main` or `master`).

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


openverse_catalog/dags/common/storage/audio.py
```diff
@@ -9,6 +9,13 @@
 
 logger = logging.getLogger(__name__)
 
+# Any changes to the AUDIO_TSV_COLUMNS should also be reflected
+# in the db columns. Please, add the column to the:
+# - dict with DbColumn instances - `util.loader.columns.DB_COLUMNS`
```
obulat, line 14:
```suggestion
# - `.sql` script in the `local_postgres` folder
# - dict with DbColumn instances - `util.loader.columns.DB_COLUMNS`
```

openverse_catalog/dags/common/storage/image.py
```diff
@@ -9,6 +9,13 @@
 
 logger = logging.getLogger(__name__)
 
+
+# Any changes to the IMAGE_TSV_COLUMNS should also be reflected
+# in the db columns. Please, add the column to the:
+# - dict with DbColumn instances - `util.loader.columns.DB_COLUMNS`
```
obulat, line 15:
```suggestion
# - `.sql` script in the `local_postgres` folder
# - dict with DbColumn instances - `util.loader.columns.DB_COLUMNS`
```

zackkrida (approved):
This is great. It is _still_ so much work to add a new column (I'm proud of my local `random_test_field` columns, haha!) but this still felt like magic.

I would like to see the comment listing all places column updates need to occur have links to the relevant files. Additionally, I think some even more comprehensive documentation about this somewhere, perhaps the handbook, would be excellent.

I think that would make this even simpler, but this is such excellent work and thank you for doing it.

For example, to add an audio column, I modified:

```shell
	modified:   archive/openledger_sql/03_create_audio_schema.sql
	modified:   openverse_catalog/dags/common/storage/audio.py
	modified:   openverse_catalog/dags/util/loader/column_names.py
	modified:   openverse_catalog/dags/util/loader/columns.py
```

dhruvkb (approved):

obulat:
I agree that it's still a lot of places, and I saw that when I was trying to test out my own instructions in the comments (I had to add a couple of items :) ).
I have realized that there is a much bigger problem with the approach in this PR. In Postgres, we can't change column order easily. It's easy when working locally, because we can create a new DB table with the new column order. However, in production, there are two ways to change the column order, and both are cumbersome:
1. Drop the old db and re-create it with the new column order.
2. Create the new columns in the old table in desired order, move the data, and drop the old columns.
I would like to discuss the best option on today's community meeting.

zackkrida:
> we can't change column order

Could you help me understand where in the codebase the postgres column order matters? That feels like a code smell to me.

I'm still not feeling great, so they're might be something _very obvious_ that I'm forgetting!

obulat:
> > we can't change column order
> 
> @obulat Could you help me understand where in the codebase the postgres column order matters? That feels like a code smell to me.
> 

Thank you for this question! We need the columns to be in a set order because we are loading data from TSV in S3, and if the order of TSV columns is not the same as the order of DB columns, we will not be able to load TSV data into the database.
However, the order of the columns **_doesn't_** matter when we load the data from the loading table into the main table, so you are right that we could leave it. 
Maybe adding a `index in the main image database` and `index in the main audio database` will help to create the list of columns for inserting the data from loading table into the main table easily, without changing the column order. Will have to think about it :)

zackkrida:
Thanks @obulat, so that's this line then, right? 

https://github.com/WordPress/openverse-catalog/blob/277de7a861b863178869dd0bbb93b06f503f074a/openverse_catalog/dags/util/loader/sql.py#L145-L145

Interestingly, the postgres `COPY` command lets you specify a list of columns to import. This list needs to match the order of the tsv file, but _neither_ needs to match the order of the actual table.

Here is an example where my CSV order is _opposite_ the column order of the database:

> The sql command (note that the specified column order is reversed from the actual table, and the table name is 'testing')

```sql
copy testing (three, two, one) 
  from '/Users/zackkrida/Code/test-file.csv'
  with (delimiter ',', format csv);
```

> The `.csv` file (with no headers)

```
third,second,first
third-a,second-b,first-c
```

> And finally, the local DB with the imported data!

CleanShot 2021-09-16 at 11 22 14@2x

Because we're using the `bulk_load` method of `airflow.providers.postgres.hooks.postgres`, I'm not sure we can pass the column order as we would if we were writing raw SQL ourselves.

However, al of this is to say:

1. Our TSV files can have a different order than our DB columns
2. Future TSVs _could have_ different order than past TSVs, and we could write logic to give them different copy commands.

CC @AetherUnbound for fun 😄

AetherUnbound:
Note that this comment comes with very little context as I still haven't familiarized myself with this project :sweat_smile:: 
I do agree @zackkrida, it _should_ be possible to import data irrespective of the column order into Postgres. Frustratingly, the `PostgresHook` is expecting a [string for the filename instead of a buffer](https://github.com/apache/airflow/blob/main/airflow/providers/postgres/hooks/postgres.py#L118), and as Zack noticed too we don't have any access to the `copy_expert` options. I'm still looking over this PR but if it simplifies the workflow I'm all for it! I think down the line, we can take a couple of approaches to make this a bit more robust:
1. Use the `PostgresHook`'s cursor directly to execute the `copy_expert` command. This would give us full control over the load options.
2. We _could_ have the column order specified in a file somewhere, and then rewrite the TSV in that order. I think this could be done with low memory overhead, but we'd still have to rewrite the file and potentially double the size on disk (not sure how much of a constraint that is).

Definitely don't have to have that now, it's something I can look at once I'm off of support rotation :smile:

obulat:
The line 145 that you mention, @zackkrida , is only ran if S3 is unavailable for some reason.

As you know, writing to much to the server where Airflow runs is problematic, because the server can fall if full (that's the reason we implemented log cleanup workflow).

To solve the EC2 overflow problem, ideally, the loader saves the TSV file to S3, and then loads the TSV file from S3 to Postgres. If S3 loader finished its work correctly, then the local (on EC2 Airflow Server) file is deleted. 

The lines where the S3 to Postgres loading happens are here:
https://github.com/WordPress/openverse-catalog/blob/277de7a861b863178869dd0bbb93b06f503f074a/openverse_catalog/dags/util/loader/sql.py#L170-L182

The benefit for this process was [described by Brent](https://github.com/creativecommons/cccatalog/issues/334): 

> This is one step along the path to avoiding saving data on the local disk at all. It should also be faster to load into the DB from S3, since AWS RDS instances provide special optimized functionality to load data from S3 into tables in the DB.

Loading CSV/TSV files from S3 to Postgres using the AWS is highly efficient, however it has one problem for our use case. Postgres copy command, on which AWS loading is based, does not allow for using "header" when loading TSV files. Which means that we cannot change TSV columns without changing the columns in the loading table, and we MUST know the columns in advance.
We can also use `format CSV` as an option for AWS copy command, which _does_ allow us to use headers, and thereby change the order of columns. However, csv escape rules are different from TSV ones, and trying to load a TSV as a CSV breaks due to incorrect JSON escaping. We could use CSV instead of TSV, but  [TSV are more efficient than CSV](https://github.com/eBay/tsv-utils/blob/master/docs/comparing-tsv-and-csv.md).

openverse_catalog/dags/util/test_pg_cleaner.py
```diff
@@ -9,7 +10,7 @@
 from util.loader import test_sql
 
 
-RESOURCES = os.path.join(os.path.abspath(os.path.dirname(__file__)), "test_resources")
+RESOURCES = Path(__file__).parent.resolve() / "test_resources"
```
AetherUnbound, line 13: I love `pathlib`! :star_struck:

AetherUnbound (commented):
Just a couple of thoughts! Nice work on this :smile:

AetherUnbound:
Thanks for all that additional context @obulat! I'd love to sit down at some point (after rotation...I hate that I keep saying that :joy:) so I can wrap my head around the current process a little more :slightly_smiling_face:

zackkrida:
@obulat 

https://github.com/WordPress/openverse-catalog/blob/277de7a861b863178869dd0bbb93b06f503f074a/openverse_catalog/dags/util/loader/sql.py#L175

This line is where the argument for a list of column names goes, so we _can_ still have differing column order from the TSVs and the DB, and not add headers, which is the main thing I wanted to solve for.

openverse_catalog/dags/util/test_pg_cleaner.py
obulat: Me too! Moving the code to pathlib one file at a time :)

openverse_catalog/dags/util/loader/columns.py
```diff
@@ -0,0 +1,302 @@
...
+    """
+    Returns the list of column names that a loading or main db table
+    for the media type should have.
+    :param media_type: One of the supported media types, from
+    `util/constants.py`
+    :param table_type: Whether it is intermediary loading table, or the main
+    `image` or `audio` table.
+    :return: list of strings - names of the columns for selected db table
+    """
+    if table_type not in ["main", "loading"]:
```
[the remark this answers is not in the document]
obulat: I know that the set look up is faster, but since there are only two members in the list, I thought it wouldn't make much difference (or would it?), and so used the more familiar list :)
Consistency is best, anyways, so thank you for this suggestion!

openverse_catalog/dags/util/loader/columns.py
```diff
@@ -0,0 +1,302 @@
...
+    @property
+    def upsert_value(self):
+        strategy = {
+            UpsertStrategy.newest_non_null: _newest_non_null,
+            UpsertStrategy.now: _now,
+            UpsertStrategy.false: _false,
+            UpsertStrategy.merge_jsonb_objects: _merge_jsonb_objects,
+            UpsertStrategy.merge_jsonb_arrays: _merge_jsonb_arrays,
+            UpsertStrategy.merge_array: _merge_array,
+        }.get(self.upsert_strategy)
```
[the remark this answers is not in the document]
obulat: It sure can!

openverse_catalog/dags/util/loader/columns.py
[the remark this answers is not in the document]
AetherUnbound: Totally, it doesn't much matter here haha. I like having it as a pattern though because it's one of those places that could bring unexpected speed-ups!

openverse_catalog/dags/util/loader/test_sql.py
```diff
@@ -576,29 +577,61 @@ def test_upsert_records_replaces_data(postgres_with_load_and_image_table, tmpdir
...
+            col.PROVIDER: PROVIDER,
+            col.SOURCE: SOURCE,
+            col.INGESTION_TYPE: INGESTION_TYPE,
+            col.WIDTH: WIDTH_B,
+            col.HEIGHT: HEIGHT_B,
+        }
     )
+    load_data_query_b = f"""INSERT INTO {load_table} VALUES(
+        {query_values}
+        );"""
```
AetherUnbound, line 634: These changes are great! I'm seeing a lot of this kind of line, so we could potentially have another utility function that generates the `INSERT` SQL string for us.

AetherUnbound (approved):
Looks great!

obulat:
Closing in favor of #196

Issue #285, opened 2021-11-05
Trigger TSV loader as a subDAG or DAG run


AetherUnbound:
## Current Situation

Currently the TSV loader script runs every minute. For a vast majority of its runs (at least at the moment), it skips since there are no files available. Out of the 35,300+ runs we have so far, only 8 have ever run the entire workflow.

## Suggested Improvement

We should instead trigger the TSV loading as part of the provider DAGs. This can be done with either the `TriggerDagRunOperator`[^1] or via SubDAGs[^2].

[^1]: https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/operators/trigger_dagrun/index.html
[^2]: https://airflow.apache.org/docs/apache-airflow/stable/concepts/dags.html#subdags

## Benefit

Both of these strategies would reduce the number of runs this DAG completes significantly. This would make it easier to navigate runs in the UI, track failures, etc.

## Additional context

We should also probably clear out the table of skipped runs at some point. They don't really provide us with much useful information.

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

AetherUnbound:
I have some work in progress on `feature/tsv-loader-improvements#285` which adds loading steps per media type to the end of provider workflows. One issue here though is that it requires passing the output TSV(s) via XCom. Previously, the TSV loader would just pick up the least recently edited TSV regardless of media type and process it. In order to tie the loading directly into the provider workflows, the loading steps need to know _which_ TSV to process and what media type it is. This is further complicated with providers like Wikimedia Commons which produce both image _and_ audio TSVs.

One advantage the current system has is that it can process TSVs even if the provider script fails at some point. We'll definitely want to retain this functionality going forward. As such, we can't wait until the provider script completes to pass the produced TSV into the XComs. Although we want to change the way the provider scripts are structured in #229, for now we can potentially put a shim in between the `<provider>.main` function and the workflow creation DAG. The output directory can be pulled from the module's media store (available at the module level) before actual provider processing and be immediately pushed into XComs. Down the line, I think we'll want to invert this process and have the shim _create_ the output filename and pass it down to the media store, rather than pull it out once the run has started. The alternative to all this is to update every provider script to have a `yield` of some sort at the beginning to report the output TSV filename, which seems like more work than desired at this juncture.

mathemancer:
> One advantage the current system has is that it can process TSVs even if the provider script fails at some point. We'll definitely want to retain this functionality going forward.

[...]

PR #289, opened 2021-11-08
Reduce TSV loader complexity

AetherUnbound:
## Description

This PR attempts to reduce the DAG complexity of the TSV loader by removing redundant steps & consolidating code.

TL;DR: Extra steps were present which were covered by Airflow functionality and have now been removed.

### Technical explanation

There were a number of steps in this DAG that were `DummyOperator`s (no-op steps). From what I could tell, the only purpose of these operators was to use Airflow's Trigger Rules[^1]. Incidentally, the same rules were also added to the actual operational tasks within the DAG, which ended up being a bit redundant. I've removed all of these no-op steps, and I've tried to make the logic a bit clearer (using the `doc_md` attribute for each task). The essence of these changes is:

* The table cleanup task will always run after the create table step (regardless of other failures) to ensure the temporary table gets cleaned up.
* If any of the critical pieces of the loader DAG fail, the TSV will be moved to a "failed" location to be dealt with by an operator manually.
* The TSV will be deleted locally **only after** the data has been successfully loaded into S3. This will preserve the TSV under all failure circumstances, since it is the most expensive step to redo.

There was also a `load_local_data` step present in the original DAG. From the looks of it, this step was intended to run if there were any failures uploading the file to S3; it would attempt to load the data directly from the local machine into Postgres, circumventing S3. I removed this for two reasons:
1. We absolutely want to preserve the data in S3 where possible. I would rather have the data exist in S3 and be loaded into the database at a later time than have the data loaded into the database and no record of it in S3.
1. The specific set of trigger rules applied to this step and its parents meant that this step **would never be run**, it would always be skipped :sweat_smile:. 

I restructured the DAG code a bit to make the pieces easier to grok. There's no need to have separate functions in the `operators` file with the sole purpose of generating a `PythonOperator` or similar - that code can be defined directly in the DAG file. It can also be difficult to tell what the arguments are for `PythonOperator` steps, since we don't get to see the function signature in the DAG file. In order to help with this, I've made all the arguments to the `python_callable` explicit via `op_kwargs` (rather than `op_args`). I discovered that there's a `doc_md` field on tasks as well as DAGs (visible from the task's `Instance Details` page), which serves as a handy place for documentation too.

The first TSV staging step was set up to always succeed and pass along any files (if any) that it found. The results were then used to determine whether the rest of the DAG should be run, which is part of the `ShortCircuitOperator`. This logic itself can be shortcut by raising an `AirflowSkipException` within the called function. This also gives us a clearer indication of what's happening with a given run: if no files were found everything gets skipped since there's nothing to do. Having the first task succeed while all others are skipped seems to indicate that there was some part of the loader that ran while the rest was unneeded, which felt a bit confusing to me.

Lastly, I changed the `max_active_runs` on this DAG to `1`. The DAG is set up to move files out of the main output directory as soon as it identifies that they can be worked on, but there's still a potential race condition where more than one DAG would attempt to process the same file. This will be addressed in #285, in which case we can bump the number of simultaneous runs back up.

Whew, lots of words, thanks if you made it this far :smile:

[^1]: https://airflow.apache.org/docs/apache-airflow/1.10.12/concepts.html#trigger-rules

### Screenshots

**Before**
image

**After**
Screenshot_2021-11-08_13-50-55

## Testing Instructions

1. Run a provider workflow long enough to have it write to disk, then mark it as successful.
1. Enable the `tsv_to_postgres_loader` DAG and wait for it to pick up the TSV generated in step 1>
1. Watch as the DAG (hopefully) succeeds!

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/loader/loader.py
+++ b/openverse_catalog/dags/common/loader/loader.py
@@ -2,19 +2,6 @@ from common.loader import paths, s3, sql
 from common.loader.paths import _extract_media_type
 
 
-def load_local_data(output_dir, postgres_conn_id, identifier, ti):
-    tsv_file_name = paths.get_staged_file(output_dir, identifier)
-    tsv_version = ti.xcom_pull(task_ids="stage_oldest_tsv_file", key="tsv_version")
-
-    media_type = _extract_media_type(tsv_file_name)
-    sql.load_local_data_to_intermediate_table(
-        postgres_conn_id, tsv_file_name, identifier
-    )
-    sql.upsert_records_to_db_table(
-        postgres_conn_id, identifier, media_type=media_type, tsv_version=tsv_version
-    )
-
-
 def copy_to_s3(output_dir, bucket, identifier, aws_conn_id):
     tsv_file_name = paths.get_staged_file(output_dir, identifier)
     media_type = _extract_media_type(tsv_file_name)
--- a/openverse_catalog/dags/common/loader/operators.py
+++ b/openverse_catalog/dags/common/loader/operators.py
@@ -1,121 +1,12 @@
 import logging
 
-from airflow.operators.dummy import DummyOperator
-from airflow.operators.python import PythonOperator, ShortCircuitOperator
-from airflow.utils.trigger_rule import TriggerRule
-from common.loader import loader, paths, smithsonian_unit_codes, sql
+from airflow.operators.python import PythonOperator
+from common.loader import smithsonian_unit_codes, sql
 
 
 logger = logging.getLogger(__name__)
 
 TIMESTAMP_TEMPLATE = "{{ ts_nodash }}"
-
-
-def get_file_staging_operator(
-    output_dir, minimum_file_age_minutes, identifier=TIMESTAMP_TEMPLATE
-):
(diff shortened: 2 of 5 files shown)


zackkrida (approved):
Works great. Ran a few DAGs and confirmed everything worled as expected:
CleanShot 2021-11-08 at 18 28 59@2x

Then `docker stop`-ed the `minio` container, ran another DAG, and confirmed that the failures are moved correctly:

CleanShot 2021-11-08 at 18 40 34@2x

```shell
# note, my local `.env` still had `/tmp` for the output dir
airflow@405c1da0932c:/tmp/db_loader_failures/20211108T233800$ ls
wordpress_image_v001_20211108233707.tsv
```

krysal (approved):
So nice to have this DAG simplified _and_ documented ✨ It worked perfectly.

And having the Minio UI to inspect the bucket is quite handy too 👍

CleanShot 2021-11-10 at 14 01 12@2x

PR #296, opened 2021-12-07
Retire common_api_workflows, clean up config

### LINKED ISSUE #284: Retire/remove common_api_workflows.py
`common_api_workflows.py` looks to be an abstraction for creating DAGs that shouldn't be used. It appears that this was a transitional piece of code to replace what was previously [a single, daily DAG for multiple providers](https://github.com/WordPress/openverse-catalog/commit/0051db985b0a9ac6a50c40825837b273af11c0d6).

We should remove this code and archive the broken Thingiverse provider script, as discussed in #283

AetherUnbound:
## Fixes

Fixes #284 by @zackkrida

## Description

This PR moves the `common_api_workflows.py` file to the `retired` folder. It also removes the `common.conf` module, as `common_api_workflows` was its primary consumer. I've moved the `DAG_DEFAULT_ARGS` into `common.dag_factory`, since that's the only other place that value was used. I also updated the `dag_factory` code to prevent using mutable default arguments (and removed a `print` statement).

[^1]: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

## Testing Instructions

`just test`

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


openverse_catalog/dags/common/dag_factory.py
```diff
@@ -77,12 +85,10 @@ def create_provider_api_workflow(
...
                       a given dagrun may take.
     """
-    args = deepcopy(default_args)
-    args.update(start_date=start_date)
-    print(args)
+    default_args = default_args or DAG_DEFAULT_ARGS
     dag = DAG(
         dag_id=dag_id,
-        default_args=args,
+        default_args={**default_args, "start_date": start_date},
```
obulat, line 91: Would this always set the start date to the default for `create_provider_api_workflow` `(1970, 1,1)`? Not that it really matters for Airflow, but in that case we can probably delete the `start_date` from the `DAG_DEFAULT_ARGS`?

obulat (approved):
Great cleanup, and 17 checks passed!
I have left a question inline.

openverse_catalog/dags/common/dag_factory.py
AetherUnbound: Yes, in this case it would always replace the start date within the functions. However, I could _potentially_ see using the `DAG_DEFAULT_ARGS` in full elsewhere, in which case it might be a good idea to keep it there. What do you think?

zackkrida (approved):
Nice!

Issue #229, opened 2021-10-07
[Quality] Delay object creation until runtime


AetherUnbound:
## Current Situation

Related to #199 - we have a number of objects (`DelayedRequester`, `MediaStore`, etc.) that are created at import time rather than runtime. This has a negative performance impact on DAG processing time. The DAG parsing does not need to have these object available, and initializing them every time the parsing is done can unnecessarily increase parsing time. This can be seen from the logging that's done during DAG parsing:

```
$ airflow dags list -v
[2021-10-07 21:37:40,207] {dagbag.py:496} INFO - Filling up the DagBag from /usr/local/airflow/dags
[2021-10-07 21:37:40,242] {media.py:61} INFO - Initialized image MediaStore with provider finnishmuseums
[2021-10-07 21:37:40,242] {media.py:162} INFO - No given output directory. Using OUTPUT_DIR from environment.
[2021-10-07 21:37:40,243] {media.py:180} INFO - Output path: /tmp/finnishmuseums_image_v001_20211007213740.tsv
{'owner': 'data-eng-admin', 'depends_on_past': False, 'start_date': datetime.datetime(2020, 9, 1, 0, 0), 'email_on_retry': False, 'retries': 3, 'retry_delay': datetime.timedelta(seconds=900)}
[2021-10-07 21:37:40,248] {media.py:61} INFO - Initialized image MediaStore with provider smithsonian
[2021-10-07 21:37:40,248] {media.py:162} INFO - No given output directory. Using OUTPUT_DIR from environment.
[2021-10-07 21:37:40,248] {media.py:180} INFO - Output path: /tmp/smithsonian_image_v001_20211007213740.tsv
{'owner': 'data-eng-admin', 'depends_on_past': False, 'start_date': datetime.datetime(1970, 1, 1, 0, 0), 'email_on_retry': False, 'retries': 3, 'retry_delay': datetime.timedelta(seconds=900)}
[2021-10-07 21:37:40,257] {media.py:61} INFO - Initialized image MediaStore with provider europeana
[2021-10-07 21:37:40,257] {media.py:162} INFO - No given output directory. Using OUTPUT_DIR from environment.
[2021-10-07 21:37:40,257] {media.py:180} INFO - Output path: /tmp/europeana_image_v001_20211007213740.tsv
[2021-10-07 21:37:40,316] {media.py:61} INFO - Initialized audio MediaStore with provider wikimedia_audio
[...]

PR #357, opened 2022-02-10
Trigger TSV loading immediately after workflow

### LINKED ISSUE #285: Trigger TSV loader as a subDAG or DAG run

AetherUnbound:
## Fixes

Fixes #285 by @AetherUnbound

## Description

At long last! This PR makes some significant changes to the TSV loading process.

**TL;DR**: The TSV loading steps are now connected within the provider workflow DAG and get triggered immediately after the provider finishes/fails.

**The long-winded description**:

The current TSV workflow is:
1. Provider workflow runs
1. TSV loader runs asynchronously every minute and ingests the oldest file

It is more desirable to have synchronous loading for the case where we're ingesting a resource daily/monthly for a few reasons:
- Running the TSV loader every minute results in a very large number of skipped runs (currently), see #309.
- If a problem occurs with a sheet, that sheet is moved to a different location and we have to hunt down the errant file. This also prevents the file from being easily re-run once it's corrected as it must be moved back into the staging location.
- It makes it easier to determine which provider TSV caused the load to fail, since the steps are attached directly to the DAG.

The new change adds one or several [task groups](https://www.astronomer.io/guides/task-groups) on the end of our **basic provider workflow DAG**. I have intentionally left all pieces of the current TSV loader functioning for when we get to enabling the partitioned re-ingestion DAGs. As such, this PR only makes minor tweaks to existing functions; where differing functionality is needed, new functions have been added to support this work.

It's important to note that each provider workflow DAG must now report the media types it supports. For most of our DAGs this is `image`, but any provider script that produces media records beyond that must do so explicitly (see: audio DAGs). The setup is complicated _a bit_ by the fact that a provider script can produce multiple media types (see: Wikimedia commons). In order to handle this case, each media type generates its own ingestion task group. This also helps ensure that the loads are independent of one another.

As described above, the current method ingests the least recently modified file. This new method gets handed the file to ingest explicitly from upstream tasks via [XComs](https://airflow.apache.org/docs/apache-airflow/stable/concepts/xcoms.html). Due to the way our provider API scripts are set up, the location on disk must be retrieved from the provider API module itself. I've added a wrapper function in the DAG which accomplishes this. Ideally with the work from #229 and other efforts, we'll be able to construct the output path using the DAG's execution date and pass that into the `MediaStore` (rather than retrieving it from the store).

In addition to new tests, I've also updated some of the tests to factor out common fixtures.

Phew, okay, I think that's it. There's a lot going on in this PR, even though it isn't a _ton_ of code. Let me know if there's anything I should explain more thoroughly. Much of this is intermediate work to 1) help us get up and running now, 2) pave the way for future work, and 3) prevent this PR from being enormous!

### Screenshots

**Simple, single-media DAG**
Screenshot_2022-02-10_11-32-54

**Multi-media DAG**
Screenshot_2022-02-10_11-33-08

## Suggested review order
1. `dags/common/storage/media.py` & `dags/common/loader/paths.py`- making some attributes public so they can be used in the DAG
2. `dags/common/loader/*` - new, more generic functions for the TSV loading steps
3. `dags/providers/*_workflow.py` - added media types where necessary
4. `dags/common/dag_factory.py` - bulk of the changes, adding on the TSV loading steps to the end of the ingestion workflow
5. `tests/*` and everything else

## Testing Instructions

1. `just test`
1. Run a provider workflow
1. Either "mark successful" or "mark failed" in the UI after some lines have been written
1. Observe that the loader runs immediately after the workflow step and ingests data directly into the database

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/dag_factory.py
+++ b/openverse_catalog/dags/common/dag_factory.py
@@ -1,9 +1,11 @@
+import inspect
 import logging
 import os
 from datetime import datetime, timedelta
 from typing import Callable, Dict, List, Optional
 
 from airflow import DAG
+from airflow.models import TaskInstance
 from airflow.models.baseoperator import cross_downstream
 from airflow.operators.dummy import DummyOperator
 from airflow.operators.python import PythonOperator
@@ -32,6 +34,48 @@ DAG_DEFAULT_ARGS = {
 DATE_RANGE_ARG_TEMPLATE = "{{ macros.ds_add(ds, -%s) }}"
 TIMESTAMP_TEMPLATE = "{{ ts_nodash }}"
 XCOM_PULL_TEMPLATE = "{{ ti.xcom_pull(task_ids='%s', key='%s') }}"
+
+
+def _push_output_paths_wrapper(
+    func: Callable,
+    media_types: List[str],
+    ti: TaskInstance,
+    *args,
+):
+    """
+    Run the provided callable after pushing the calculated output directories
+    for each media store to XComs. Output locations are pushed under keys with
+    the format `<media-type>_tsv`. This is a temporary workaround due to the nature
+    of the current provider scripts. Once
+    https://github.com/WordPress/openverse-catalog/issues/229 is addressed and the
+    provider scripts are refactored into classes, this wrapper can either be updated
+    or the XCom pushing can be moved into the provider initialization.
+    """
+    logger.info("Pushing available store paths to XComs")
+    module = inspect.getmodule(func)
+    stores = {}
+
+    # Stores exist at the module level, so in order to retrieve the output values we
+    # must first pull the stores from the module.
+    for media_type in media_types:
+        if not (store := getattr(module, f"{media_type}_store", None)):
+            continue
+        stores[media_type] = store
+
+    if not stores:
+        raise ValueError(
+            f"Expected a store of the following types in {module.__name__} "
+            f"but none were found: {media_types}"
+        )
+
+    for media_type, store in stores.items():
+        logger.info(f"{media_type.capitalize()} store location: {store.output_path}")
+        ti.xcom_push(key=f"{media_type}_tsv", value=store.output_path)
+
+    logger.info("Running provider function")
+    # Not passing kwargs here because Airflow throws a bunch of stuff in there that none
+    # of our provider scripts are expecting.
+    return func(*args)
 
 
 def create_provider_api_workflow(
@@ -115,7 +159,8 @@ def create_provider_api_workflow(
         # based on the media type
         pull_data = PythonOperator(
             task_id=f"pull_{media_type_name}_data",
-            python_callable=main_function,
+            python_callable=_push_output_paths_wrapper,
+            op_kwargs={"func": main_function, "media_types": media_types},
             depends_on_past=False,
             **kwargs,
         )
--- a/openverse_catalog/dags/common/dag_factory.py
+++ b/openverse_catalog/dags/common/dag_factory.py
@@ -62,7 +62,7 @@ def _push_output_paths_wrapper(
             continue
         stores[media_type] = store
 
-    if not stores:
+    if len(stores) != len(media_types):
         raise ValueError(
             f"Expected a store of the following types in {module.__name__} "
             f"but none were found: {media_types}"
--- a/openverse_catalog/dags/common/dag_factory.py
+++ b/openverse_catalog/dags/common/dag_factory.py
@@ -155,8 +155,6 @@ def create_provider_api_workflow(
                 "execution_timeout": dagrun_timeout,
             }
 
-        # Assumption - this will push an XCOM with the TSV location using a key
-        # based on the media type
         pull_data = PythonOperator(
             task_id=f"pull_{media_type_name}_data",
             python_callable=_push_output_paths_wrapper,
@@ -175,6 +173,7 @@ def create_provider_api_workflow(
                         "identifier": TIMESTAMP_TEMPLATE,
                         "media_type": media_type,
                     },
+                    trigger_rule=TriggerRule.NONE_SKIPPED,
                     doc_md="Create a temporary loading table for "
                     f"ingesting {media_type} data from a TSV",
                 )
@@ -188,6 +187,7 @@ def create_provider_api_workflow(
                         "s3_prefix": f"{media_type}/{provider_name}",
                         "aws_conn_id": AWS_CONN_ID,
                     },
+                    trigger_rule=TriggerRule.NONE_SKIPPED,
                 )
                 load_from_s3 = PythonOperator(
                     task_id="load_from_s3",
--- a/openverse_catalog/dags/common/dag_factory.py
+++ b/openverse_catalog/dags/common/dag_factory.py
@@ -1,5 +1,6 @@
 import logging
 from datetime import datetime, timedelta
+from typing import Callable, Dict, Optional
 
 from airflow import DAG
 from airflow.models.baseoperator import cross_downstream
@@ -25,16 +26,17 @@ DATE_RANGE_ARG_TEMPLATE = "{{ macros.ds_add(ds, -%s) }}"
 
 
 def create_provider_api_workflow(
-    dag_id,
-    main_function,
-    default_args=None,
-    start_date=datetime(1970, 1, 1),
-    max_active_tasks=1,
-    schedule_string="@daily",
-    dated=True,
-    day_shift=0,
-    dagrun_timeout=timedelta(hours=12),
-    doc_md="",
+    dag_id: str,
+    main_function: Callable,
+    default_args: Optional[Dict] = None,
+    start_date: datetime = datetime(1970, 1, 1),
+    max_active_tasks: int = 1,
+    schedule_string: str = "@daily",
+    dated: bool = True,
+    day_shift: int = 0,
+    dagrun_timeout: timedelta = timedelta(hours=12),
+    doc_md: str = "",
+    media_type: str = "image",
 ):
     """
     This factory method instantiates a DAG that will run the given
@@ -71,6 +73,9 @@ def create_provider_api_workflow(
                       be run (if `dated=True`).
     dagrun_timeout:   datetime.timedelta giving the total amount of time
                       a given dagrun may take.
+    doc_md:           string which should be used for the DAG's documentation markdown
+    media_type:       string describing the media type that this provider handles
+                      (e.g. "image", "audio", "video", "mixed", etc)
     """
     default_args = default_args or DAG_DEFAULT_ARGS
     dag = DAG(
@@ -89,7 +94,7 @@ def create_provider_api_workflow(
     with dag:
         if dated:
             PythonOperator(
-                task_id="pull_image_data",
+                task_id=f"pull_{media_type}_data",
                 python_callable=main_function,
                 op_args=[DATE_RANGE_ARG_TEMPLATE % day_shift],
                 execution_timeout=dagrun_timeout,
@@ -97,7 +102,7 @@ def create_provider_api_workflow(
             )
         else:
             PythonOperator(
-                task_id="pull_image_data",
+                task_id=f"pull_{media_type}_data",
                 python_callable=main_function,
                 depends_on_past=False,
             )
--- a/openverse_catalog/dags/common/dag_factory.py
+++ b/openverse_catalog/dags/common/dag_factory.py
@@ -89,7 +89,7 @@ def create_provider_api_workflow(
     day_shift: int = 0,
     dagrun_timeout: timedelta = timedelta(hours=12),
     doc_md: str = "",
-    media_types: List[str] = ("image",),
+    media_types: Sequence[str] = ("image",),
 ):
     """
     This factory method instantiates a DAG that will run the given
@@ -128,7 +128,7 @@ def create_provider_api_workflow(
                       a given dagrun may take.
     doc_md:           string which should be used for the DAG's documentation markdown
     media_types:      list describing the media type(s) that this provider handles
-                      (e.g. ("image",), ["audio"], ["image", "audio"], etc.)
+                      (e.g. `["audio"]`, `["image", "audio"]`, etc.)
     """
     default_args = default_args or DAG_DEFAULT_ARGS
     media_type_name = "mixed" if len(media_types) > 1 else media_types[0]
--- a/openverse_catalog/dags/common/dag_factory.py
+++ b/openverse_catalog/dags/common/dag_factory.py
@@ -2,7 +2,7 @@ import inspect
 import logging
 import os
 from datetime import datetime, timedelta
-from typing import Callable, Dict, List, Optional
+from typing import Callable, Dict, List, Optional, Sequence
 
 from airflow import DAG
 from airflow.models import TaskInstance
@@ -40,7 +40,7 @@ def _push_output_paths_wrapper(
     func: Callable,
     media_types: List[str],
     ti: TaskInstance,
-    *args,
+    args: Sequence = None,
 ):
     """
     Run the provided callable after pushing the calculated output directories
@@ -51,6 +51,7 @@ def _push_output_paths_wrapper(
     provider scripts are refactored into classes, this wrapper can either be updated
     or the XCom pushing can be moved into the provider initialization.
     """
+    args = args or []
     logger.info("Pushing available store paths to XComs")
     module = inspect.getmodule(func)
     stores = {}
@@ -148,19 +149,16 @@ def create_provider_api_workflow(
     )
 
     with dag:
-        kwargs = {}
+        pull_kwargs = {"func": main_function, "media_types": media_types}
         if dated:
-            kwargs = {
-                "op_args": [DATE_RANGE_ARG_TEMPLATE % day_shift],
-                "execution_timeout": dagrun_timeout,
-            }
+            pull_kwargs["args"] = [DATE_RANGE_ARG_TEMPLATE % day_shift]
 
         pull_data = PythonOperator(
             task_id=f"pull_{media_type_name}_data",
             python_callable=_push_output_paths_wrapper,
-            op_kwargs={"func": main_function, "media_types": media_types},
+            op_kwargs=pull_kwargs,
             depends_on_past=False,
-            **kwargs,
+            execution_timeout=dagrun_timeout if dated else None,
         )
 
         for media_type in media_types:
--- a/openverse_catalog/dags/common/dag_factory.py
+++ b/openverse_catalog/dags/common/dag_factory.py
@@ -64,8 +64,8 @@ def _push_output_paths_wrapper(
 
     if len(stores) != len(media_types):
         raise ValueError(
-            f"Expected a store of the following types in {module.__name__} "
-            f"but none were found: {media_types}"
+            f"Expected stores in {module.__name__} were missing: "
+            f"{list(set(media_types) - set(stores))}"
         )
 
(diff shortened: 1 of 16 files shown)


obulat (commented):
The code changes are really nice!

I tried to run jamendo, science_museum, wikimedia_commons and wordpress workflows. For Wikipedia, I get this error: 
```
[2022-02-11, 18:31:16 +03] {taskinstance.py:1700} ERROR - Task failed with exception
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/airflow/models/taskinstance.py", line 1329, in _run_raw_task
    self._execute_task_with_callbacks(context)
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/airflow/models/taskinstance.py", line 1455, in _execute_task_with_callbacks
    result = self._execute_task(context, self.task)
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/airflow/models/taskinstance.py", line 1506, in _execute_task
    result = execute_callable(context=context)
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/airflow/operators/python.py", line 172, in execute
    self.op_kwargs = determine_kwargs(self.python_callable, self.op_args, context)
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/airflow/utils/operator_helpers.py", line 111, in determine_kwargs
    raise ValueError(f"The key {name} in args is part of kwargs and therefore reserved.")
ValueError: The key func in args is part of kwargs and therefore reserved.
```

AetherUnbound:
Oh shoot, thanks - I'll look into that!

AetherUnbound:
Update: this issue is _really_ difficult to track down O_o it has to do with some magic Airflow is doing to manage context.

AetherUnbound:
Figured it out! We had to explicitly pass in `args` as a `kwarg` along with the others in order to prevent Airflow from freaking out. I also addressed a few other issues:
* The loading now skips if no file was written, which is what we discussed in #285 anyway
* The loading table `identifier` was constructed in a way that it could cause a race condition if two providers of the same media type were ran on the same day. Since we're going to be doing that A LOT (ideally), I changed this so the `identifier` now includes the provider name for better distinction.

tests/dags/common/test_dag_factory.py
```diff
@@ -1,8 +1,63 @@
...
+import requests
+from airflow.models import TaskInstance
+from common import dag_factory
+
+from tests.dags.common.test_resources import fake_provider_module
+
+
+@pytest.mark.parametrize(
+    "func, media_types, stores",
+    [
```
krysal, line 13: Great testing, covering a wide range of paths!

openverse_catalog/dags/database/loader_workflow.py
```diff
@@ -104,12 +104,19 @@
         """
         ),
     )
+
+    media_type_xcom = (
+        "{{ ti.xcom_pull(task_ids='%s', key='media_type') }}"
+        % stage_oldest_tsv_file.task_id
```
sarayourfriend, line 110: I always preferred a good old `.format` call than the `%` version when possible. Too many rules to remember with the `%` method.

```suggestion
        "{{{{ ti.xcom_pull(task_ids='{task_id}', key='media_type') }}}}".format(
            task_id=stage_oldest_tsv_file.task_id
        )
```

Or even just an `f` string:

```suggestion
        "{{{{ ti.xcom_pull(task_ids='{stage_oldest_tsv_file.task_id}', key='media_type') }}}}"
```

I suppose some folks don't like the curly brace escaping though?

krysal (approved):
Tested with StockSnap and Jamendo scripts and it works perfectly! Thanks for the detailed explanation and comments 🌟

openverse_catalog/dags/common/dag_factory.py
```diff
@@ -71,8 +126,15 @@ def create_provider_api_workflow(
                       be run (if `dated=True`).
     dagrun_timeout:   datetime.timedelta giving the total amount of time
                       a given dagrun may take.
+    doc_md:           string which should be used for the DAG's documentation markdown
+    media_types:      list describing the media type(s) that this provider handles
+                      (e.g. ("image",), ["audio"], ["image", "audio"], etc.)
```
sarayourfriend, line 131: Mixing the types in the documentation is confusing to me for whatever reason... or at least it took a little longer of reading it to understand what was going on. Maybe some back ticks would help?

```suggestion
                      (e.g. `("image",)`, `["audio"]`, `["image", "audio"]`, etc.)
```

It already being nested inside of grammatical parenthesis makes the tuple version not stand out as much.

openverse_catalog/dags/common/dag_factory.py
```diff
@@ -87,20 +149,71 @@ def create_provider_api_workflow(
...
+                    python_callable=sql.drop_load_table,
+                    op_kwargs={
+                        "postgres_conn_id": DB_CONN_ID,
+                        "identifier": identifier,
+                        "media_type": media_type,
+                    },
+                    trigger_rule=TriggerRule.ALL_DONE,
+                )
+                [create_loading_table, copy_to_s3] >> load_from_s3
+                [create_loading_table, load_from_s3] >> drop_loading_table
```
sarayourfriend, line 214: What's going on here? :magic_wand:

openverse_catalog/dags/common/dag_factory.py
```diff
@@ -22,19 +32,64 @@
...
+    # must first pull the stores from the module.
+    for media_type in media_types:
+        if not (store := getattr(module, f"{media_type}_store", None)):
+            continue
+        stores[media_type] = store
+
+    if len(stores) != len(media_types):
+        raise ValueError(
+            f"Expected a store of the following types in {module.__name__} "
+            f"but none were found: {media_types}"
```
stacimc, line 68: It might be useful to log which stores were missing, or else change the wording slightly for clarity (`none were found` -> `not all were found`?).

openverse_catalog/dags/common/dag_factory.py
stacimc: I think this is [setting the dependencies](https://airflow.apache.org/docs/apache-airflow/stable/concepts/tasks.html#relationships) for the tasks within the `load_data` TaskGroup. 

@AetherUnbound I'm curious if this is logically equivalent to:
```
[create_loading_table, copy_to_s3] >> load_from_s3 >> drop_loading_table
```
? I think it's still clear this way just wondering what the difference is.

I understand the sequence here but it took a struggle, mostly because of unfamiliarity with Airflow (and the actual process!). I think it would be worth adding some comments (or maybe expanding the docstring for the factory method) to describe the flow.

openverse_catalog/dags/common/loader/s3.py
```diff
@@ -28,6 +38,34 @@ def copy_file_to_s3_staging(
...
+    ti,
+):
+    """
+    Copy a TSV file to S3 with the given prefix.
+    The TSV's version is pushed to the `tsv_version` XCom, and the constructed
+    S3 key is pushed to the `s3_key` XCom.
+    The TSV is removed after the upload is complete.
+    """
+    if tsv_file_path is None:
+        raise FileNotFoundError(f"Expected file {tsv_file_path} was not provided")
```
stacimc, line 55: In this case it's not so much that the file wasn't provided but the path itself, so there shouldn't be a `tsv_file_path` to log:
```suggestion
        raise FileNotFoundError(f"No TSV file path was provided.")
```

openverse_catalog/dags/database/loader_workflow.py
obulat: I would prefer fstrings if they work here.

openverse_catalog/dags/common/dag_factory.py
```diff
@@ -22,19 +32,64 @@
...
+    main_function: Callable,
+    default_args: Optional[Dict] = None,
+    start_date: datetime = datetime(1970, 1, 1),
+    max_active_tasks: int = 1,
+    schedule_string: str = "@daily",
+    dated: bool = True,
+    day_shift: int = 0,
+    dagrun_timeout: timedelta = timedelta(hours=12),
+    doc_md: str = "",
+    media_types: List[str] = ("image",),
```
obulat, line 92: Does `List` type also include tuples? I understand that the default value cannot be a list because it's mutable. Does the type declaration match the default value?

obulat (commented):
What happens when the `pull_data` task is manually set to successful? Does it not run `commit` for any media that hadn't been written to disk yet?
Because when I run `wikimedia` DAG and set `pull_mixed_data` to `successful`, `load_image_data` tasks are all green, but the `load_audio_data` `copy_to_s3` and `load_from_s3` ares skipped. I guess it's because there is not file written to the disk? It would be nice to skip `create_loading_table` task if there is no data available, if that's at all possible as it is run concurrently with `copy_to_s3`...

openverse_catalog/dags/common/dag_factory.py
```diff
@@ -22,19 +32,64 @@
...
-    dated=True,
-    day_shift=0,
-    dagrun_timeout=timedelta(hours=12),
-    doc_md="",
+    dag_id: str,
+    main_function: Callable,
+    default_args: Optional[Dict] = None,
+    start_date: datetime = datetime(1970, 1, 1),
+    max_active_tasks: int = 1,
+    schedule_string: str = "@daily",
```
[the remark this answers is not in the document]
AetherUnbound: It does not, unfortunately. Literals could be helpful here, but the `@daily`/`@monthly` schedule intervals are actually just [aliases to cron strings, which is what this field represents](https://airflow.apache.org/docs/apache-airflow/stable/dag-run.html?highlight=schedule%20interval#dag-runs).

openverse_catalog/dags/common/dag_factory.py
sarayourfriend: :exploding_head: 

This is cryptic, but I understand that if you know Airflow already then it's pretty normal syntax! I'm happy to just accept it as is (and learn more about it at a later date!) :slightly_smiling_face:

openverse_catalog/dags/common/loader/loader.py
```diff
@@ -31,3 +31,19 @@ def load_s3_data(
...
+    media_type,
+    tsv_version,
+    identifier,
+):
+    sql.load_s3_data_to_intermediate_table(
+        postgres_conn_id, bucket, key, identifier, media_type
+    )
+    sql.upsert_records_to_db_table(
+        postgres_conn_id, identifier, media_type=media_type, tsv_version=tsv_version
+    )
```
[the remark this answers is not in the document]
AetherUnbound: I'm not sure the original context/rationale for this decision, but an intermediate table has a few benefits:
1. Isolation from the primary table, so it isn't impacted by the additional time it might take to get the data from S3, only the time it takes to upsert the records from another table
2. We may have loads that affect a large enough number of records that couldn't adequately fit in memory and would be easier to aggregate on disk
3. The intermediate tables get indexed, and those indices can be carried forward into the primary table & reduce the upsertion time

I'm not sure we're at the place _yet_ where we're processing that much data, but we haven't run all the providers regularly so it's possible!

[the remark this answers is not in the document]
sarayourfriend: Okay, cool! Thanks for the explanation of benefits :grinning:

openverse_catalog/dags/common/loader/s3.py
```diff
@@ -28,6 +38,34 @@ def copy_file_to_s3_staging(
...
+    tsv_file = Path(tsv_file_path)
+    if not tsv_file.exists():
+        raise AirflowSkipException(f"TSV file {tsv_file} had no data.")
```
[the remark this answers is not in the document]
AetherUnbound: Yes! That's a more appropriate message. This case typically occurs when there is no data to process, but the current message is misleading.

openverse_catalog/dags/database/loader_workflow.py
AetherUnbound: I was trying to avoid having to escape the brackets, since it ends up with a lot of `{{{{` :sweat_smile: If y'all prefer it though, I'll change it!

openverse_catalog/dags/common/dag_factory.py
AetherUnbound: Ah, totally fair. My intent here was to point out that any sequence type is accepted. We're using a tuple as the default argument because lists are mutable and mutable default arguments are probably the most common "gotcha" in Python[^1]. If you think that this example is better served by consistent types, I'm happy to change it!

[^1]: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

openverse_catalog/dags/common/loader/s3.py
AetherUnbound: :woman_facepalming: totally true, I'll update this!

openverse_catalog/dags/common/dag_factory.py
AetherUnbound: Ugh, even after [all my best intentions](https://github.com/WordPress/openverse-catalog/pull/357#discussion_r809458638) I still get this type wrong! I'll fix this to make it more generic.

AetherUnbound:
> What happens when the `pull_data` task is manually set to successful? Does it not run `commit` for any media that hadn't been written to disk yet? Because when I run `wikimedia` DAG and set `pull_mixed_data` to `successful`, `load_image_data` tasks are all green, but the `load_audio_data` `copy_to_s3` and `load_from_s3` ares skipped. I guess it's because there is not file written to the disk? It would be nice to skip `create_loading_table` task if there is no data available, if that's at all possible as it is run concurrently with `copy_to_s3`...

Ha, I know exactly the case you're talking about! The wikimedia dag is `dated`, so IIRC it's just pulling _all_ data for that day and sorting it into image & audio bins depending on the media type. I think it's likely that the provider script just hadn't come across any audio data yet, so there was nothing to write to disk.

On your second question: I think that'd be possible, we'd need some additional logic to determine whether the file has any data. That could be done as part of the create table step, or perhaps it'd be better served as its own step which skips the rest of the workflow. As it stands currently, it's easier to unconditionally create the table & delete it at the end of the load, even if nothing happens with it.

openverse_catalog/dags/common/dag_factory.py
stacimc: Oh that's perfect! Definitely, I think copying it over would make a lot of sense 😄

openverse_catalog/dags/common/dag_factory.py
AetherUnbound: How is this?
image

openverse_catalog/dags/common/loader/s3.py
```diff
@@ -28,6 +35,34 @@ def copy_file_to_s3_staging(
...
+    if tsv_file_path is None:
+        raise FileNotFoundError("No TSV file path was provided")
+    tsv_file = Path(tsv_file_path)
+    if not tsv_file.exists():
+        raise AirflowSkipException(f"TSV file {tsv_file} does not exist.")
+    tsv_version = paths.get_tsv_version(tsv_file_path)
+    s3_key = f"{s3_prefix}/{tsv_file.name}"
+    logger.info(f"Uploading {tsv_file_path} to {s3_bucket}:{s3_key}")
+    s3 = S3Hook(aws_conn_id=aws_conn_id)
+    s3.load_file(tsv_file_path, s3_key, bucket_name=s3_bucket)
```
rbadillap, line 60: Is it possible that there can be a file with the same name already placed in the bucket? I'm wondering if worth the chance to run some validation to make sure if there's a file with the same name.

[As per the source code](https://github.com/apache/airflow/blob/4dd751d2fc3f091417146079d7d1b3ac6dd86c9c/airflow/providers/amazon/aws/hooks/s3.py#L497) the default behavior is to overwrite the file, is this the expected behavior?

AetherUnbound: Good question! The S3 key is constructed using the provider, media type, TSV version, and timestamp for the run:
https://github.com/WordPress/openverse-catalog/blob/df7f68460a7ceb14cc39a02743ba8190411c017c/openverse_catalog/dags/common/storage/media.py#L175

The only case where we'll have a duplicate file is if the TSV is attempted to be uploaded after it has already been successfully uploaded. This would require some pretty intentional manual intervention to accomplish, so I think we're good :slightly_smiling_face: 

Also it looks like the default is actually to raise an error if the file exists, so we won't perform any unintentional overwrites either:
https://github.com/apache/airflow/blob/4dd751d2fc3f091417146079d7d1b3ac6dd86c9c/airflow/providers/amazon/aws/hooks/s3.py#L483

openverse_catalog/dags/common/dag_factory.py
sarayourfriend: I do think consistent types would be good as well as the back-ticks to help delineate the code parens from the grammatical ones. If anything, always using a tuple at least communicates to the reader that the list will not be modified.

obulat:
> Ha, I know exactly the case you're talking about! The wikimedia dag is dated, so IIRC it's just pulling all data for that day and sorting it into image & audio bins depending on the media type. I think it's likely that the provider script just hadn't come across any audio data yet, so there was nothing to write to disk.

Yes, I suspected that it hadn't gotten any audio, so I reran the script untill I saw the `Got x audio so far` logging message, with x being around 30. Then I set the `pull_mixed_media` task to sucessful in the UI

obulat (approved):

AetherUnbound:
> Yes, I suspected that it hadn't gotten any audio, so I reran the script until I saw the `Got x audio so far` logging message, with x being under 100.

Ah, in that case you're totally write. We only write to buffer every 100 records, so if we never reach 100 records, the only write happens at the end unfortunately.
 
> I wonder what happens in case of an error during the pull stage? All the data would be lost, and hopefully recollected in a retry, right? I wish there was a way to re-start a failed pull task from where it failed, and not from the start, but it probably goes against Airflow philosophy of "tasks should always be deterministic and return the same result for the same inputs".

That's a good thing to be thinking about! It's possible that we could store some recovery information on failure (maybe the current set of query params?) and pick those up if the task is retried. Currently, if the data pull fails, one of two things will happen:
1. If =>100 records have been processed, the generated TSV will exist on disk with the last buffer save, and that will be picked up by the TSV loader.
2. If <100 records have been processed, no data has been written so no data is written to disk (I believe). 

I think it's worth looking into how the latter can be changed! I'll make an issue for it.

stacimc (approved):
Looks great 🚀 

> And frankly, I guess it's OK because normally you wouldn't stop the task manually.
We really _do_ seem to get bit by this a lot! Thanks for all the detailed explanations in this discussion 😄

PR #361, opened 2022-02-12
Reconfigure retries & timeouts for typical ingestion DAGs

### LINKED ISSUE #348: Use `execution_timeout` rather than `dagrun_timeout`
## Description

Currently we're using `dagrun_timeout` to set a timeout for our provider workflows. According to [this documentation](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/index.html), the `dagrun_timeout` allows one to:

> specify how long a DagRun should be up before timing out / failing, so that new DagRuns can be created. The timeout is only enforced for scheduled DagRuns.

It seems like that second sentence may not be true for v2.2.3, since we saw an instance where a manually triggered DAG run also (inappropriately) respected the `dagrun_timeout`. The `dagrun_timeout` is a bit confusing, and [has been discussed on the dev mailing list before](https://lists.apache.org/thread/h70wnb0jqod6wbxml5bvbtj5018sh3mj).

Additionally it looks like DAGs that hit the `dagrun_timeout` _skip_ rather than failing, which then doesn't trigger our failure alerting.

What we really want is [`execution_timeout` at the task level](https://airflow.apache.org/docs/apache-airflow/stable/concepts/tasks.html#timeouts). This would allow us to have fine-grain control of task execution _and_ set the tasks to failed rather than skipped once they're run (thus triggering failure alerting).

## Additional context

We should probably avoid using `dagrun_timeout` altogether, but this is a priority for the provider DAGs.

## Resolution

### LINKED ISSUE #349: Improve provider workflow retries
## Description

Presently the provider DAG factory defaults to 4 (total) retries per task:

https://github.com/WordPress/openverse-catalog/blob/49d30ed1043b566350ebf4c8645feefc823675f6/openverse_catalog/dags/common/dag_factory.py#L30

We should make a few adjustments here:

- Retries should be set to `0` for the data pull steps so that if a provider fails, we want it to 1) stop trying so we can investigate and don't immediately run into the same issue and 2) ingest what we've been able to pull up to that point.
- Global retries should be knocked down to `2` (meaning `3` total attempts).
- Retry delay should be set to 5 minutes rather than 15. Currently 4 retries after 15 minutes means that we won't get error reporting until a full hour after the task was initially started. This feedback loop should be a bit tighter & shorter (while still allowing for retries). 

## Additional context

This work will likely need to come after #285 since the task flow will be modified by that PR.

## Implementation

AetherUnbound:
## Fixes

Fixes #348 by @AetherUnbound, fixes #349 by @AetherUnbound

## Description

This PR tweaks a few settings regarding task retries and timeouts for the typical ingestion DAGs. By "typical ingestion DAGs", I mean the dags that are run daily or monthly and ingest new data, either from a particular date range or for all of time.

Retries for all ingestion DAG related task have been reduced from 4 attempts total to 3. The time between retries has also been reduced from 15 minutes to 5 minutes.

Retries for the data pull steps have been disabled completely. This is to prevent cases where an ingestion fails, retries, gets to the same place (usually hours later) only to fail again. Failing immediately gives the subsequent loading steps the opportunity to ingest data that's already been collected, it notifies the team via Slack, and prevents unnecessary API calls to the provider.

I've changed the `dagrun_timeout` variable to the `execution_timeout` variable. More discussion on this can be found in #348, but the short of it is that `execution_timeout` gives us granular control on the data pull task, and allows the opportunity to ingest data that's already been collected even if the task times out. I've also added a base `execution_timeout` of 1 hour to all other tasks within the DAG. This will prevent the loader steps from hanging, potentially indefinitely. With these guards in place, a DAG should never run more than a few hours over the `execution_timeout` of the ingestion task. We can also further tighten this loop by increasing the maximum number of DAGs which can run at once while constraining the ingestion tasks to use [pools](https://airflow.apache.org/docs/apache-airflow/stable/concepts/pools.html).

## Testing Instructions

Most of these changes are likely best observed in production. If you want to test the retry changes, alter a provider script to fail and then observe that it fails and loads any available data immediately.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/dag_factory.py
+++ b/openverse_catalog/dags/common/dag_factory.py
@@ -82,6 +82,7 @@ DAG_DEFAULT_ARGS = {
     "email_on_retry": False,
     "retries": 2,
     "retry_delay": timedelta(minutes=5),
+    "execution_timeout": timedelta(hours=1),
     "on_failure_callback": slack.on_failure_callback,
 }
 DATE_RANGE_ARG_TEMPLATE = "{{{{ macros.ds_add(ds, -{}) }}}}"
@@ -140,7 +141,7 @@ def create_provider_api_workflow(
     schedule_string: str = "@daily",
     dated: bool = True,
     day_shift: int = 0,
-    dagrun_timeout: timedelta = timedelta(hours=12),
+    execution_timeout: timedelta = timedelta(hours=12),
     doc_md: str = "",
     media_types: Sequence[str] = ("image",),
 ):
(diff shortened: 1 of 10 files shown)


openverse_catalog/dags/common/dag_factory.py
```diff
@@ -158,7 +158,10 @@ def create_provider_api_workflow(
             python_callable=_push_output_paths_wrapper,
             op_kwargs=pull_kwargs,
             depends_on_past=False,
-            execution_timeout=dagrun_timeout if dated else None,
```
stacimc, line 161: It seems reasonable to me that we'd want the timeout across the board, but I'm curious why it was previously applied only for dated data pulls? 🤔

AetherUnbound: I'm not sure! This is [how it was prior to my interactions with the codebase](https://github.com/WordPress/openverse-catalog/blob/7ee62451343ee04a895290b23df9d9c79703dc74/openverse_catalog/dags/util/dag_factory.py#L77-L81) and I wanted to preserve behavior when I refactored this.

stacimc: Mysterious! 😄 At any rate I don't see any harm in adding the timeout to the undated runs. Timeouts can always be adjusted.

stacimc (approved):
This looks good as far as my understanding goes 😄 Thank you for the excellent description, your reasoning seems sound to me! I tested the retry changes locally and all looks well 👍

openverse_catalog/dags/common/dag_factory.py
AetherUnbound: Exactly! I think while we're activating the DAGs, we'll probably need to go back and adjust some of the timeouts (for instance, I know that Freesound will likely take quite a while to process).

zackkrida: This makes sense to me too FWIW 😄

zackkrida (approved):
lgtm! thanks for explaining the motivation so well and adding new documentation. there's fewer code changes here than I anticipated.

Issue #366, opened 2022-02-18
Write current buffer to disk on failure


AetherUnbound:
## Problem

See #357

If a data pull task fails (or is stopped) before 100 items are processed, no data will be written to disk and all records will be lost. Even in the case of a typical data pull, the media store writes the buffer to disk after 100 items. This means that any failure that occurs after a disk write but before the next disk write will result in lost data.

## Description

We should look into making the provider script always flush the buffer on failure. This could be accomplished with a context manager or some similar scope mechanism. Due to the way Airflow works with "Mark as failed" and "Mark as success", we may also need to capture incoming signals and handle them explicitly (otherwise Python may just exit immediately).

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

PR #299, opened 2021-12-09
Remove unnecessary logging.basicConfig calls

AetherUnbound:
## Fixes

Fixes #199 by @AetherUnbound

## Description

As described in #199, these calls are redundant and unnecessary for the DAG/test files. They are useful to keep in files that are likely to be run as individual scripts from the command line, so those ones have been retained.

## Testing Instructions

`just test` and `just run airflow list dags` should both behave normally.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #369, opened 2022-02-18
Add slack message on TSV load complete

### LINKED ISSUE #356: TSV loader completion slack message
## Description

Once #285 is done, it would be great to have a slack message once a TSV loading has completed for a given provider.

This could be done a couple of ways. One approach could be using the [`on_success_callback`](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/index.html#airflow.models.BaseOperator) parameter for the `load_from_s3` step of the DAG. The callback receives the Airflow context as arguments ([these values are made available](https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html#variables)), which can be used to build up a Slack notification that provides sufficient context.

I think it may be easier to have the Slack message as a separate task which runs only on the successful completion of the `load_from_s3` task. This would allow us to pull in [XComs](https://airflow.apache.org/docs/apache-airflow/stable/concepts/xcoms.html) from upstream tasks to provide more information in the Slack message. It would be great to have the following pieces of information as part of the message:

* Provider
* Media type
* Duration of the data pull task
* Number of records upserted

The best location for related functions is probably under a `dags/common/loader/reporting.py` (or similarly named) module. This function could then be added to the DAG using a [PythonOperator](https://airflow.apache.org/docs/apache-airflow/stable/howto/operator/python.html).

### Provider

We should be able to pass this in directly from the DAG, since we're defining it for each one here:

https://github.com/WordPress/openverse-catalog/blob/42b8d78778b242a6bd844755e33d818d6ba0d206/openverse_catalog/dags/common/dag_factory.py#L135

### Media type

This one may be a bit trickier since the ingestion DAGs can have multiple media types. Since each media type has its own loader, each one should probably have its own success message. I believe it can also be handed in by the DAG though, since we're iterating over the media types:

https://github.com/WordPress/openverse-catalog/blob/42b8d78778b242a6bd844755e33d818d6ba0d206/openverse_catalog/dags/common/dag_factory.py#L164

### Duration of the data pull task

This is a bit tougher. We can probably add a duration counter wrapper to the [data pull call step](https://github.com/WordPress/openverse-catalog/blob/42b8d78778b242a6bd844755e33d818d6ba0d206/openverse_catalog/dags/common/dag_factory.py#L78). This value could be pushed via XComs and retrieved by the slack message task. 

Airflow also records task duration in its metastore (example from the UI [here](https://airflow.apache.org/docs/apache-airflow/stable/ui.html#task-duration)), so we could also query the metastore and retrieve the information that way.

It's _possible_ that the Airflow context may provide an object or objects we can use to access the information. We may have to look into that further, as it could be the simplest way to get the duration.

### Number of records upserted

This one will be the hardest. It's possible that the [upsert command](https://github.com/WordPress/openverse-catalog/blob/42b8d78778b242a6bd844755e33d818d6ba0d206/openverse_catalog/dags/common/loader/sql.py#L247-L257) would return the number of rows affected ([since `DbApiHook::run` returns the execution results](https://github.com/apache/airflow/blob/main/airflow/hooks/dbapi.py#L209)). We'll have to do some testing to see if that's actually the case. In that case, we could just bubble up that value into the XComs and retrieve it in the message step.

Alternatively, we could have the `copy_to_s3` step report the number of lines present in the TSV as an XCom, and retrieve _that_ from the slack message.

Either would give us an accurate number of the affected records.

## Implementation

stacimc:
## Fixes

Fixes #356 by @AetherUnbound 

## Description

This PR adds a Slack message that is triggered when TSV loading is completed for a particular provider. It does this by adding a new `report_load_completion` task which runs if the `load_from_s3` task completes successfully.

The message is logged to the Task log, and is also sent as a Slack message to the notifications channel if in the production environment (or if the `slack_message_override` env variable is turned on)

**Currently opened against #285 for ease of testing as it is dependent on those changes. Will be rebased onto main after #285 is merged.**

### Notes/Todo:
* The `pull_data` duration has a few sticking points:
    * The `pull_data` task pulls all the data for _all media types_; there isn't a great way (as far as I can see) to determine how long it took to pull data for an individual type. The _total time for all media_ will have to be reported for each  media type.
    * Although Airflow clearly tracks[ task duration](https://airflow.apache.org/docs/apache-airflow/1.10.2/ui.html#task-duration), I wasn't able to come up with an easy way of accessing this data 🤔 Instead I'm "manually" determining the duration in the pull_data wrapper.
        * When we test the DAGs in Airflow, we generally don't wait for the `pull_data` task to fully resolve, and instead manually set it to `success` after a few 100 lines have been written. This sends a SIGINT to the task, such that my duration calculation is not run or broadcast when testing this way.

## Testing Instructions

* Run a provider workflow in Airflow, manually setting the `pull_data` task to `successful` after a few 100 lines have been written.
    * Make sure that the remaining tasks complete successfully
    * You should see a `report_load_completion` task _for each media type in your provider_
    * Check the logged info in the task log. Note that duration is not reported accurately if you stopped the task early.

* Set the ~`force_slack_alert`~ `slack_message_override` environment variable and set up a Slack connection. Re-run the provider workflow. This time a slack notification should also be sent.
* Run the provider workflow a third time, but manually set the `pull_data` task to `failed`. The `report_load_completion` task should be in the `upstream_failed` state, and the Slack message should not send.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/dag_factory.py
+++ b/openverse_catalog/dags/common/dag_factory.py
@@ -268,7 +268,7 @@ def create_provider_api_workflow(
                     task_id="report_load_completion",
                     python_callable=reporting.report_completion,
                     op_kwargs={
-                        "provider_name": dag_id.replace("_workflow", ""),
+                        "provider_name": provider_name,
                         "media_type": media_type,
                         "duration": XCOM_PULL_TEMPLATE.format(
                             pull_data.task_id, "duration"
--- a/openverse_catalog/dags/common/dag_factory.py
+++ b/openverse_catalog/dags/common/dag_factory.py
@@ -273,6 +273,9 @@ def create_provider_api_workflow(
                         "duration": XCOM_PULL_TEMPLATE.format(
                             pull_data.task_id, "duration"
                         ),
+                        "record_count": XCOM_PULL_TEMPLATE.format(
+                            load_from_s3.task_id, f"{media_type}_record_count"
+                        ),
                     },
                 )
                 drop_loading_table = PythonOperator(
--- a/openverse_catalog/dags/common/dag_factory.py
+++ b/openverse_catalog/dags/common/dag_factory.py
@@ -65,7 +65,7 @@ from airflow.operators.python import PythonOperator
 from airflow.utils.task_group import TaskGroup
 from airflow.utils.trigger_rule import TriggerRule
 from common import slack
-from common.loader import loader, s3, sql
+from common.loader import loader, reporting, s3, sql
 
 
 logger = logging.getLogger(__name__)
@@ -255,6 +255,14 @@ def create_provider_api_workflow(
                         "identifier": identifier,
                     },
                 )
+                report_load_completion = PythonOperator(
+                    task_id="report_load_completion",
+                    python_callable=reporting.report_completion,
+                    op_kwargs={
+                        "provider_name": dag_id.replace("_workflow", ""),
+                        "media_type": media_type,
+                    },
+                )
                 drop_loading_table = PythonOperator(
                     task_id="drop_loading_table",
                     python_callable=sql.drop_load_table,
@@ -265,7 +273,8 @@ def create_provider_api_workflow(
                     },
                     trigger_rule=TriggerRule.ALL_DONE,
                 )
-                [create_loading_table, copy_to_s3] >> load_from_s3 >> drop_loading_table
+                [create_loading_table, copy_to_s3] >> load_from_s3
+                load_from_s3 >> report_load_completion >> drop_loading_table
 
             pull_data >> load_data
 
--- a/openverse_catalog/dags/common/dag_factory.py
+++ b/openverse_catalog/dags/common/dag_factory.py
@@ -54,6 +54,7 @@ https://github.com/creativecommons/cccatalog/issues/334)
 import inspect
 import logging
 import os
+import time
 from datetime import datetime, timedelta
 from typing import Callable, Dict, List, Optional, Sequence
 
@@ -126,9 +127,17 @@ def _push_output_paths_wrapper(
         ti.xcom_push(key=f"{media_type}_tsv", value=store.output_path)
 
     logger.info("Running provider function")
+
+    start_time = time.perf_counter()
     # Not passing kwargs here because Airflow throws a bunch of stuff in there that none
     # of our provider scripts are expecting.
-    return func(*args)
+    data = func(*args)
+    end_time = time.perf_counter()
+
+    duration = end_time - start_time
+    ti.xcom_push(key="duration", value=duration)
+
+    return data
 
 
 def create_provider_api_workflow(
@@ -261,6 +270,9 @@ def create_provider_api_workflow(
                     op_kwargs={
                         "provider_name": dag_id.replace("_workflow", ""),
                         "media_type": media_type,
+                        "duration": XCOM_PULL_TEMPLATE.format(
+                            pull_data.task_id, "duration"
+                        ),
                     },
                 )
                 drop_loading_table = PythonOperator(
--- a/openverse_catalog/dags/common/dag_factory.py
+++ b/openverse_catalog/dags/common/dag_factory.py
@@ -274,7 +274,7 @@ def create_provider_api_workflow(
                             pull_data.task_id, "duration"
                         ),
                         "record_count": XCOM_PULL_TEMPLATE.format(
-                            load_from_s3.task_id, f"{media_type}_record_count"
+                            load_from_s3.task_id, "return_value"
                         ),
                     },
                 )
--- a/openverse_catalog/dags/common/dag_factory.py
+++ b/openverse_catalog/dags/common/dag_factory.py
@@ -289,7 +289,7 @@ def create_provider_api_workflow(
                     trigger_rule=TriggerRule.ALL_DONE,
                 )
                 [create_loading_table, copy_to_s3] >> load_from_s3
-                load_from_s3 >> report_load_completion >> drop_loading_table
+                load_from_s3 >> [report_load_completion, drop_loading_table]
 
             pull_data >> load_data
 
--- a/openverse_catalog/dags/common/loader/loader.py
+++ b/openverse_catalog/dags/common/loader/loader.py
@@ -40,10 +40,13 @@ def load_from_s3(
     media_type,
     tsv_version,
     identifier,
+    ti,
 ):
     sql.load_s3_data_to_intermediate_table(
         postgres_conn_id, bucket, key, identifier, media_type
     )
-    sql.upsert_records_to_db_table(
+    record_count = sql.upsert_records_to_db_table(
         postgres_conn_id, identifier, media_type=media_type, tsv_version=tsv_version
     )
+
+    ti.xcom_push(key=f"{media_type}_record_count", value=record_count)
--- a/openverse_catalog/dags/common/loader/loader.py
+++ b/openverse_catalog/dags/common/loader/loader.py
@@ -40,13 +40,11 @@ def load_from_s3(
     media_type,
     tsv_version,
     identifier,
-    ti,
 ):
     sql.load_s3_data_to_intermediate_table(
         postgres_conn_id, bucket, key, identifier, media_type
     )
-    record_count = sql.upsert_records_to_db_table(
+    # Returns record count
+    return sql.upsert_records_to_db_table(
         postgres_conn_id, identifier, media_type=media_type, tsv_version=tsv_version
     )
-
-    ti.xcom_push(key=f"{media_type}_record_count", value=record_count)
--- a/openverse_catalog/dags/common/loader/reporting.py
+++ b/openverse_catalog/dags/common/loader/reporting.py
@@ -13,7 +13,9 @@ logging.getLogger(common.urls.__name__).setLevel(logging.WARNING)
 def report_completion(
     provider_name,
     media_type,
+    duration,
 ):
     logger.info("Load complete")
     logger.info(provider_name)
     logger.info(media_type)
+    logger.info(duration)
--- a/openverse_catalog/dags/common/loader/reporting.py
+++ b/openverse_catalog/dags/common/loader/reporting.py
@@ -16,7 +16,7 @@ def report_completion(provider_name, media_type, duration, record_count):
 *Provider*: `{provider_name}`
 *Media Type*: `{media_type}`
 *Number of Records Upserted*: {record_count}
-*Duration of data pull task**: {duration}
+*Duration of data pull task*: {duration}
 
 * Duration includes time taken to pull data of all media types.
 """
--- a/openverse_catalog/dags/common/loader/reporting.py
+++ b/openverse_catalog/dags/common/loader/reporting.py
@@ -12,6 +12,11 @@ def report_completion(provider_name, media_type, duration, record_count):
     Messages are only sent out in production and if a Slack connection is defined.
     In all cases the data is logged.
     """
+
+    # This happens when the task is manually set to `success` in Airflow before
+    # completing.
+    duration = "_No data_" if duration == "None" else duration
+
     message = f"""
 *Provider*: `{provider_name}`
 *Media Type*: `{media_type}`
--- a/openverse_catalog/dags/common/loader/reporting.py
+++ b/openverse_catalog/dags/common/loader/reporting.py
@@ -10,12 +10,9 @@ logger = logging.getLogger(__name__)
 logging.getLogger(common.urls.__name__).setLevel(logging.WARNING)
 
 
-def report_completion(
-    provider_name,
-    media_type,
-    duration,
-):
+def report_completion(provider_name, media_type, duration, record_count):
     logger.info("Load complete")
     logger.info(provider_name)
     logger.info(media_type)
     logger.info(duration)
+    logger.info(record_count)
--- a/openverse_catalog/dags/common/loader/reporting.py
+++ b/openverse_catalog/dags/common/loader/reporting.py
@@ -1,0 +1,19 @@
+import logging
+
+import common
+
+
+logging.basicConfig(
+    format="%(asctime)s - %(name)s - %(levelname)s:  %(message)s", level=logging.INFO
+)
+logger = logging.getLogger(__name__)
+logging.getLogger(common.urls.__name__).setLevel(logging.WARNING)
+
+
+def report_completion(
+    provider_name,
+    media_type,
+):
+    logger.info("Load complete")
+    logger.info(provider_name)
+    logger.info(media_type)
--- a/openverse_catalog/dags/common/loader/reporting.py
+++ b/openverse_catalog/dags/common/loader/reporting.py
@@ -1,6 +1,6 @@
 import logging
 
-from common.slack import send_message, should_send_message
+from common.slack import send_message
 
 
 logger = logging.getLogger(__name__)
@@ -20,7 +20,5 @@ def report_completion(provider_name, media_type, duration, record_count):
 
 * _Duration includes time taken to pull data of all media types._
 """
-    if should_send_message():
-        send_message(message, username="Airflow DAG Load Data Complete")
-
+    send_message(message, username="Airflow DAG Load Data Complete")
     logger.info(message)
--- a/openverse_catalog/dags/common/loader/reporting.py
+++ b/openverse_catalog/dags/common/loader/reporting.py
@@ -18,7 +18,7 @@ def report_completion(provider_name, media_type, duration, record_count):
 *Number of Records Upserted*: {record_count}
 *Duration of data pull task*: {duration}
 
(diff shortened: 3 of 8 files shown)


openverse_catalog/dags/common/dag_factory.py
```diff
@@ -255,6 +264,20 @@ def create_provider_api_workflow(
                         "identifier": identifier,
                     },
                 )
+                report_load_completion = PythonOperator(
+                    task_id="report_load_completion",
+                    python_callable=reporting.report_completion,
+                    op_kwargs={
+                        "provider_name": dag_id.replace("_workflow", ""),
```
AetherUnbound, line 271: I think we can reuse the `provider_name` defined on L197 here!

openverse_catalog/dags/common/loader/loader.py
```diff
@@ -40,10 +40,13 @@ def load_from_s3(
...
 ):
     sql.load_s3_data_to_intermediate_table(
         postgres_conn_id, bucket, key, identifier, media_type
     )
-    sql.upsert_records_to_db_table(
+    record_count = sql.upsert_records_to_db_table(
         postgres_conn_id, identifier, media_type=media_type, tsv_version=tsv_version
     )
+
+    ti.xcom_push(key=f"{media_type}_record_count", value=record_count)
```
AetherUnbound, line 52: I think we can remove the need to use `ti` in this function by just returning `record_count`! By default, the `PythonOperator` pushes the return value of a called function to the XComs already (under the `return_value` key, which is the default when calling `ti.xcom_pull` I believe).

More on that here: https://airflow.apache.org/docs/apache-airflow/stable/concepts/xcoms.html#xcoms

It's not the most pressing change, we could definitely leave as-is. But it prevents us from needing to use Airflow-specific resources in this function at least. Up to you!

openverse_catalog/dags/common/loader/reporting.py
```diff
@@ -0,0 +1,18 @@
+import logging
+
+import common
+
+
+logging.basicConfig(
+    format="%(asctime)s - %(name)s - %(levelname)s:  %(message)s", level=logging.INFO
+)
+logger = logging.getLogger(__name__)
+logging.getLogger(common.urls.__name__).setLevel(logging.WARNING)
```
AetherUnbound, line 10: The logging config steps aren't needed here (more on that in #299). We just need the following:

```suggestion
logger = logging.getLogger(__name__)
```

openverse_catalog/dags/common/loader/sql.py
```diff
@@ -254,7 +254,7 @@ def upsert_records_to_db_table(
           {upsert_conflict_string}
         """
     )
-    postgres.run(upsert_query)
+    return postgres.run(upsert_query, handler=lambda c: c.rowcount)
```
AetherUnbound, line 257: So simple! This is great!

AetherUnbound (commented):
This is an _excellent_ start! I'm excited to see the additional changes you've outlined :smile: :partying_face: 

> When we test the DAGs in Airflow, we generally don't wait for the pull_data task to fully resolve, and instead manually set it to success after a few 100 lines have been written. This sends a SIGINT to the task, such that my duration calculation is not run or broadcast when testing this way.

Oof, we continue to be bit by that! (see: #366) I think #248 will help with that when we get there, this might be something we just need to deal with for now.

openverse_catalog/dags/common/loader/loader.py
stacimc: Whoa, slick! That's great to know, I'll definitely update.

openverse_catalog/dags/common/slack.py
```diff
@@ -220,24 +220,32 @@ def send_message(
...
-        return
+        return False
 
     # Exit early if we aren't on production or if force alert is not set
     environment = Variable.get("environment", default_var="dev")
     force_alert = Variable.get(
         "force_slack_alert", default_var=False, deserialize_json=True
     )
-    if not (environment == "prod" or force_alert):
+    return environment == "prod" or force_alert
```
stacimc, line 240: I pulled this logic out from `on_failure_callback` since we'll want to make sure that the TSV load complete message doesn't fire outside of production. I want to call out that this might be an overuse of the `force_slack_alert` environment variable, using it to force slack _notifications_ as well. Would it be worthwhile to have a separate flag, or I suppose to rename the existing one (`force_slack_message`?)

tests/dags/common/test_dag_factory.py
```diff
@@ -47,11 +47,18 @@ def test_push_output_paths_wrapper(func, media_types, stores):
         ti_mock,
         args=[func_mock, value],
     )
-    assert ti_mock.xcom_push.call_count == len(
-        media_types
-    ), "# of output paths didn't match # of stores expected"
-    for args, store in zip(ti_mock.xcom_push.calls, stores):
```
stacimc, line 53: This was a fun bug, so for anyone interested: this had to be updated from `xcom_push.calls` -> `xcom_push.mock_calls`, or else _calls were actually added_ in the slicing/assertion. Ie after exiting the loop, `mock_calls` was:

```
[call(key='image_tsv', value='/var/workflow_output/None_image_v001_20220218234009.tsv'),
call(key='duration', value=0.00024759999359957874),
call.calls.__getitem__(slice(None, -1, None)),
call.calls.__getitem__().__iter__()]
```
😄

AetherUnbound: Oh jeez, that must have been an awful one to find! :cold_sweat:

openverse_catalog/dags/common/slack.py
AetherUnbound: That's a great call! +1 for having it be a separate function. And agreed, I think `force_slack_message` or maybe even `slack_message_override`? The name is from before #359 when we only had slack alerts.

tests/dags/common/loader/test_reporting.py
```diff
@@ -0,0 +1,26 @@
+from unittest import mock
+
+import pytest
+from common.loader.reporting import report_completion
+
+
+@pytest.fixture(autouse=True)
+def send_message_mock() -> mock.MagicMock:
+    with mock.patch("common.slack.send_message") as SendMessageMock:
+        yield SendMessageMock.return_value
```
AetherUnbound, line 10: Should this just be `yield SendMessageMock`? I think we want to check that the function itself was called, not that the function's return value was. I might be interpreting this wrong

tests/dags/common/loader/test_reporting.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@pytest.mark.parametrize(
+    "should_send_message",
+    [
+        (True,),
+        (False,),
+    ],
+)
+def test_report_completion(should_send_message, send_message_mock):
```
AetherUnbound, line 20: If `autouse` is set to `True` on the fixture above, we don't need to explicitly add `send_message_mock` in the test parameters here.

tests/dags/common/loader/test_reporting.py
```diff
@@ -0,0 +1,26 @@
...
+    with mock.patch("common.slack.send_message") as SendMessageMock:
+        yield SendMessageMock.return_value
+
+
+@pytest.mark.parametrize(
+    "should_send_message",
+    [
+        (True,),
+        (False,),
+    ],
```
AetherUnbound, line 18: For multiple parameters in a single `parametrize`, tuples are necessary. If there's only one parameter though, this can be simplified to: 
```suggestion
    [True, False],
```

tests/dags/common/test_slack.py
```diff
@@ -336,3 +341,19 @@ def test_on_failure_callback_does_nothing_without_hook(http_hook_mock):
...
+@pytest.mark.parametrize(
+    "environment, force_slack_alert, expected_result",
+    [
+        ("dev", False, False),
+        ("dev", True, True),
+        ("prod", False, True),
+        ("prod", True, True),
+    ],
+)
+def test_should_send_message(environment, force_slack_alert, expected_result):
```
AetherUnbound, line 355: Excellent test!

tests/dags/common/test_slack.py
```diff
@@ -336,3 +341,19 @@ def test_on_failure_callback_does_nothing_without_hook(http_hook_mock):
...
+    with mock.patch("common.slack.Variable") as MockVariable:
+        # Mock the calls to Variable.get, in order
+        MockVariable.get.side_effect = [environment, force_slack_alert]
```
AetherUnbound, line 358: This is perfectly fine as-is, just noting for pedagogical reasons:
The other way I've seen something simple like this done is to create a dictionary and then use the `__getitem__` call as the actual function:
```suggestion
        d = {"environment": environment, "force_slack_alert": force_slack_alert}
        MockVariable.get.side_effect = d.__getitem__
```

This makes the test robust to changes in the call order to `Variable`.

AetherUnbound (commented):
This is so close! Just a couple more thoughts!

openverse_catalog/dags/common/loader/reporting.py
```diff
@@ -0,0 +1,26 @@
...
+    Messages are only sent out in production and if a Slack connection is defined.
+    In all cases the data is logged.
+    """
+    message = f"""
+*Provider*: `{provider_name}`
+*Media Type*: `{media_type}`
+*Number of Records Upserted*: {record_count}
+*Duration of data pull task*: {duration}
+
+* Duration includes time taken to pull data of all media types.
```
AetherUnbound, line 21: I might suggest italicizing this to emphasize that it's a note!
```suggestion
* _Duration includes time taken to pull data of all media types._
```

tests/dags/common/loader/test_reporting.py
stacimc: Oh yikes, left if like this after removing another parameter -- that doesn't look great 😆 Updated!

tests/dags/common/loader/test_reporting.py
stacimc: No I think you're right! Fixed, thanks for the catch 👍

openverse_catalog/dags/common/slack.py
stacimc: Oh I like `slack_message_override` even better! Updated, thanks for the context 👍

zackkrida:
Unscientific size comparison:

```bash
# Get the initial size
curl -so /dev/null http://localhost:8000/v1/audio/\?format\=json -w '%{size_download}'
21583

# Generate all waveform data
> curl http://localhost:8000/v1/audio/\?format\=json | jq '.results[].waveform' | xargs -L1 open

# Get the new size
❯ curl -so /dev/null http://localhost:8000/v1/audio/\?format\=json -w '%{size_download}'
201138
```

So about 9.3x bigger.

AetherUnbound (approved):
This looks great! Awesome work :smile: 

I think it might also be good to add some documentation to the top of `slack.py` regarding the `slack_message_override` Variable, how to set it & what its purpose is, etc. If you want to throw it in this PR you can, otherwise we can make a separate issue for it and follow up with another PR :slightly_smiling_face:

stacimc:
> I think it might also be good to add some documentation to the top of slack.py regarding the slack_message_override Variable

@AetherUnbound Agreed, and done! Sorry to add something last minute but this made me realize that we should move the `should_send_message` check into `send_message` so that _all_ messages are gated, and that folks don't have to remember to make that check themselves for notifications in the future.

AetherUnbound:
> Sorry to add something last minute but this made me realize that we should move the `should_send_message` check into `send_message` so that _all_ messages are gated, and that folks don't have to remember to make that check themselves for notifications in the future.

No worries! And I was thinking about that too - sometimes I like being able to send slack messages while testing something with ipython locally for instance, so I thought it might be okay to not leave it gated. On the flip side though, I could always just change the variable in the DB and test that way :smile: good call!

openverse_catalog/dags/common/slack.py
```diff
@@ -271,4 +278,6 @@ def on_failure_callback(context: dict) -> None:
 *Log*: {ti.log_url}
 {exception_message}
 """
-    send_message(message, username="Airflow DAG Failure")
+    send_message(
+        message, username="Airflow DAG Failure", http_conn_id=SLACK_ALERTS_CONN_ID
```
AetherUnbound, line 282: Oh good catch!

AetherUnbound (approved):
:rocket:

openverse_catalog/dags/common/dag_factory.py
```diff
@@ -126,9 +127,17 @@ def _push_output_paths_wrapper(
...
+
+    start_time = time.perf_counter()
     # Not passing kwargs here because Airflow throws a bunch of stuff in there that none
     # of our provider scripts are expecting.
-    return func(*args)
+    data = func(*args)
+    end_time = time.perf_counter()
+
+    duration = end_time - start_time
+    ti.xcom_push(key="duration", value=duration)
```
zackkrida, line 138: til about [xcoms](https://airflow.apache.org/docs/apache-airflow/stable/concepts/xcoms.html), really nice feature.

openverse_catalog/dags/common/loader/loader.py
```diff
@@ -44,6 +44,7 @@ def load_from_s3(
     sql.load_s3_data_to_intermediate_table(
         postgres_conn_id, bucket, key, identifier, media_type
     )
-    sql.upsert_records_to_db_table(
+    # Returns record count
+    return sql.upsert_records_to_db_table(
```
zackkrida, line 48: So Airflow automatically sends an xcom with a key of `return_value`? That's great to know 😅

openverse_catalog/dags/common/dag_factory.py
AetherUnbound: They are so useful!! I discuss it a bit more in the issue (#356) - they're great for splitting out more complicated tasks into multiple steps and pass the information between them :smile:

openverse_catalog/dags/common/loader/loader.py
AetherUnbound: [It does!](https://github.com/WordPress/openverse-catalog/pull/369#discussion_r810315410) Not super obvious though haha

AetherUnbound:
@obulat @krysal - we'd like to deploy this tomorrow, any chance y'all could review?

zackkrida (approved):
LGTM! This was easy to test and I learned a bit more about Airflow that I failed to explore previously.

My run of the WordPress DAG:

image

I know we haven't _explicitly_ confirmed that the duration logic is working, but I'm happy for us to revise that in a subsequent PR if we get this working in production and notice large discrepancies between the notifications and the logs and/or airflow UI regarding durations.

openverse_catalog/dags/common/loader/reporting.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    Send a Slack notification when the load_data task has completed.
+    Messages are only sent out in production and if a Slack connection is defined.
+    In all cases the data is logged.
+    """
+    message = f"""
+*Provider*: `{provider_name}`
+*Media Type*: `{media_type}`
+*Number of Records Upserted*: {record_count}
+*Duration of data pull task*: {duration}
```
zackkrida, line 19: Are there other scenarios where this could be _None_ besides the task being marked as done manually? Only asking because if there _aren't_ perhaps we could use a more verbose default instead of "None", something like "Task was ended early" or something (better than that, it's late for me haha)

stacimc: Sorry, took me awhile to get my environment ready to test -- yeah, I think something like this would be good. I believe this should only happen when the task was ended manually, since it needs to have been interrupted but still in a `success` state for this task to even run.

I think `duration` won't be falsy, though, it'll be the actual string `None` returned by xcom.

openverse_catalog/dags/common/dag_factory.py
zackkrida: I just found [_your comment_ ](https://github.com/WordPress/openverse-catalog/pull/369#discussion_r810315410) which mentions everything I noticed, ha 😅

openverse_catalog/dags/common/loader/reporting.py
AetherUnbound: Oh shoot, seriously?? :sob: I could have sworn I've gotten Airflow to handle Python objects as XComs appropriately... maybe we need [`render_template_as_native_obj` set to `True`?](https://airflow.apache.org/docs/apache-airflow/stable/concepts/operators.html#rendering-fields-as-native-python-objects)

AetherUnbound: OH NOW I REMEMBER - we need to turn on [XCom pickling!](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#enable-xcom-pickling) I'll make an issue for it :smile:

Issue #205, opened 2021-10-01
Consolidate provider workflows using dynamic DAGs


AetherUnbound:
## Current Situation

It seems that many of our `<provider>_workflow.py` DAGs include very similar DAG definitions. In many cases, the files are almost exactly identical and differ by only a few paramters.

## Suggested Improvement

We can move the provider parameters into [dataclasses](https://docs.python.org/3/library/dataclasses.html), which would allow us to define one python object per provider. We could build up a list of these objects, and iterate over them to generate all of the DAGs necessary.

Since the DAGs themselves reference provider API modules, we can probably have those defined as strings and use `importlib` to import the necessary modules at DAG generation time.

## Benefit

This would significantly reduce redundant code, as we would only have to define the DAG logic once and could reuse it for every new provider.

## Additional context

I gave a talk about doing something similar at [Airflow Summit](https://airflowsummit.org/sessions/2021/dataclasses-as-pipeline-definitions/)

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

PR #397, opened 2022-03-05
Add data refresh to Airflow

### LINKED ISSUE #353: Data refresh orchestration DAG
## Problem

Data refreshes are currently triggered manually, and previously they were triggered by a cron job on the ingestion server EC2 instance.

## Description

We should leverage Airflow to set up a regularly scheduled data refresh. The ingestion server has improved status reporting now, so we may not need Airflow to monitor the state of the data refresh. Data refreshes cannot occur simultaneously though, so it might be possible/preferred to have Airflow manage that constraint. We can use [Sensors](https://airflow.apache.org/docs/apache-airflow/stable/concepts/sensors.html) to poll the ingestion server for data refresh status in order to determine state. (We'll want to make sure our sensor mode is left at `poke`, since that will ensure a pool slot is reserved for the running sensor). 

This work is currently blocked because the ingestion server doesn't have its own DNS record.

## Additional context

We'll want to make the reference to the ingestion server an Airflow HTTP connection environment variable similar to the Slack connections: https://github.com/WordPress/openverse-catalog/blob/cb19f839e96de7ae1a55e8b7dc82a7d2bf5588e8/env.template#L53

This will also let us point to a local ingestion server for testing.

## Implementation

stacimc:
## Fixes

Fixes #353

## Description

The aim of this PR is to move data refreshes (which currently are triggered manually) into Airflow, where they can be scheduled regularly and easily tracked. The main complication to keep in mind is that **data refreshes cannot occur simultaneously**, so ideally we want Airflow to enforce that for us.

### Reviewer Guide
This is a big PR so here's a breakdown of how to look at the files!

- `env.template`: _Templated changes to new environment variables_
- `common/sensors/single_run_external_dags_sensor.py`: _The new custom Sensor that we're using to make sure that the data refresh DAGs don't run simultaneously. Read the explanation below for more details._
- `tests/dags/common/sensors/test_single_run_external_dags_sensor.py`: _Tests for the new sensor_
- `dags/data_refresh/dag_factory.py`: _The bulk of the work, this is the dag factory that puts it all together._
- `dags/data_refresh/data_refresh_types.py`: _Defines configuration for the data refresh DAGs (currently just audio and image)_
- `common/constants.py`, `provider_dag_factory.py`: _The provider_dag_factory was just renamed, and some constants were pulled out to be more easily shared_
- `dags/providers/*`, `dags/common/test/test_provider_dag_factory.py`: _Just changes from `provider_dag_factory` being renamed_

The Sensor and DAGs are documented within the code, but it may be helpful to first read through a breakdown of how this PR works and ensures that no two data refreshes can occur at the same time. Feel free to skip to testing instructions if you'd like 😄, but here's the rundown:

Screen Shot 2022-03-04 at 4 18 27 PM

* All tasks in the `data_refresh` DAGs run in a special `data_refresh` pool with a single worker slot. This ensures that no two tasks can run at the same time.
* When the DAG starts, it first runs a `wait_for_data_refresh` task. This is a custom [Sensor](https://airflow.apache.org/docs/apache-airflow/stable/concepts/sensors.html) that takes a list of `external_dag_ids` -- in this case, the ids of all the other data refresh DAGs -- and checks to see whether any of those DAGs are running. A "running" DAG in this context is one that (1) is itself in the 'RUNNING' state and (2) its own `wait_for_data_refresh` task has passed successfully.
    * If none are running, the task completes successfully, and downstream tasks can now run.
    * If any of the dependent DAGs are found to be running, the Sensor sleeps and tries again later. Importantly, when it does so it _suspends itself and releases the worker slot_. This is necessary because otherwise the entire pool would deadlock on the waiting Sensor.
* When a DAG's `wait_for_data_refresh` task passes, it proceeds to `trigger_data_refresh`. This POSTs to the ingestion (aka data refresh) server and kicks off the actual data refresh. It parses a `status_check` url out of the response from the ingestion server and passes it along via XCom.
* Finally the `wait_for_completion` task begins. It polls the `status_check` url until it sees either a failure or that the data refresh has succeeded.
    * This task _doesn't_ need to relinquish the worker slot while sleeping, since the pool _should_ lock while waiting for a data refresh to complete.

Here's how that might look if, for example, the `audio_data_refresh` and `image_data_refresh` DAGs are triggered at the exact same time:

1. The Audio and Image DAGs are both set to 'RUNNING' and each `wait_for_data_refresh` task is queued.
1. Let's say the Audio DAG's wait task gets the worker slot first. It starts `wait_for_data_refresh`, which sees that while the Image DAG is 'RUNNING', its wait task has not succeeded -- so we can continue. The Audio wait task passes with `success`, frees the worker slot, and the Audio `trigger_data_refresh` is enqueued.
2. The Image DAG wait task is already queued, so its `wait_for_data_refresh` task gets the worker slot. It sees that the Audio DAG is running _and its wait task has succeeded, meaning the Audio refresh is now in progress_. It suspends itself and enters the `up_for_reschedule` state, freeing the worker slot.
3. The Audio DAG `trigger_data_refresh` takes the slot and runs successfully.
4. Audio DAG `wait_for_completion` takes the worker slot, polling the ingestion server until the refresh finally completes. The entire Audio DAG is set to `success`.
5. The Image DAG `wait_for_refresh` runs again. This time it sees that the Audio DAG isn't in the running state, so it passes and the remaining Image data refresh tasks can be triggered 🎉 

Note that the same logic works if there's a failure somewhere in the Audio DAG, because Image's `wait` task will pass if it sees that Audio is in the `failed` DAG state.

### Alternatives considered

Briefly illustrated here mostly to demonstrate why they don't work/weren't selected:

* A two step DAG using a pool with one slot, ie `trigger_data_refresh` --> `wait_for_completion`
    * This falls short because of the race condition where DAG_A's trigger step runs but DAG_B's trigger step grabs the slot right before DAG_A's waiting step. The same issue applies if we invert the order, polling the ingestion server to make sure it has no tasks running before triggering a refresh (`wait_for_ingestion_server` --> `trigger_data_refresh`). We'd need to set some kind of Semaphore variable, which Sensors aren't really built to do.
* Just having a single data refresh DAG with tasks for each media type
    * This would be _easy_, but it doesn't allow us to do cool things like (1) switch on refreshes for different media types independently, and (2) put different media types on different refresh schedules. 

## Testing Instructions

Some pre-reqs to test 😄:

* Set up an Airflow connection to the ingestion server. To hit your local server (recommended), you can set this in your `openverse-catalog/.env` (Note that it MUST be escaped/prepended like this. Use `172.17.0.1` instead of `host.docker.internal` on linux):
`AIRFLOW_CONN_DATA_REFRESH="http://http%3A%2F%2Fhost.docker.internal%3A8001"`
* For testing purposes, you'll want to set the `DATA_REFRESH_POKE_INTERVAL` environment variable in your `.env`, or else you'll have to wait 15 minutes for the second DAG to start. I set mine to `5`, so the task gets rescheduled every 5 seconds.
* Run `just up`
* Create the special `data_refresh` pool that the refresh DAGs will run in. This is done through the Airflow UI:
    * In the main menu at the top of the page, navigate to Admin > Pools
    * Click the `+` Icon to add a new Pool
    * Name it `data_refresh` and set the `Slots` to `1`.
    * Save
* Make sure that you have the API running locally, since we need to hit it to actually start the refresh! 😄  

Now you should be good to go! You should see `audio_data_refresh` and `image_data_refresh` in the DAG list in the Airflow UI at `localhost:9090`.

To test:
* Open the Airflow UI in two separate browser windows. We want to be able to trigger one DAG and then the other very quickly.
* Trigger the Audio DAG
* Immediately trigger the Image DAG

It's easiest to observe from the `Tree` view for each DAG (url like: `http://localhost:9090/tree?dag_id=audio_data_refresh&root=`). You should see that the `wait_for_data_refresh` task for whichever DAG you started first succeeds, while the other enters the `deferred` state. Watch the `trigger` step run for the first DAG, and watch `wait_for_completion` run (you can look at its logs to verify that it's polling the status check endpoint).

When the first DAG fully completes, observe that the `wait_for_data_refresh` task for the second DAG tries again and passes. This may not happen _instantly_ -- you may have to wait a minute for the deferred task to be picked up again. Once it passes, the entire DAG should complete.

As a bonus, you can also log into the actual ingestion server's container and watch its logs. You should see be able to see the refreshes actually running on the server 😄 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


openverse_catalog/dags/common/dag_factory.py
```diff
@@ -427,3 +434,140 @@ def _build_ingest_operator_list_list(
...
+            dag=dag,
+        )
+
+        data_refresh_post_data = {"model": media_type, "action": "INGEST_UPSTREAM"}
+
+        # Trigger the data refresh by hitting the `/task` endpoint on the data
+        # refresh server. A successful response will include the `status_check`
+        # url used to check on the status of the refresh, which we should pass
+        # to the next task via XCom.
+        trigger_data_refresh = SimpleHttpOperator(
```
AetherUnbound, line 543: LOVE it when we can just use Airflow built-ins :grin:

openverse_catalog/dags/common/sensors.py
```diff
@@ -0,0 +1,93 @@
...
+
+    external_dag_ids: A list of dag_ids that you want to wait for
+    :param check_existence: Set to `True` to check if the external task exists (when
+        external_task_id is not None) or check if the DAG to wait for exists (when
+        external_task_id is None), and immediately cease waiting if the external task
+        or DAG does not exist (default value: False).
+    """
+
+    # TODO: Not sure if this is necessary
+    template_fields = ["external_dag_ids"]
```
AetherUnbound, line 27: This is only necessary if we want Airflow to evaluate the `external_dag_ids` variable like it does with some other fields, [populating in macro values](https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html). Allowing this to be templated would enable us to do something like `external_dag_ids = ["{{ ds }}", "{{ ts_nodash }}"]`, etc. I agree, I don't think it's necessary here!

openverse_catalog/dags/common/sensors.py
```diff
@@ -0,0 +1,93 @@
...
+class ExternalDAGsSensorAsync(BaseSensorOperator):
+    """
+    Waits for a list of different DAGs to not be running, freeing up
+    a worker slot while it's waiting.
+
+    external_dag_ids: A list of dag_ids that you want to wait for
+    :param check_existence: Set to `True` to check if the external task exists (when
+        external_task_id is not None) or check if the DAG to wait for exists (when
+        external_task_id is None), and immediately cease waiting if the external task
+        or DAG does not exist (default value: False).
```
AetherUnbound, line 23: Hmm, I think we may need to update some of these references, since we're referring to DAGs instead of tasks :sweat_smile:

openverse_catalog/dags/common/sensors.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # TODO: Not sure if this is necessary
+    template_fields = ["external_dag_ids"]
+
+    # TODO: I don't understand what operator extra links are, and whether it's
+    # necessary to implement this.
+    # @property
+    # def operator_extra_links(self):
+    #     """Return operator extra links"""
+    #     return [ExternalDAGsSensorLink()]
```
AetherUnbound, line 34: We certainly could, but I think it would just point to the other data refresh DAGs :thinking: might be a bit redundant, since each one will point to all the others?

openverse_catalog/dags/common/sensors.py
```diff
@@ -0,0 +1,93 @@
...
+        check_existence: bool = False,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self.external_dag_ids = external_dag_ids
+        self.check_existence = check_existence
+        self._has_checked_existence = False
+
+    @provide_session
+    def execute(self, context, event=None, session=None):
```
AetherUnbound, line 49: I think `BaseSensorOperator::poke` is the function we want to override!

https://github.com/apache/airflow/blob/78490f86bfab195cc98a7b700c26c24a1aa2eb17/airflow/sensors/base.py#L149-L154

openverse_catalog/dags/common/sensors.py
```diff
@@ -0,0 +1,93 @@
...
+        self.log.info("%s DAGs are in the running state", count_running)
+        if count_running == 0:
+            return True
+
+        # If there are running DAGs, trigger deferral of the Sensor so that
+        # the worker slot will be freed and the pool does not become
+        # deadlocked waiting on this task.
+        self.defer(
+            trigger=TimeDeltaTrigger(
+                timedelta(seconds=5)
```
AetherUnbound, line 77: I think we'll want this configurable at the operator level!

openverse_catalog/dags/common/sensors.py
```diff
@@ -0,0 +1,93 @@
...
+        )
+
+        self.log.info("%s DAGs are in the running state", count_running)
+        if count_running == 0:
+            return True
+
+        # If there are running DAGs, trigger deferral of the Sensor so that
+        # the worker slot will be freed and the pool does not become
+        # deadlocked waiting on this task.
+        self.defer(
```
AetherUnbound, line 75: I'm not sure that we need to use a deferrable sensor here :open_mouth: I think the [`reschedule` mode](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/sensors/base/index.html) + `poke_interval` will work just fine! That means we also won't have to spin up the `airflow triggerer`.

openverse_catalog/dags/common/dag_factory.py
```diff
@@ -427,3 +434,140 @@ def _build_ingest_operator_list_list(
...
+    start_date:        datetime.datetime giving the
+                       first valid execution_date of the DAG.
+    schedule_string:   string giving the schedule on which the DAG should
+                       be run.  Passed to the airflow.dag.DAG __init__
+                       method.
+    execution_timeout: datetime.timedelta giving the amount of time a given data
+                       pull may take.
+    doc_md:            string which should be used for the DAG's documentation markdown
+    """
+    # TODO verify if schedule_interval is None, DAG only runs when triggered manually
```
AetherUnbound, line 514: I can confirm this is true :smile:

AetherUnbound (commented):
This is an incredible start! You've really picked up Airflow quick, I'm so happy to see it :smile:

I have a couple of general suggestions I'd like to make:

## Sensor type

As I said in some of the comments, I think we'll be fine with a non-async `reschedule`d sensor. This would reduce the need to spin up another process within the docker container and allows us to simplify a few things. Happy to discuss this more if there's an aspect of the deferred sensor I might be missing!

## Code location

I agree that `dags/data_refresh` is probably the best place for all the operational code! (save for the sensor). We could do `dags/data_refresh/dag_factory.py`, and if that's best then I think we should probably rename `common/dag_factory` to `common/provider_dag_factory` or something similar (doesn't have to be part of this PR).

## DAG Factory

The way the previous DAG factory was set up is kind of an anti-pattern :sweat_smile: we have a ticket to address this that's pending some other work: #205. I think this is the _perfect_ opportunity to create a small setup similar to the one described in #205 but for the data refresh. What I'm thinking is that we'll have one file, maybe `dags/data_refresh/data_refresh_types.py` (names are hard) that will have a list of `dataclasses`. Each `dataclass` will define the different attributes that a data refresh DAG might need. Here's some pseudocode:

```python
# data_refresh_types.py

from dataclasses import dataclass, field

@dataclass
class DataRefresh:
    media_type: str
    start_date: datetime = datetime(2020, 1, 1)  # default but can be overridden
    execution_timeout: timedelta = timedelta(hours=24)
    dag_id: str = field(init=False)

    def __post_init__(self):
        self.dag_id = f"{self.media_type}_data_refresh"


DATA_REFRESH_CONFIGS = [
    DataRefresh(
        media_type="image",
        execution_timeout=timedelta(days=3),
    ),
    DataRefresh(media_type="audio"),
]
```

Then, in our DAG factory, we can just iterate over that list to create our [dynamic DAGs](https://www.astronomer.io/guides/dynamically-generating-dags/):

```python
# dag_factory.py
from data_refresh_types import DATA_REFRESH_CONFIGS


def create_dag(data_refresh: DataRefresh, other_dag_ids: Sequence[str]):
    # Any other info we want to configure can be tacked onto the DataRefresh dataclass,
    # or just set at the top of this file.
    dag = DAG(dag_id=data_refresh.dag_id, schedule_interval=None, default_args=DEFAULT_ARGS))

    with dag:
        ...


all_dag_ids = {refresh.dag_id for refresh in DATA_REFRESH_CONFIGS}

for data_refresh in DATA_REFRESH_CONFIGS:
    # Construct a set of all other DAG IDs
    other_dag_ids = all_dag_ids - {data_refresh.dag_id}
    globals()[data_refresh.dag_id] = create_dag(data_refresh, other_dag_ids)
```

This lets us have **one** file that contains all the configuration for each data refresh DAG, and another file which only defines the DAG factory!

That's a lot to take in I'm sure :sweat_smile: My [talk at Airflow Summit last year](https://airflowsummit.org/sessions/2021/dataclasses-as-pipeline-definitions/) goes over these concepts if more context would be helpful!

openverse_catalog/dags/common/sensors.py
stacimc: What!! The docs had me convinced that Deferrable operators were the only solution, but I must have missed this! Amazing, this lets us write off the question about dealing with the triggerer. Thank you 😄

stacimc:
> This means that N DAGs will be set to RUNNING simultaneously, and whichever DAG grabs the spot first will initiate the wait_for_refresh step. 

Thanks for this context -- I incorrectly assumed that a DAG is not `RUNNING` until it has a running task. That could definitely be a problem 🤔 One of the things I liked about checking DAG status was that the states are so simple relative to the tasks 😄 But yes, I can definitely extend the logic a little in the Sensor to look at the wait task states as well and that should handle this. I also prefer separating out the tasks to collapsing.

Really good catch, thank you!

openverse_catalog/dags/common/dag_factory.py
```diff
@@ -427,3 +434,140 @@ def _build_ingest_operator_list_list(
...
+        raise AirflowException("Error triggering data refresh.")
+
+    logger.info(
+        f"Data refresh done with {data['percent_completed']}% \
+        completed."
+    )
+    return True
+
+
+def create_data_refresh_dag(
```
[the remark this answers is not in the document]
stacimc: I wouldn't want to see the docstring that encompasses both the provider dag_factory _and_ the data refresh ones 😆

openverse_catalog/dags/common/dag_factory.py
```diff
@@ -427,3 +434,140 @@ def _build_ingest_operator_list_list(
...
+    dag_id: str,
+    media_type: str,
+    external_dag_ids: List[str],
+    start_date: datetime = datetime(1970, 1, 1),
+    default_args: Optional[Dict] = None,
+    schedule_string: Optional[str] = None,
+    # TODO is this a good timeout?
+    execution_timeout: timedelta = timedelta(hours=12),
```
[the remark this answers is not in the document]
stacimc: That makes sense -- good idea, updated for image!

openverse_catalog/dags/common/sensors.py
stacimc: Thank you for the context!

openverse_catalog/dags/common/sensors.py
stacimc: Oops 😄 Updated with the Sensor's new logic.

openverse_catalog/dags/common/sensors.py
stacimc: I think this is how I had to do it with Deferrable sensors -- I can't believe I missed you can free the worker slot with 'normal' sensors! Thanks, changed it back to `poke` and added `reschedule` mode. So much nicer not to have to deal with Triggerers, amazing 🎉 😄

openverse_catalog/dags/common/sensors.py
stacimc: Now that we're using `reschedule` mode, I've changed this to just use the `poke_interval`!

I think this is what you mean but just confirming -- I didn't think it was necessary to be configurable _per media type_ since it seems like a single interval for the data refresh DAGs would be fine.

I'm keeping it at 5 seconds for now. The thought is that since this is running inside a pool with one slot, at worst the waiting `wait_for_data_refresh` tasks can each run in between the three steps of whatever DAG is actually running (ie they can't interrupt `wait_for_completion` once it starts). Having a short poke interval means that when the `wait_for_completion` finally passes, you won't need to wait a long time for the next data refresh to kick off.

Does that logic make sense to you? 🤔

openverse_catalog/dags/common/sensors.py
stacimc: Done and works like a charm 🥳

openverse_catalog/dags/common/sensors.py
AetherUnbound: I think the deferrable stuff is pretty new and (frankly) _very_ cool, but it solves a problem we (fortunately) don't yet have haha

openverse_catalog/dags/common/sensors.py
AetherUnbound: It does! I agree that we won't need this different per media type, but since this was essentially filling in the role of `poke_interval`, I think it would have made sense to pull up as an operator level variable (similar to `external_dag_ids`) so it could be different were we to use this sensor in other contexts.

I will say, though, that I think we can actually have it be 5 minutes rather than 5 seconds. With the data refreshes taking on the order of days (rather than minutes), I don't think we need that fine a resolution once `wait_for_completion` is done. Additionally, rescheduling the task every 5 seconds puts a lot of unnecessary pressure on the Airflow scheduler (see #309 for some other context). So I think we could have a `poke_interval` of 5 minutes at least, if not something like 10 or 15 even.

stacimc:
Okay, pushed up a couple of changes!

**Re the DAG Factory:**
> What I'm thinking is that we'll have one file, maybe dags/data_refresh/data_refresh_types.py (names are hard) that will have a list of dataclasses. Each dataclass will define the different attributes that a data refresh DAG might need. 

This is exactly what I had in mind! 💯 I like the suggestion to collapse the dataclasses into a single configuration file -- all of this will make it even easier to set up a data refresh for a new media type, and protects against incorrectly configuring the DAG dependencies ✨ 

I've kept the Sensor under `common`. I think it's just generic enough that it might be useful in another scenario.

**Re the [race condition](https://github.com/WordPress/openverse-catalog/pull/397#issuecomment-1061941455):**
I've updated the Sensor and believe it should handle this scenario. All we need to do (I think 😄) is check that a DAG is `running` _and_ that its own `wait` task has succeeded.
* If the `wait` task fails the entire DAG fails, so this case can never occur.
* The `wait` task cannot be `running` because there is a single worker slot. We could check for this anyway to support some hypothetical generic case, but I don't think this Sensor is actually useful outside of a pool with one slot.
* All other task states indicate that this DAG has not begun its data refresh.

There's also lots of documentation updates. Tests are unfortunately still incoming, so I'm leaving this as a draft for now.

openverse_catalog/dags/common/sensors.py
stacimc: Ah, I was under the impression that because of the singular pool, the task wouldn't continually reschedule while waiting -- but I think I'm confusing that with something I read about the triggers in the original implementation with the Deferred Sensor. That would be a lot of overhead. 10-15 min seems totally reasonable given the length of the refreshes.

This does make testing more annoying, but we can always edit it locally while testing. Maybe an opportunity for a testing env variable?

AetherUnbound: I think a testing env var makes sense here! I would suggest an Airflow Variable as well but we're not likely to update that value in dev more than once I'd imagine.

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+XCOM_PULL_TEMPLATE = "{{{{ ti.xcom_pull(task_ids='{}', key='{}') }}}}"
+DATA_REFRESH_POOL = "data_refresh"
+
+
+# Filter for the `trigger_data_refresh` task, used to grab the endpoint needed
+# to poll for the status of the triggered data refresh. This information will
+# then be available via XCom in the downstream tasks.
+def response_filter_data_refresh(response):
```
AetherUnbound, line 69: Sorry, I might have been unclear - I think this is a perfect docstring:
```suggestion
def response_filter_data_refresh(response):
    """
    Filter for the `trigger_data_refresh` task, used to grab the endpoint needed
    to poll for the status of the triggered data refresh. This information will
    then be available via XCom in the downstream tasks.
    """
```

Same for the function below! Type annotations would also be useful on these two functions.

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -0,0 +1,186 @@
...
+# determine whether the task can complete.
+def response_check_wait_for_completion(response):
+    data = response.json()
+
+    if data["active"]:
+        # The data refresh is still running. Poll again later.
+        return False
+
+    if data["error"]:
+        raise AirflowException("Error triggering data refresh.")
```
AetherUnbound, line 84: Would it make sense to include the contents of `data["error"]` here, or any error info we have? Or is that just True/False?

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -0,0 +1,186 @@
...
+    Required Arguments:
+
+    data_refresh:     dataclass containing configuration information for the
+                      DAG
+    external_dag_ids: list of ids of the other data refresh DAGs. This DAG
+                      will not run concurrently with any dependent DAG.
+    """
+    dag = DAG(
+        dag_id=data_refresh.dag_id,
+        default_args=data_refresh.default_args,
```
AetherUnbound, line 120: I think we may want to merge the data refresh's default arguments with the `DAG_DEFAULT_ARGS` present in the original DAG factory file. That provides some useful defaults, like the owner, `depends_on_past`, and failure handlers for tasks (e.g. slack notifications). I think it might be time to move that out into `common.constants`, since we're likely to reuse that in the future. This could look like:

```suggestion
        default_args={**constants.DAG_DEFAULT_ARGS, **data_refresh.default_args},
```

openverse_catalog/dags/data_refresh/data_refresh_types.py
```diff
@@ -0,0 +1,54 @@
...
+    execution_timeout: datetime.timedelta giving the amount of time a given data
+                       pull may take.
+    doc_md:            string which should be used for the DAG's documentation markdown
+    """
+
+    dag_id: str = field(init=False)
+    media_type: str
+    start_date: datetime = datetime(2020, 1, 1)
+    execution_timeout: timedelta = timedelta(hours=24)
+    schedule_string: Optional[str] = None
```
AetherUnbound, line 40: Maybe we could have this be `schedule_interval`, so it's consistent with the value we're setting in the DAG?

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -0,0 +1,186 @@
...
+        start_date=data_refresh.start_date,
+        schedule_interval=data_refresh.schedule_string,
+        catchup=False,
+        doc_md=data_refresh.doc_md,
```
AetherUnbound, line 124: I think we can ditch the per-DAG markdowns, and instead use the awesome docstring you defined at the top of this module!
```suggestion
        doc_md=__doc__,
```

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -0,0 +1,186 @@
...
+        tags=[f"data_refresh: {data_refresh.media_type}"],
```
AetherUnbound, line 125: Since we'll only have one media type per data refresh DAG, I think it might make sense to have a generic `data_refresh` tag instead (since we can search by individual DAG already)
```suggestion
        tags=["data_refresh"],
```

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -0,0 +1,186 @@
...
+
+    with dag:
+        # Wait to ensure that no other Data Refresh DAGs are running.
+        wait_for_data_refresh = ExternalDAGsSensor(
+            task_id="wait_for_data_refresh",
+            external_dag_ids=external_dag_ids,
+            check_existence=True,
+            pool=DATA_REFRESH_POOL,
+            dag=dag,
+            poke_interval=5,
```
AetherUnbound, line 136: I think this may need to be a bit longer, since it's in seconds:
```suggestion
            poke_interval=60 * 5,
```

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -0,0 +1,186 @@
...
+        # Wait for the data refresh to complete.
+        wait_for_completion = HttpSensor(
+            task_id="wait_for_completion",
+            http_conn_id="data_refresh",
+            endpoint=XCOM_PULL_TEMPLATE.format(
+                trigger_data_refresh.task_id, "return_value"
+            ),
+            method="GET",
+            response_check=response_check_wait_for_completion,
+            pool=DATA_REFRESH_POOL,
```
AetherUnbound, line 168: Since `pool` is used in every task in this DAG, we can actually just add it to the DAG's `default_args`! That will apply it automatically to every task.

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -0,0 +1,186 @@
...
+all_data_refresh_dag_ids = {refresh.dag_id for refresh in DATA_REFRESH_CONFIGS}
+
+for data_refresh in DATA_REFRESH_CONFIGS:
+    # Construct a set of all data refresh DAG ids other than the current DAG
+    other_dag_ids = all_data_refresh_dag_ids - {data_refresh.dag_id}
+
+    # Generate the DAG for this config, dependent on all the others
+    globals()[data_refresh.dag_id] = create_data_refresh_dag(
+        data_refresh, other_dag_ids
+    )
```
AetherUnbound, line 186: :star_struck:

AetherUnbound (commented):
The new structure looks great! I'm going to try this out today and I'll drop any additional comments after that if I have any :smile:

openverse_catalog/dags/data_refresh/dag_factory.py
stacimc: Nice! I _think_ it's actually only important that the `wait` task be in the data refresh pool, actually, since that should gate the other tasks. But I think I'll keep it on all of them as it's a little easier to reason about. I'll add this to clean it up 🎉

openverse_catalog/dags/data_refresh/dag_factory.py
stacimc: Totally! I will update this for sure -- this was in seconds for testing purposes. I was thinking about [using an environment variable](https://github.com/WordPress/openverse-catalog/pull/397#discussion_r822946132) here before updating it, so we could increase to 10-15 minutes for production but be able to easily shorten that when testing 🤔

openverse_catalog/dags/data_refresh/dag_factory.py
stacimc: Oh you're totally right -- copy/paste issue from the provider DAG, in this context a simple `data_refresh` tag will be way more useful!

openverse_catalog/dags/data_refresh/dag_factory.py
stacimc: Yeah I'd like to report something more useful here, but we just get back True/False in the response. For the time being I think we can probably rely on the ingestion server's own [error reporting](https://github.com/WordPress/openverse-api/blob/main/ingestion_server/ingestion_server/tasks.py#L130)

openverse_catalog/dags/data_refresh/dag_factory.py
stacimc: Ah yep-- just missed these as I was moving comments around. These should definitely be docstrings, will do.

openverse_catalog/dags/data_refresh/dag_factory.py
AetherUnbound: I agree, it should only be necessary there. But the actual task check times aren't too long, so I think having them all share the pool is fine!

openverse_catalog/dags/data_refresh/dag_factory.py
stacimc: I just saw your comment about the [testing variable](https://github.com/WordPress/openverse-catalog/pull/397#discussion_r823042545) 😄 I'll go ahead and add it, and update this for a more production-minded value!

openverse_catalog/dags/common/sensors.py
```diff
@@ -1,38 +1,26 @@
...
+from airflow.models import DagBag, DagModel, DagRun, TaskInstance
 from airflow.sensors.base import BaseSensorOperator
-from airflow.triggers.temporal import TimeDeltaTrigger
 from airflow.utils.session import provide_session
 from airflow.utils.state import State
-from sqlalchemy import func
 
 
-class ExternalDAGsSensorAsync(BaseSensorOperator):
+class ExternalDAGsSensor(BaseSensorOperator):
```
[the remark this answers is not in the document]
stacimc: Oo, I really struggled to come up with something more specific while also still generic -- I like `SingleRunExternalDAGsSensor`!

openverse_catalog/dags/data_refresh/dag_factory.py
AetherUnbound: Ah drat, that sounds good though! You're right, we should be getting sufficient information elsewhere.

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -0,0 +1,186 @@
...
+These DAGs initiate a data refresh for a given media type and awaits the
+success or failure of the refresh. Importantly, they are also configured to
+ensure that no two data refresh DAGs can run concurrently, as required by
+the server.
+
+A data refresh occurs on the data refresh server in the openverse-api project.
+This is a task which imports data from the upstream Catalog database into the
+API, copies contents to a new Elasticsearch index, and makes the index "live".
+This process is necessary to make new content added to the Catalog by our
+provider DAGs available on the frontend.
```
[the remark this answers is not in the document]
stacimc: I thought about this but then got in my head about broken links (especially considering open issues to rename the ingestion server). The README is very helpful, though -- I can link it and make a note to update it.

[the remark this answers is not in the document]
AetherUnbound: Very fair point! Thanks :smile:

stacimc:
Feedback addressed, and I also added a first pass at tests for the new Sensor. Unfortunately the `test_retries_if_running_dags_with_completed_sensor_task` has been flaky for me when running locally. It sporadically fails at different assertions.

The test creates a DAG with a Sensor task, starts a `running` DagRun for the task and a successful TaskInstance for the DagRun's sensor task. Then it starts a test Sensor which is dependent on the `running` task; I expect this to report that 1 DAG is in the running state and then reschedule itself.

Sometimes the test fails at the [assertion](https://github.com/WordPress/openverse-catalog/pull/397/commits/90a0fb7a14c497f30e5c9f42e81fb9e9be7f36be#diff-05899f051a9e57827ca7b9c3efd5b180748a6a0f493865c749f8bb1fa7a0299aR161) that the test TaskInstance succeeded, even though this code is taken from the Airflow examples for[ unit testing](https://airflow.apache.org/docs/apache-airflow/stable/best-practices.html?highlight=testing%20dags#unit-tests) (scroll down to "unit tests for custom operators"). Sometimes these assertions all pass, but the Sensor finds 0 dags in the running state. 🤔

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -42,38 +44,45 @@
...
+    then be available via XCom in the downstream tasks.
+    """
     status_check_url = response.json()["status_check"]
     return urlparse(status_check_url).path
 
 
 # Response check to the `wait_for_completion` Sensor. Processes the response to
 # determine whether the task can complete.
-def response_check_wait_for_completion(response):
+def response_check_wait_for_completion(response: Response):
```
AetherUnbound, line 81:
```suggestion
def response_check_wait_for_completion(response: Response) -> bool:
```

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -115,25 +124,29 @@ def create_data_refresh_dag(data_refresh: DataRefresh, external_dag_ids: Sequenc
     external_dag_ids: list of ids of the other data refresh DAGs. This DAG
                       will not run concurrently with any dependent DAG.
     """
+    default_args = {
+        **DAG_DEFAULT_ARGS,
+        **data_refresh.default_args,
+        "pool": DATA_REFRESH_POOL,
```
AetherUnbound, line 130: :fire:

tests/dags/common/test_sensors.py
```diff
@@ -0,0 +1,197 @@
...
+
+class TestExternalDAGsSensor(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        Pool.create_or_update_pool("test_pool", slots=1, description="test pool")
+
+    def tearDown(self):
+        with create_session() as session:
+            session.query(DagRun).delete()
+            session.query(TaskInstance).delete()
```
AetherUnbound, line 74: Is this conflicting at all with the `clean_db` function? Even if not, is it redundant?

tests/dags/common/test_sensors.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def test_fails_if_external_dag_does_not_exist(self):
+        with pytest.raises(AirflowException):
```
AetherUnbound, line 77: I think for all the tests that use `pytest.raises`, it might be a good idea to use the [`match` parameter](https://docs.pytest.org/en/6.2.x/reference.html#pytest-raises) to narrow down the specific exception being raised!

tests/dags/common/test_sensors.py
```diff
@@ -0,0 +1,197 @@
...
+        execution_date=DEFAULT_DATE,
+        state=dag_state,
+        run_type=DagRunType.MANUAL,
+    )
+
+
+class TestExternalDAGsSensor(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        Pool.create_or_update_pool("test_pool", slots=1, description="test pool")
```
AetherUnbound, line 69: This test pool name could probably be a class variable that's used in all the tests

AetherUnbound (commented):
Just a few more thoughts, this is so close!! 

One other thing I didn't realize initially: I think we may want to change the folder structure a bit here. It's a solid idea to have one file per sensor, rather than one file with all the sensors. I think we'll want to change the current file to `common.sensors.singlerunexternaldags` or something similar (same for the test file).

These tests look perfect!

AetherUnbound:
I'll also note that we realized the flaky tests were being caused by `pytest-xdist` :upside_down_face: we're looking into using the [`--dist=loadgroup` option](https://pypi.org/project/pytest-xdist/#running-tests-across-multiple-cpus) to help prevent issues like this in the future!

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -42,38 +44,45 @@
...
 
-XCOM_PULL_TEMPLATE = "{{{{ ti.xcom_pull(task_ids='{}', key='{}') }}}}"
 DATA_REFRESH_POOL = "data_refresh"
 
 
-# Filter for the `trigger_data_refresh` task, used to grab the endpoint needed
-# to poll for the status of the triggered data refresh. This information will
-# then be available via XCom in the downstream tasks.
-def response_filter_data_refresh(response):
+def response_filter_data_refresh(response: Response):
```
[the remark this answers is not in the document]
stacimc: Oh I completely missed adding in the types 🤦‍♀️ Thanks for the catch (again)!

tests/dags/common/test_sensors.py
stacimc: This can be taken out -- one of the things I added while trying to diagnose the flakiness, not needed :)

tests/dags/common/test_sensors.py
stacimc: This was a great suggestion and actually revealed that the two exception tests weren't really testing the underlying exception (they were both raising an AirflowException because the test TI wasn't mapped to a DAG).

The bad news is that I am having a tough time getting `test_fails_if_external_dag_missing_sensor_task` to work correctly now that I've updated it. It's failing to find the test DAG (with the incorrectly named sensor task) in the DagBag. I can't figure out why this would be any different from the Dags that are configured in the other tests 🤔

stacimc:
Rebased to include the changes from #409, so the tests should no longer be flaky! Unfortunately I've encountered a _new_ [problem](https://github.com/WordPress/openverse-catalog/pull/397#discussion_r826491976) with the tests, working on diagnosing 👀

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -0,0 +1,197 @@
...
+    default_args = {
+        **DAG_DEFAULT_ARGS,
+        **data_refresh.default_args,
+        "pool": DATA_REFRESH_POOL,
+    }
+    dag = DAG(
+        dag_id=data_refresh.dag_id,
+        default_args=default_args,
+        start_date=data_refresh.start_date,
+        schedule_interval=data_refresh.schedule_interval,
```
AetherUnbound, line 136: One thing that came up while I was testing - we'll want to add `max_active_runs=1` here to ensure that only one DAG run per type is running at any given time.

AetherUnbound (commented):
This is seriously awesome. I've learned so much myself during this PR process! I tested this locally, and it seems to work great. My only concern is around multiple runs per media-type. Super unlikely, but I changed my schedule interval locally to every 5 minutes for testing and a number of DAG runs from the same DAG ended up running at the same time.

One other note for testers using linux, the address for the ingestion server needs to be `172.17.0.1` (rather than `host.docker.internal`).

Do you mind adding one more test case? To ensure this will work for >2 DAGs going forward, it would be great to have a case where there are 3 running DAGs, and we ensure that one gets to completion before any others are triggered.

openverse_catalog/dags/data_refresh/dag_factory.py
stacimc: Good idea!

AetherUnbound (approved):
Thanks to some explanation from Staci, I've realized that what I was asking for is 1) functionally covered by the existing unit tests and 2) would actually function as an integration test requiring a DAG to run to completion, which we don't have any of yet and would be difficult to spin up! So as it stands now this looks GREAT and I can't wait to get it going! :partying_face:

sarayourfriend:
Sorry I haven't reviewed this yet, it totally flew under my radar somehow. I'm prioritizing reviewing it today.

openverse_catalog/dags/common/sensors/single_run_external_dags_sensor.py
```diff
@@ -0,0 +1,88 @@
...
+    @provide_session
+    def poke(self, context, session=None):
+        self.log.info(
+            "Poking for DAGs %s ...",
+            self.external_dag_ids,
+        )
+
+        # Check DAG existence only once, on the first execution.
+        if self.check_existence and not self._has_checked_existence:
+            self._check_for_existence(session=session)
```
sarayourfriend, line 45: nit (and a very minor one at that) about readability for `poke`: I'm not sure if this is just me, but setting `_has_checked_existence` inside of `_check_for_existence` is slightly confusing when reading the overall flow of the `poke` method, unless you dig into `check_for_existence` while reading through `poke`. But it's sort of like a side effect that seems to belong more to `poke` given `poke` itself is the thing commanding the checking :thinking: I suppose if the conditional was moved all into `check_for_existence` (as an early return conditional) then it would make more sense and would reduce the scope of `poke`; then `poke` doesn't concern itself with short-circuiting the check and lets `check_for_existence` handle all the existence checking logic including the short-circuiting.

openverse_catalog/dags/common/sensors/single_run_external_dags_sensor.py
```diff
@@ -0,0 +1,88 @@
...
+                session.query(DagModel).filter(DagModel.dag_id == dag_id).first()
+            )
+
+            if not dag_to_wait:
+                raise AirflowException(f"The external DAG {dag_id} does not exist.")
+
+            if not os.path.exists(dag_to_wait.fileloc):
+                raise AirflowException(f"The external DAG {dag_id} was deleted.")
+
+            refreshed_dag_info = DagBag(dag_to_wait.fileloc).get_dag(dag_id)
```
sarayourfriend, line 64: `DagBag` dear lord :sweat_smile:

openverse_catalog/dags/data_refresh/data_refresh_types.py
```diff
@@ -0,0 +1,53 @@
...
+for each of our media types. This configuration information is used to generate
+the dynamic Data Refresh dags.
+"""
+from dataclasses import dataclass, field
+from datetime import datetime, timedelta
+from typing import Dict, Optional
+
+
+@dataclass
+class DataRefresh:
```
sarayourfriend, line 14: :star_struck:

tests/dags/common/sensors/test_single_run_external_dags_sensor.py
```diff
@@ -0,0 +1,219 @@
...
+        run_id=f"{dag.dag_id}_test",
+        start_date=DEFAULT_DATE,
+        execution_date=DEFAULT_DATE,
+        data_interval=(DEFAULT_DATE, DEFAULT_DATE),
+        state=dag_state,
+        run_type=DagRunType.MANUAL,
+    )
+
+
+class TestExternalDAGsSensor(unittest.TestCase):
```
sarayourfriend, line 72: I wasn't aware you could mix unittest.TestCase classes with pytest stuff, TIL!

sarayourfriend (approved):
The code LGTM. I think I more or less understand it (or at least the contours of it). Just one nit picky comment there.

I ran this locally but kept running into issues, I think they're just configuration problems on my end but I don't really understand what the problem could be. It could also just be that I don't know how to use Airflow.

I kept misconfiguring the airflow connection URL, so let's just say I've thoroughly tested _that_ edge case :sweat_smile: 

Once I got it configured correctly it worked great.

The URL encoded `http` url with also an `http://` prefix is baffling to me and frustrating to work with. Is there any way to sort of obviate the problem there and URL encode it ourselves in the code instead of having to carefully maintain URL encoded stuff by hand in the environment file?

env.template
```diff
@@ -65,6 +65,10 @@ AIRFLOW_CONN_EMR_TEST=emr://?host=http://s3:5000
 EMR_CONN_ID=emr_empty
 EMR_TEST_CONN_ID=emr_test
 
+# Connection to the Ingestion Server, used for managing data refreshes. (This must be URL encoded,
+# with a pre-pended https://, e.g. "https://https%3A%2F%2Fhost.docker.internal%3A8001"))
```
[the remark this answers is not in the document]
stacimc: It is intentional and yes, very counter-intuitive due to some Airflow weirdness: https://airflow.apache.org/docs/apache-airflow-providers-http/stable/operators.html#simplehttpoperator

I'm not sure there's anything we can do about this. FWIW the same pre-pending business is also used for our [Slack connection](https://github.com/WordPress/openverse-catalog/blob/main/env.template#L50).

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -0,0 +1,198 @@
+"""
+# Data Refresh DAG Factory
```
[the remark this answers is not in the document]
stacimc: I'm not sure if there are plans for that in the Catalog either 🤔 At the moment I'm just mirroring the docs for the `provider_dag_factory`, although it could be worthwhile to update them across files. I could try moving this into the doc for the factory method, but it's already quite long.

openverse_catalog/dags/common/sensors/single_run_external_dags_sensor.py
stacimc: [DagBag.bag_dag](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/dagbag/index.html#airflow.models.dagbag.DagBag.bag_dag) 😆

env.template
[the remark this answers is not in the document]
AetherUnbound: Yea, this is unfortunately how Airflow handles connection environment variables. It uses the scheme to determine which connection type this refers to (e.g. `s3` is the `S3Hook`, `postgres` is the `PostgresHook`, `https` is the `HttpHook`). These have to be present in the environment when Airflow starts up, so we would have to add a shim in the docker entrypoint in order to do any conversions of this. It's useful to know that this is a particular pain point - we _can_ always add that shim logic, particularly if we continue to add HTTP connection types to the environment file.

openverse_catalog/dags/data_refresh/dag_factory.py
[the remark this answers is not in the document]
AetherUnbound: This docstring is fed directly into the DAG's `doc_md` field, gets turned into rendered markdown in the Airflow UI. Here's an example from the TSV loader DAG:

image

I think having the docstring as the DAG's documentation is useful, but for DAG factories I've also generated _different_ markdown documentation that includes specifics to the generated DAG (rather than having the same documentation for all DAGs generated by the factory).

tests/dags/common/sensors/test_single_run_external_dags_sensor.py
AetherUnbound: Pytest is super robust this way, thankfully! I think it would be feasible to turn this into function-based tests (this was based off of Airflow's own tests which are class structured) but that's totally up to @stacimc :slightly_smiling_face:

env.template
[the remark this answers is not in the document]
sarayourfriend: Yes please!

tests/dags/common/sensors/test_single_run_external_dags_sensor.py
sarayourfriend: I guess unittest.TestCase has some nice stuff Staci is using like assertLogs and such. `setUp` I guess would just be an auto-used fixture?

env.template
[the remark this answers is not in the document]
AetherUnbound: Ticket made :slightly_smiling_face: #433

tests/dags/common/sensors/test_single_run_external_dags_sensor.py
AetherUnbound: Yea that's exactly what we could do! Although `assertLogs` would be just a hair harder to change, though the [`caplog` fixture](https://docs.pytest.org/en/6.2.x/logging.html#caplog-fixture) makes that easier.

stacimc: I was surprised too! I was using Airflow's own tests as a reference to test the Sensor 😰 and found that we were mixing unittest.TestCase with pytest in some ingestion server tests 🤯

stacimc:
Rebased and re-tested! Thanks for making the follow-up ticket for the connection variables, @AetherUnbound 💯

PR #418, opened 2022-03-17
Fix invalid license urls from Finnish Museum API

### LINKED ISSUE #392: Finnish Museum `pull_data` freezes and times out
## Description

The Finnish Museum provider DAG was recently turned on. The `pull_image_data` task begins running but then becomes stuck while getting the object list for the `0/Museovirasto` building. It eventually times out, incomplete.

The logs:

```
[2022-02-25, 00:46:13 UTC] {finnish_museums.py:37} INFO - Begin: Finnish museum provider script
[2022-02-25, 00:46:13 UTC] {finnish_museums.py:39} INFO - Obtaining Images of building 0/Suomen kansallismuseo/
[2022-02-25, 00:46:33 UTC] {finnish_museums.py:63} WARNING - No more retries 
[2022-02-25, 00:46:33 UTC] {finnish_museums.py:39} INFO - Obtaining Images of building 0/Museovirasto/
```

## Reproduction

Check out the `pull_image_data` logs for the Finnish Museum workflow in Airflow. You may also try re-running the DAG.

## Expectation
At minimum if the DAG has encountered an error, I would expect it to report the error and fail the task.

## Resolution

stacimc:
## Fixes

Fixes #392 by @stacimc

## Description

The Finnish Museum DAG's `pull_data` task was running without errors, but would eventually time out without having pulled any records (or logged any issues). It turns out this is because the API returns `license_url`s for each image linking to the Finnish version of the license _deed_, so like this:

```
http://creativecommons.org/licenses/by/4.0/deed.fi
```
instead of what our license validator logic expects, which is:
```
http://creativecommons.org/licenses/by/4.0
```

All the records were discarded as having an invalid license, so nothing got written to disk and the DAG appeared to be doing nothing.

This PR fixes the issue by removing the `deed.fi` suffix _within the provider script_. Initially I wanted to update the license validation logic to remove valid suffixes, but I think this will be a much more complex problem to solve for a couple reasons:
* Not all of our expected license urls are the same length (for example, we have `licenses/by/4.0` but also `licenses/by-nd-nc/2.0/jp` and `licenses/publicdomain`
* It might be necessary to check that a suffix is a valid one before chopping
* This needs to be done _efficiently_ because `is_valid_license_info` is called for every record

Since we're only seeing this problem in the Finnish Museum DAG, I think for now it might be over-engineering to try to fix the issue more generally. I'm happy to write an issue to track the possibility of adding this in the future.

This PR also extends the timeout for the DAG, since I personally ran it for 19 hours and verified that it was still pulling results 😄 

## Testing Instructions

`just test`

Run the Finnish Museum DAG and verify that you start seeing lines written to disk. Mark the task as successful and verify they are loaded into the db.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
@@ -136,7 +136,7 @@ def get_license_url(obj):
     # The API returns urls linking to the Finnish version of the license deed,
     # (eg `licenses/by/4.0/deed.fi`), but the license validation logic expects
     # links to the license page (eg `license/by/4.0`).
-    return license_url.removesuffix("/deed.fi")
+    return license_url.removesuffix("deed.fi")
 
 
 def _get_raw_tags(obj):
--- a/openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
@@ -103,7 +103,7 @@ def _process_object_list(object_list):
 
 def _process_object(obj, sub_providers=SUB_PROVIDERS, provider=PROVIDER):
     total_images = 0
-    license_url = obj.get("imageRights", {}).get("link")
+    license_url = get_license_url(obj)
     if license_url is None:
         return None
     foreign_identifier = obj.get("id")
@@ -125,6 +125,18 @@ def _process_object(obj, sub_providers=SUB_PROVIDERS, provider=PROVIDER):
             raw_tags=raw_tags,
         )
     return total_images
+
+
+def get_license_url(obj):
+    license_url = obj.get("imageRights", {}).get("link")
+
+    if license_url is None:
+        return None
+
+    # The API returns urls linking to the Finnish version of the license deed,
+    # (eg `licenses/by/4.0/deed.fi`), but the license validation logic expects
+    # links to the license page (eg `license/by/4.0`).
+    return license_url.removesuffix("/deed.fi")
 
 
 def _get_raw_tags(obj):
--- a/openverse_catalog/dags/common/licenses/licenses.py
+++ b/openverse_catalog/dags/common/licenses/licenses.py
@@ -232,9 +232,6 @@ def is_valid_license_info(license_info: LicenseInfo) -> bool:
         license_path = license_info.url.replace(base_path, "")
         if license_path[-1] == "/":
             license_path = license_path[:-1]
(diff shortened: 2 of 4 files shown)


openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
```diff
@@ -127,6 +127,18 @@ def _process_object(obj, sub_providers=SUB_PROVIDERS, provider=PROVIDER):
     return total_images
 
 
+def get_license_url(obj):
```
AetherUnbound, line 130: Do you mind adding unit tests for this function as well? Or do you feel like it's sufficiently encapsulated in the current testing examples?

AetherUnbound (approved):
Looks great! Thanks for the explanation on why we aren't planning on generalizing this (_for now_). 

I do think we should slap an [`lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache) on the `is_valid_license_info` function. We're already doing that for `get_license_info`:

https://github.com/WordPress/openverse-catalog/blob/3034e31f6f204c641a89f7acbcbb85f026b9c9c3/openverse_catalog/dags/common/licenses/licenses.py#L28

I'll make a separate ticket for that :slightly_smiling_face:

openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
stacimc: I think this is covered in the current example -- we simulate receiving [this example json](https://github.com/WordPress/openverse-catalog/blob/main/tests/dags/providers/provider_api_scripts/resources/finnishmuseums/object_complete_example.json#L28) from the API, which includes the `/deed.fi` suffix on the license url, and then assert that the item added to the ImageStore has the [correctly formatted url](https://github.com/WordPress/openverse-catalog/pull/418/files#diff-17dd91ce4ca74ccedfbf229306a08034f674305c6b5dacbebe8bfc6741728a37R100). Although maybe I could add an additional test to verify it still works if the API returns urls without that suffix

zackkrida (approved):
LGTM! The new test is a nice touch.

obulat:
> This needs to be done efficiently because `is_valid_license_info` is called for every record

I just realized that `is_valid_license_info` is called for every license, and we probably should use `lru_cache` for it.

PR #421, opened 2022-03-18
Enable XCom pickling in Airflow

### LINKED ISSUE #377: Enable XCom pickling
## Problem

XComs by default are inserted into the database as strings. While this is more secure, it can make handling native Python types very difficult.

## Description

We should [enable XCom pickling](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#enable-xcom-pickling) in the Airflow configuration. This will make operations which use upstream XComs easier since we will not have to do string/JSON conversions each time we pull the XCom (e.g. see [this comment](https://github.com/WordPress/openverse-catalog/pull/369#discussion_r813215801)). Particularly as we start to utilize XComs more, those conversions will become very tedious.

This is something we can add to the [Docker Airflow configuration block in the Dockerfile](https://github.com/WordPress/openverse-catalog/blob/69b2eb7aaa2daf54029fac57b3b99405734da7de/docker/airflow/Dockerfile#L16).

## Alternatives

Leave as-is and perform JSON SerDes whenever accessing a non-string XCom.

## Additional context

[Pickling](https://docs.python.org/3/library/pickle.html) serializes/deserializes native Python objects, which introduces an opportunity for remote code execution. Since our Airflow is behind authentication, I believe that the benefit we gain from pickling is worth this extra risk. In any case, **we should always be careful what data we're passing via XComs**. Those values go into the database, so we need to 1) make sure we trust them and 2) be careful not to try and store too much information in an XCom (e.g. passing Pandas DataFrames via XComs is generally considered a Bad Idea, unless one is certain they will be small).

We should also update the string check added in #359 as part of this PR, since the XCom values will once again be `None`. 

## Implementation

stacimc:
## Fixes

Fixes #377 by @aetherunbound

## Description

Enables XCom pickling in the Airflow configuration. This will make it easier to use XCom without having to do string/JSON conversions each time. As a good example, it also updates the TSV loader slack reporting to avoid String comparisons 🎉

## Testing Instructions

Run a provider DAG (I chose Jamendo) and wait until you see that lines are being written to disk. Then manually set the task to `success` -- this will cause the `duration` passed to XCom to be `None`.

Now check the logs for the `report_load_completion` task once it completes, and verify that you see `_No data_` reported for the duration.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


AetherUnbound (approved):
Looks fantastic! Thanks for also updating the completion message workaround we had :partying_face: 

I also tried out the Staten Museum DAG and let it run to completion, the existing XComs performed normally as well!
image
Verified in the database that they're also pickled:
```
airflow> select * from xcom order by timestamp desc limit 4;
+--------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------+-------------------------------+------------------------+------------------------------+------------------------+
| key          | value                                                                                                                                                          | timestamp                     | execution_date         | task_id                      | dag_id                 |
|--------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------+-------------------------------+------------------------+------------------------------+------------------------|
| return_value | \x80049504000000000000004d16b42e                                                                                                                               | 2022-03-18 18:08:08.803303+00 | 2022-02-01 00:00:00+00 | load_image_data.load_from_s3 | staten_museum_workflow |
| s3_key       | \x80049543000000000000008c3f696d6167652f73746174656e5f6d757365756d2f73746174656e736d757365756d5f696d6167655f763030315f32303232303331383138303531372e747376942e | 2022-03-18 18:08:02.835271+00 | 2022-02-01 00:00:00+00 | load_image_data.copy_to_s3   | staten_museum_workflow |
| tsv_version  | \x80049507000000000000008c03303031942e                                                                                                                         | 2022-03-18 18:08:02.82602+00  | 2022-02-01 00:00:00+00 | load_image_data.copy_to_s3   | staten_museum_workflow |
| duration     | \x8004950a000000000000004740644bd5857d24402e                                                                                                                   | 2022-03-18 18:08:00.466257+00 | 2022-02-01 00:00:00+00 | pull_image_data              | staten_museum_workflow |
+--------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------+-------------------------------+------------------------+------------------------------+------------------------+
SELECT 4
Time: 0.016s
```

krysal (approved):
Works great! I tested it with the **WP Photo Directory** script and can confirm it shows the expected output in the logs:

```
*Provider*: `wordpress`
*Media Type*: `image`
*Number of Records Upserted*: 300
*Duration of data pull task*: _No data_
```

I was looking for `duration` in the Airflow UI but it doesn't appear, I guess I would have to let the DAG to finish normally as @AetherUnbound :)
 
CleanShot 2022-03-18 at 16 26 24@2x

PR #442, opened 2022-03-25
Show duplicate record count in completion slack message

### LINKED ISSUE #411: <Replace this with actual title>
## Problem

#395 adds handling for duplicate keys in the load_data task, by stripping records with duplicate keys before upsertion. Currently there isn't an at-a-glance way to tell whether duplicates were encountered/deleted during a DAG run.

## Description

Add a new item to the `report_load_completion` Slack message which reports the number of duplicates found (or omit the item entirely if there were no duplicates).

Note that this isn't as simple as getting the number of items found in the `pull_data` task and subtracting the final count from `load_data`, because there are other reasons a record might be stripped before upsertion (records might be removed in this [cleaning step](https://github.com/WordPress/openverse-catalog/blob/38ee4938afa7794ff56e92cb162ce0348838ef33/openverse_catalog/dags/common/loader/sql.py#L159).

## Alternatives

It's possible to 'eyeball' it by just checking the logs for the `pull_data` task to see what the number of records pulled is, and then looking at the number upserted by `load_data`. (Keeping in mind that this may not be totally precise).

## Implementation

AetherUnbound:
## Fixes

Fixes #411 by @stacimc

Dependent on #434

## Description

This PR adds some count collecting throughout the ingestion process so that we can better track the number of duplicates encountered for a given provider + media type. While we currently handle duplicates thanks to #395, we would still like some reporting on this (that way we can look into providers that have high rates of duplicates).

image

The test functions for these changes are a little unorthodox; I wanted to to test the message generated by multiple media types and make sure that they don't interfere with each other. If the functions are confusing or could use more comments, let me know!

I also added type hints in a few different places to make it easier for us down the road to understand what's happening, especially since the `report_counts_by_media_type` object became more complex in this PR.

## Testing Instructions

`just test` to see the new tests

If you'd like to see a message:
1. Add ` - 20` to the very end of `loader.py` line 50, this will simulate an offset due to duplicates
1. Run the Staten Museum DAG
1. View the logs in the `report_completion` task and ensure they look similar to this:

Screenshot_2022-03-25_11-06-54

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/loader/reporting.py
+++ b/openverse_catalog/dags/common/loader/reporting.py
@@ -39,10 +39,13 @@ def report_completion(provider_name, duration, record_counts_by_media_type):
         duration = humanize_time_duration(duration)
 
     # List record count per media type
-    media_type_reports = "\n".join(
-        f"  - `{media_type}`: {record_count or '_No data_'}"
-        for media_type, record_count in record_counts_by_media_type.items()
-    )
+    media_type_reports = ""
+    for media_type, (loaded, upserted) in record_counts_by_media_type.items():
+        duplicates = loaded - upserted
+        media_type_reports += f"  - `{media_type}`: {upserted or '_No data_'}"
+        if duplicates:
+            media_type_reports += f" _({duplicates} duplicates)_"
+        media_type_reports += "\n"
 
     # Collect data into a single message
     message = f"""
--- a/openverse_catalog/dags/common/loader/reporting.py
+++ b/openverse_catalog/dags/common/loader/reporting.py
@@ -50,7 +50,11 @@ def report_completion(
     media_type_reports = ""
     for media_type, counts in record_counts_by_media_type.items():
         loaded, cleaned, upserted = counts
-        media_type_reports += f"  - `{media_type}`: {upserted or '_No data_'}"
+        if not upserted:
+            upserted_human_readable = "_No data_"
+        else:
+            upserted_human_readable = f"{upserted:,}"
+        media_type_reports += f"  - `{media_type}`: {upserted_human_readable}"
         if any([count is None for count in counts]):
             # Can't make calculation without data
             continue
@@ -59,9 +63,9 @@ def report_completion(
         if duplicates or removed:
             extras = []
             if removed:
-                extras.append(f"{removed} cleaned")
+                extras.append(f"{removed:,} cleaned")
             if duplicates:
-                extras.append(f"{duplicates} duplicates")
+                extras.append(f"{duplicates:,} duplicates")
             if joined := ", ".join(extras):
                 media_type_reports += f" _({joined})_"
         media_type_reports += "\n"
--- a/openverse_catalog/dags/common/loader/reporting.py
+++ b/openverse_catalog/dags/common/loader/reporting.py
@@ -77,8 +77,8 @@ def report_completion(
                 extras.append(f"{cleaned:,} cleaned")
             if duplicates:
                 extras.append(f"{duplicates:,} duplicates")
-            if joined := ", ".join(extras):
-                media_type_reports += f" _({joined})_"
+            if extras:
+                media_type_reports += f" _({', '.join(extras)})_"
         media_type_reports += "\n"
 
     # Collect data into a single message
--- a/openverse_catalog/dags/common/loader/reporting.py
+++ b/openverse_catalog/dags/common/loader/reporting.py
@@ -58,12 +58,11 @@ def report_completion(
         if any([count is None for count in counts]):
             # Can't make calculation without data
             continue
-        removed = loaded - cleaned
-        duplicates = cleaned - upserted
-        if duplicates or removed:
+        duplicates = loaded - cleaned - upserted
+        if duplicates or cleaned:
             extras = []
-            if removed:
-                extras.append(f"{removed:,} cleaned")
+            if cleaned:
+                extras.append(f"{cleaned:,} cleaned")
             if duplicates:
                 extras.append(f"{duplicates:,} duplicates")
             if joined := ", ".join(extras):
--- a/openverse_catalog/dags/common/loader/reporting.py
+++ b/openverse_catalog/dags/common/loader/reporting.py
@@ -49,8 +49,11 @@ def report_completion(
     # List record count per media type
     media_type_reports = ""
     for media_type, (loaded, upserted) in record_counts_by_media_type.items():
-        duplicates = loaded - upserted
         media_type_reports += f"  - `{media_type}`: {upserted or '_No data_'}"
+        if upserted is None or loaded is None:
+            # Can't make calculation without data
+            continue
+        duplicates = loaded - upserted
         if duplicates:
             media_type_reports += f" _({duplicates} duplicates)_"
         media_type_reports += "\n"
--- a/openverse_catalog/dags/common/loader/reporting.py
+++ b/openverse_catalog/dags/common/loader/reporting.py
@@ -18,7 +18,7 @@ TIME_DURATION_UNITS = (
     ("min", 60),
     ("sec", 1),
 )
-RecordCounts = dict[str, tuple[Optional[int], Optional[int]]]
+RecordCounts = dict[str, tuple[Optional[int], Optional[int], Optional[int]]]
 
 
 def humanize_time_duration(seconds: float) -> str:
@@ -48,14 +48,22 @@ def report_completion(
 
     # List record count per media type
     media_type_reports = ""
-    for media_type, (loaded, upserted) in record_counts_by_media_type.items():
+    for media_type, counts in record_counts_by_media_type.items():
+        loaded, cleaned, upserted = counts
         media_type_reports += f"  - `{media_type}`: {upserted or '_No data_'}"
-        if upserted is None or loaded is None:
+        if any([count is None for count in counts]):
             # Can't make calculation without data
             continue
-        duplicates = loaded - upserted
-        if duplicates:
-            media_type_reports += f" _({duplicates} duplicates)_"
+        removed = loaded - cleaned
+        duplicates = cleaned - upserted
+        if duplicates or removed:
+            extras = []
+            if removed:
+                extras.append(f"{removed} cleaned")
+            if duplicates:
+                extras.append(f"{duplicates} duplicates")
+            if joined := ", ".join(extras):
+                media_type_reports += f" _({joined})_"
         media_type_reports += "\n"
 
     # Collect data into a single message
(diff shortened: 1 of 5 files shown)


openverse_catalog/dags/common/loader/reporting.py
```diff
@@ -28,7 +32,11 @@ def humanize_time_duration(seconds):
     return ", ".join(parts)
 
 
-def report_completion(provider_name, duration, record_counts_by_media_type):
+def report_completion(
+    provider_name: str,
+    duration: float | str | None,
```
stacimc, line 37: When is `duration` a string? (Now that pickling is turned on)

openverse_catalog/dags/common/loader/sql.py
```diff
@@ -151,12 +152,15 @@ def load_s3_data_to_intermediate_table(
               'us-east-1'
             );
             """
-        )
+        ),
+        handler=lambda c: c.fetchone()[0],
```
stacimc, line 156: Cool!

openverse_catalog/dags/common/loader/reporting.py
```diff
@@ -15,9 +18,10 @@
     ("min", 60),
     ("sec", 1),
 )
+RecordCounts = dict[str, tuple[Optional[int], Optional[int]]]
```
stacimc, line 21: Ah, the `Optional` here is because we get `None` instead of `0` when 0 rows are returned for either `loaded` or `upserted`, right?

stacimc (approved):
🎉 Looks great! I'll be very happy to see this getting logged 😄

AetherUnbound:
After some discussion with @stacimc and I yesterday, we will ALSO want to add the number of records affected by the cleaning. We found for some providers this can be quite high, so we'd like to know when that happens!

AetherUnbound:
I've updated the code to now show the number of **cleaned** records, as well as the number of duplicates. The logic is a bit more complex as a result - I've been fiddling with it a bunch, so it might not be optimally written. If anyone has thoughts on revising, please let me know!

image

openverse_catalog/dags/common/loader/reporting.py
AetherUnbound: Hmm, it shouldn't be, but I thought I ran into a case where it was :thinking: certainly couldn't hurt to retain the logic that accounts for it being a non-float type...

openverse_catalog/dags/common/loader/sql.py
AetherUnbound: It's kind of annoying that this isn't something immediately available (I could have sworn that `DbApiHook` did this by default before...) but it's nice that we can add it in where necessary!

openverse_catalog/dags/common/loader/reporting.py
AetherUnbound: That's right!

openverse_catalog/dags/common/loader/reporting.py
```diff
@@ -39,10 +47,16 @@ def report_completion(provider_name, duration, record_counts_by_media_type):
...
-    )
+    media_type_reports = ""
+    for media_type, (loaded, upserted) in record_counts_by_media_type.items():
+        media_type_reports += f"  - `{media_type}`: {upserted or '_No data_'}"
+        if upserted is None or loaded is None:
+            # Can't make calculation without data
+            continue
+        duplicates = loaded - upserted
+        if duplicates:
+            media_type_reports += f" _({duplicates} duplicates)_"
```
sarayourfriend, line 58: I'm having a hard time following the logic across all the files, but is `duplicates` coming from the `affected_count` in `_clean_intermediate_table_data`? If so, then labeling them "duplicates" is incorrect, as the `_clean_intermediate_table_data` function also removes rows missing certain required columns (as well as removing duplicates).

If we want the duplicate count specifically, do we need to partition the affected counting against the missing required rows and the duplicates separately?

sarayourfriend (commented):
LGTM! Just one question about whether these are just duplicates or also include the rows removed for missing required columns. Otherwise happy to approve once I understand the distinction going on there :slightly_smiling_face:

openverse_catalog/dags/common/loader/reporting.py
stacimc: We get the `duplicates` by subtracting the number of records that got `upserted` from the number that were `loaded`. We retrieve those counts from XCom; they're pushed [here](https://github.com/WordPress/openverse-catalog/blob/1ace8ea118c9360ec1e0d3469f0988c182b5637c/openverse_catalog/dags/common/loader/loader.py#L51). So the `loaded_count` is the count that comes back from `load_s3_data_to_intermediate_table`, which already [takes into account](https://github.com/WordPress/openverse-catalog/blob/1ace8ea118c9360ec1e0d3469f0988c182b5637c/openverse_catalog/dags/common/loader/sql.py#L160) rows that get cleaned and doesn't count them. So I think it should be good!

@AetherUnbound was testing changes to actually report the count of cleaned rows in addition to duplicates though -- did those changes get pushed or is that going to be a separate PR?

AetherUnbound: Oh my god, I forgot to push :sob: Doing that now, so sorry!

AetherUnbound: I can add some more documentation to the reporting function here to better describe what we're seeing though!

tests/dags/common/loader/test_reporting.py
```diff
@@ -18,11 +18,56 @@ def test_report_completion(should_send_message):
...
+
+
+# This sets up parameterizations for both audio and image simultaneously, in order
+# to test that the statistics are reported accurately independent of each other.
+@pytest.mark.parametrize(
+    "audio_data, audio_expected", _make_report_completion_contents_data("audio")
+)
+@pytest.mark.parametrize(
+    "image_data, image_expected", _make_report_completion_contents_data("image")
+)
```
stacimc, line 59: Very cool! ✨

stacimc (approved):
LGTM and tests well! Very curious to see what kind of numbers we get when this is deployed 👀

openverse_catalog/dags/common/loader/reporting.py
```diff
@@ -28,21 +32,54 @@ def humanize_time_duration(seconds):
...
+            continue
+        duplicates = loaded - cleaned - upserted
+        if duplicates or cleaned:
+            extras = []
+            if cleaned:
+                extras.append(f"{cleaned:,} cleaned")
+            if duplicates:
+                extras.append(f"{duplicates:,} duplicates")
+            if joined := ", ".join(extras):
+                media_type_reports += f" _({joined})_"
```
sarayourfriend, line 81: nit: This packs a lot of logic into this little line! Took me a while to understand that it works. To be fair, I think I'm one of the poor souls that does _not_ like the walrus operator most of the time... but in this case it seems like too much going on, splitting it like this would be the same number of lines but much easier to read, at least in my opinion. Feel free to take it or leave it though!

```suggestion
            if extras:
                media_type_reports += f" _({', '.join(extras)})_"
```

sarayourfriend (approved):
I have to admit that I can't follow the logic that partitions the counts between cleaned and duplicates. I've tried tracing what I can of the xcomms and stuff but I just don't see it.

Is the affected count the "cleaned" count? It still looks like the duplicates are counted in there, in the `_clean_intermediate_table_data` function. I get that the three numbers being used _should_ be able to be used to derive the duplicate count, but it doesn't look like the counting logic for what is being cleaned doesn't already include the duplicate count to me.

Maybe there's a different set of duplicates though? I'm thinking of the duplicates described by this line specifically: https://github.com/WordPress/openverse-catalog/pull/442/files#diff-7393ac84fd49d7f2763f9e4a69f1d0372dbb41fbe351b49d5b15f0cdef36034cR169

I trust the unit tests you wrote though, they make sense to me! But they only cover the printing/math logic of the already retrieved counts, not the counting itself, which is the place that I'm not necessarily understanding.

If I'm totally wrong though and just not seeing something obvious to someone who understands the flow of data through these kinds of things better than me, feel free to fire away! On the other hand, it doesn't seem like it would be a disaster to have the wrong numbers used here for a short period of time anyway as long as we can notice it, I suppose.

stacimc:
> Maybe there's a different set of duplicates though? I'm thinking of the duplicates described by this line specifically: https://github.com/WordPress/openverse-catalog/pull/442/files#diff-7393ac84fd49d7f2763f9e4a69f1d0372dbb41fbe351b49d5b15f0cdef36034cR169

Thanks for calling this out, because it is confusing. The `clean_intermediate_table_data` removes records that have duplicate provider/foreign_id pairs. The `duplicates` we're referring to in the reporting are records that have unique foreign_ids but the same URL ([as described here](https://github.com/WordPress/openverse-catalog/pull/442/files#diff-bafec9830977e7a481a38ad8e3ea799e2bc161967cdca37130fe6f512241b5c5R51)).

Those get stripped out [here](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/loader/sql.py#L258) in the upsert, and it's why that calculation should work. I guess we could also calculate the count in the upsert and pass it through XCom instead of doing it in the reporting step, if that makes it clearer 🤔

openverse_catalog/dags/common/loader/reporting.py
AetherUnbound: Oh, this is much better! I think I was avoiding the `join` call within the f-string because we so frequently have problems with escape characters not being allowed in f-strings.

AetherUnbound:
This is indeed confusing, I think because both of the values we're trying to report could be interpreted as "cleaned" records OR "duplicate" records :disappointed: As Staci linked, I tried to document each of these pieces here: 

https://github.com/WordPress/openverse-catalog/blob/7feba18b5d2abbbb7cdb2d3b5ee993ba8c9fd662/openverse_catalog/dags/common/loader/reporting.py#L46-L55

The reality is that `cleaned` does represent duplicates along provider + foreign ID _as well as_ missing columns. Technically the field we're considering as "duplicates" could be folded into "cleaned" so we end up with just one number, but I think the distinction is helpful for us currently. We might even want to separate the provider + foreign ID & missing column records into distinct numbers so we can better identify the root cause of an issue. But without any real data yet it's hard to know how useful that'll be!

sarayourfriend:
Ah, okay. There are two sets of duplicates.

Perhaps there's some way to make this clearer, either in the message or elsewhere in the code. Is there a "type" of duplicate for each one that could differentiate?

AetherUnbound:
There definitely is. I think I'd like to get this PR in, and then I want to take a stab at turning this into 3 separate metrics that we report. I'm worried that might clutter things, but it's at least worth trying! If we find it's too much we can drop the PR or revert it down the line.

Issue #433, opened 2022-03-21
Make Airflow HTTP connection env vars more readable


AetherUnbound:
## Problem

We are currently using [environment variables to define Airflow connections for local development](https://airflow.apache.org/docs/apache-airflow/stable/howto/connection.html#storing-a-connection-in-environment-variables). This allows us to jettison the database readily while still retaining connection information.

This approach, however, can be quite confusing. HTTP connections require that the connection start with `http://` and then URL encode the rest of the string. This can lead to weird situations[^1] like:

```
export AIRFLOW_CONN_SLACK='http://https%3a%2f%2fhooks.slack.com%2fservices%2f...'
```

or this snippet from our existing env file:

https://github.com/WordPress/openverse-catalog/blob/0e3675ca71f49461382489af8a0bbadfa3b5a4d1/env.template#L50

As [@sarayourfriend pointed out](https://github.com/WordPress/openverse-catalog/pull/397#discussion_r830115653) in #397, this can be difficult and confusing to interpret initially.

[^1]: https://issues.apache.org/jira/browse/AIRFLOW-2910?focusedCommentId=16832468&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16832468

## Description

We should find a way to automatically munge the HTTP connection URLs before booting up Airflow. The munging would URL encode the provided URL, then tack `https://` onto the front.

Since these variables need to exist in the environment _before_ Airflow boots up, we'll likely need to create a shim in the Docker entrypoint for altering these variables before launching Airflow. We'll also need to be careful not to alter any _other_ connection types, e.g. `postgres://` or `s3://`.

## Alternatives

This _may_ be possible with an Airflow plugin, but we'll have to look into that a bit more.

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

sarayourfriend:
@AetherUnbound would this be a good-first-issue by any change?

AetherUnbound:
I don't think so, unfortunately. There are a lot of Airflow-specific pieces at play here, and it might be unclear for new contributors how to check that the connections exist (even if creating the munging piece of the entrypoint is straightforward).

AetherUnbound:
I guess I could document that here! Let me write up some steps for that, then we can mark it as a good first issue :slightly_smiling_face:

AetherUnbound:
[...]

PR #452, opened 2022-03-30
Update Smithsonian Unit code checker DAG to alert to Slack

### LINKED ISSUE #420: Add all sub-providers to Smithsonian DAG
## Current Situation

#412 
We have a `check_new_smithsonian_unit_codes` DAG that we can run to check for the addition of new sub-providers for the Smithsonian API.
https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/loader/smithsonian_unit_codes.py
https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/check_new_smithsonian_unit_codes_workflow.py

Currently, this DAG will alert you to the existence of _new_ sub-providers, but it will also search for sub-providers that aren't currently returning any data and suggest they be deleted.

## Suggested Improvement

The DAG doesn't throw any errors when querying for a sub-provider with no data, and the list of sub-providers is pretty small (so there doesn't appear to be any significant cost to querying _all_ of them). We could update this DAG to only check for new providers and pro-actively configure the provider script with all existing ones, and then add all of the currently available unit-codes (including inactive ones) to the configuration.

I grabbed the list of unit-codes from the Smithsonian API ('[/terms/unit_code](http://edan.si.edu/openaccess/apidocs/#api-search-terms)') and found that there are 13 unit-codes (sub-providers) supported by the API that are not configured in Openverse:
```
‘AAA’, ‘AAG’, ‘ACAH’, ‘ACMA’, ‘CFCHFOLKLIFE’, ‘EEPA’, ‘FSA’, ‘HSFA’, ‘NMAIA’, ‘OCIO-DPO3D’, ‘SAAMPAIK’, ‘SC’, ‘SI’
```

## Benefit

If any of these 13 sub-providers start sending data, the provider DAG will _immediately_ pick it up. Currently the DAG throws an Exception if it encounters an unfamiliar unit code.

## Implementation

stacimc:
## Fixes

Fixes #420

## Description

On main, we have a DAG called `check_new_smithsonian_unit_codes_workflow` which can only be run manually (it's not on a schedule), which compares the configuration in the `SMITHSONIAN_SUB_PROVIDERS` dictionary with the unit codes available in the Smithsonian API and raises an exception if they have fallen out of sync. It also creates and populates a DB table with the codes that need to be synced, which appears to be for legacy purposes and is no longer used anywhere.

This PR:

* Removes the two outdated unit codes from the Smithsonian sub provider configuration
* Entirely removes the `check_new_smithsonian_unit_codes_workflow` DAG and moves the logic into the Smithsonian provider script itself, so that it will fail fast if the unit code configuration needs to be changed.
* Removes logic for creating/updating DB table and instead reports the needed changes to Slack in a message like this:

Screen Shot 2022-04-04 at 3 15 24 PM

## Notes
* Currently if the unit code validation fails, we'll get the Slack message with the error, but the `load_complete` Slack message will _also_ fire. There's an issue to address that: #465

## Testing Instructions

**Prereqs to test Slack:**
* You'll need to make sure you have the Slack connection set up for the Alerts channel. In your .env, that's `AIRFLOW_CONN_SLACK_ALERTS=<slack_connection_url>`. Troubleshooting tip: make sure it is formatted correctly with double-prepended 'https' and URL encoding (eg `"https://https%3A%2F%2Fhooks.slack.com%2Fservices%2everythingelse"`)
* You'll need to override the default behavior of omitting Slack messages locally. Go into the Airflow UI and select Admin > Variables, then add a variable named `slack_message_override` with value `true`.

Test:
* Check out this branch and run the Smithsonian provider script. With the changes in this PR, it should run successfully.
    * Watch the logs for the `pull_data` task and verify it starts writing records to disk, then mark the task as successful.
* Manually update the `SMITHSONIAN_SUB_PROVIDERS` dictionary in `provider_details.py` to make the unit codes out of sync. Try adding in non-existing codes ("FOO" etc), and deleting existing codes.
    * Run the provider script again and verify that the `pull_data` task fails and the error is sent to Slack. 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


stacimc:
Rebased to fix tests!

I keep going back and forth about whether to omit outdated unit codes or intentionally configure all of them. I don't think there's necessarily a _wrong_ choice so curious what others think.

Another consideration is the scheduling of the `check_new_smithsonian_unit_codes_workflow` DAG. Currently it's not scheduled at all and has to be run manually. The Smithsonian DAG itself is run weekly, and we'd like the checker to catch new codes beforehand. We _could_ run it daily (since it's such a small task).

On the other hand and especially now that we've removed the database aspect to it, I actually think it might be a good idea to eliminate the checker DAG entirely and simply perform the unit code validation as a first step in the Smithsonian's `main` function. That way we alert to Slack and fail early, wasting no time. Thoughts?

openverse_catalog/dags/common/loader/smithsonian_unit_codes.py
```diff
@@ -61,74 +28,34 @@ def get_new_and_outdated_unit_codes(unit_code_set, sub_prov_dict=SUB_PROVIDERS):
...
+    """
     response = requests.get(units_endpoint, params=query_params)
     unit_code_set = set(response.json().get("response", {}).get("terms", []))
     new_unit_codes, outdated_unit_codes = get_new_and_outdated_unit_codes(unit_code_set)
 
-    initialise_unit_code_table(postgres_conn_id, unit_code_table)
-
-    postgres = PostgresHook(postgres_conn_id=postgres_conn_id)
+    if bool(new_unit_codes) or bool(outdated_unit_codes):
+        message = "\n*Updates needed to the SMITHSONIAN_SUB_PROVIDERS dictionary*:\n\n"
```
AetherUnbound, line 47: Since we're using `bool(<value>)` a few times, it might be worth doing assignment expressions here?

```suggestion
    if (has_new_codes := bool(new_unit_codes)) or (has_outdated_codes := bool(outdated_unit_codes)):
        message = "\n*Updates needed to the SMITHSONIAN_SUB_PROVIDERS dictionary*:\n\n"
```

And then use `has_<new|outdated>_codes` below

openverse_catalog/dags/common/loader/smithsonian_unit_codes.py
```diff
@@ -61,74 +28,34 @@ def get_new_and_outdated_unit_codes(unit_code_set, sub_prov_dict=SUB_PROVIDERS):
...
-                  '{new_unit_code}', 'add'
-                );
-                """
-            )
-        )
+        if bool(new_unit_codes):
+            codes_string = "\n".join(f"  - `{code}`" for code in new_unit_codes)
+            message += "New unit codes must be added:\n"
+            message += codes_string
+            message += "\n"
```
AetherUnbound, line 53: Might be worth doing a docstring here:
```suggestion
            message += f"""\
New unit codes must be added:
{codes_string}
"""
```

openverse_catalog/dags/common/slack.py
```diff
@@ -253,6 +253,21 @@ def should_send_message(http_conn_id=SLACK_NOTIFICATIONS_CONN_ID):
     return environment == "prod" or force_message
 
 
+def send_alert(
+    text: str,
+    username: str = "Airflow",
```
AetherUnbound, line 258: Maybe a bit more specific here? Not sure if that really benefits us here, but thought I'd throw it out there :woman_shrugging: 
```suggestion
    username: str = "Airflow Alert",
```

AetherUnbound (commented):
I love deleting code! :partying_face: The approach here looks good and I love the tests. Do you mind adding a screenshot of an example message if possible?

I think I would also be totally okay with moving this logic _into_ the provider script. It seems like it's a necessary step for its operation, so having it as part of the operational code makes sense to me. I suspect that many of these provider scripts were written in such a way that they could be tested on bare-metal, outside the docker stack. We've now (for better or worse) made running these within the docker stack a requirement, at least for providers that require an API key. I'm sure that the separation of these two steps was an attempt to make it easier to run the provider script locally - but our DevX is better now so I think it's okay that we run/test them within the stack :smile:

openverse_catalog/dags/common/loader/smithsonian_unit_codes.py
stacimc: We could do that! I was annoyed by what the lack of whitespace does to the formatting, but obviously not a big deal 😄

openverse_catalog/dags/common/loader/smithsonian_unit_codes.py
stacimc: I generally agree, but in this case I think if `has_new_codes` is True, then the conditional will immediately be evaluated to True and `has_outdated_codes` won't get assigned 🤔 Since we're only evaluating `bool(<value>)` twice per value, I'm not sure it's worth working around that.

openverse_catalog/dags/common/loader/smithsonian_unit_codes.py
AetherUnbound: I hear it :sweat_smile: We could use [`textwrap.dedent`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) as well, but sometimes that can misbehave if you're formatting in a variable that includes newlines (because those lines don't have the same number of starting spaces).

openverse_catalog/dags/common/loader/smithsonian_unit_codes.py
AetherUnbound: WOAH, okay I definitely didn't consider that! Good to know for the future :memo:

openverse_catalog/dags/providers/provider_api_scripts/smithsonian.py
```diff
@@ -186,6 +190,48 @@ def _gather_unit_sample(unit, sample_dir, retries=RETRIES, endpoint=SEARCH_ENDPO
             break
 
 
+def get_new_and_outdated_unit_codes(unit_code_set, sub_prov_dict=SUB_PROVIDERS):
+    sub_provider_unit_code_set = set()
+
+    for sub_prov, unit_code_sub_set in sub_prov_dict.items():
+        sub_provider_unit_code_set = sub_provider_unit_code_set.union(unit_code_sub_set)
```
AetherUnbound, line 197: If I'm interpreting this correctly, I think we can just do something like the following:
```python
[ins] In [1]: d = {1: {1, 2, 3}, 2: {3, 4, 5}, 3: {8, 9}}

[ins] In [2]: set().union(*d.values())
Out[2]: {1, 2, 3, 4, 5, 8, 9}
```

```suggestion
    all_sub_provider_unit_codes = set().union(*sub_prov_dict.values())
```

openverse_catalog/dags/providers/provider_api_scripts/smithsonian.py
```diff
@@ -186,6 +190,48 @@ def _gather_unit_sample(unit, sample_dir, retries=RETRIES, endpoint=SEARCH_ENDPO
             break
 
 
+def get_new_and_outdated_unit_codes(unit_code_set, sub_prov_dict=SUB_PROVIDERS):
```
AetherUnbound, line 193: Nit: A number of these variables are suffixed with the type (e.g. `_set`, `_dict`). It might be better to remove the suffixes and add type annotations to the function instead.

tests/dags/providers/provider_api_scripts/test_smithsonian.py
```diff
@@ -48,6 +49,47 @@ def test_get_hash_prefixes_with_len_one():
...
+        ((), ("e",)),
+    ],
+)
+def test_validate_unit_codes_from_api_raises_exception(
+    new_unit_codes, outdated_unit_codes
+):
+    with patch.object(
+        si,
+        "get_new_and_outdated_unit_codes",
+        return_value=(set(new_unit_codes), set(outdated_unit_codes)),
```
AetherUnbound, line 76: Could `new_unit_codes` and `outdated_unit_codes` be defined as sets in the parametrization?

AetherUnbound (approved):
So much simpler!!! I have a few comments, but nothing to stop a merge :smile: :rocket:

openverse_catalog/dags/providers/provider_api_scripts/smithsonian.py
stacimc: For sure -- this is just a copy/paste of the old logic 😄 I'll update the types while I'm at it!

AetherUnbound: D'oh! Okay that makes a lot of sense :sweat_smile:

stacimc:
Sorry about the linting errors, there was a mismatch between my local and CI. Rebased and it seems to be good now 👍

tests/dags/providers/provider_api_scripts/test_smithsonian.py
```diff
@@ -73,11 +73,11 @@ def test_validate_unit_codes_from_api_raises_exception(
     with patch.object(
         si,
         "get_new_and_outdated_unit_codes",
-        return_value=(set(new_unit_codes), set(outdated_unit_codes)),
+        return_value=(new_unit_codes, outdated_unit_codes),
     ):
         with pytest.raises(
             AirflowException,
-            match="^\n\*Updates needed to the SMITHSONIAN_SUB_PROVIDERS dictionary\**",
+            match="^\n\\*Updates needed to the SMITHSONIAN_SUB_PROVIDERS dictionary\\**",
```
AetherUnbound, line 80: I think you could maybe use a [raw string](https://www.askpython.com/python/string/python-raw-strings) here so you don't have to escape the escape:
```suggestion
            match=r"^\n\*Updates needed to the SMITHSONIAN_SUB_PROVIDERS dictionary\**",
```

sarayourfriend:
@obulat and @dhruvkb can one of y'all review this PR soon?

obulat (approved):
I love the simplification of the Smithsonian data collection! 
Would it be possible to just add / remove the codes in the loader itself programmatically, without requiring the manual fixes?

PS. I meant to comment on the code, but approved instead. I need some help actually testing the Slack integration. Where do I get the connection URL?

AetherUnbound:
> PS. I meant to comment on the code, but approved instead. I need some help actually testing the Slack integration. Where do I get the connection URL?

It should be in our secret store! Note that there are some annoying things you have to do in order to get the connection set up correctly: 
https://github.com/WordPress/openverse-catalog/blob/c928589b96138efd823c4aca099cbc4baf020e62/env.template#L49-L54

We're working on improving those in #433 🙂

obulat (approved):
Thank you for all the detailed instructions for testing this PR. I got it to send a Slack message, as you describe.

stacimc:
Rebased for merge conflicts :)

> Would it be possible to just add / remove the codes in the loader itself programmatically, without requiring the manual fixes?

There actually used to be a DB table being populated by the checker (which I removed), but it was just used to store codes that need to be added/removed -- you still had to inspect the table and manually make changes to the constant in `provider_details.py`. We could reintroduce that and use it in the loader _instead_ of the configuration file, but I prefer the current setup. The loader will fail immediately, so there's no wasted time. I also like getting a chance to review the changes before automatically updating 😄 If the codes change more frequently than we expect and require a ton of manual intervention, we should revisit the process!

PR #455, opened 2022-03-31
Remove `apt upgrade` from PG image, upgrade to 13.6

AetherUnbound:
## Fixes

This PR fixes an issue that cropped up spontaneously on the CI for our main branch: https://github.com/WordPress/openverse-catalog/runs/5740531711?check_suite_focus=true

## Description

Upon investigating, it appeared postgres was failing to start during testing due to the following error:

```
initdb: error: invalid locale settings; check LANG and LC_* environment variables
The files belonging to this database system will be owned by user "postgres".
This user must also own the server process.
```

While attempting to resolve this issue, I noticed some odd things happening in the `apt` steps of the Postgres Docker image. Namely:
* Postgres was being updated to 13.6 (from the 13.2 present in our base image)
* Locales _were_ being generated

Looking into this more, I found a GitHub issue that exactly described the problem we were having and its resolution: https://github.com/docker-library/postgres/issues/415#issuecomment-467997249

TL;DR: We should **never** be running `apt upgrade` as part of a Docker image definition. Due to the sheer number of packages that are upgraded, this can (and has in our case!) causes some inconsistencies that Docker intentionally tries to prevent from happening. The general guidance is that one should only ever install packages they know they'll need, and doing so will upgrade any that are already installed. (TIL!)

Since one of the libraries we're installing for `pgcli` was installing Postgres 13.6 anyway, I thought it best to just bump the base image to that version. We don't have to worry about upgrading the database files since it's a minor version upgrade.

I also added a comment for the build step - we're rebuilding the webserver container, which seemingly gets built in the previous step. These are, however, different versions: the `just build` step installs the dev dependencies, while the first build step only installs the prod ones. The latter image is what we publish to ghcr.io, so we want to keep that all as-is.

## Testing Instructions

Hopefully the CI on this works!

To replicate this locally (on main), force a rebuild of the postgres image. You should immediately encounter the same error. After building with the version on this branch, you should no longer see the error.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -75,6 +75,8 @@ jobs:
         run: chmod -R 777 openverse_catalog/ tests/
 
       - name: Test
+        # The containers must be rebuilt here because we need the dev dependencies
+        # rather than the production ones
         run: |
           just build
           just test
--- a/docker/local_postgres/Dockerfile
+++ b/docker/local_postgres/Dockerfile
@@ -1,4 +1,4 @@
-FROM postgres:13.2
+FROM postgres:13.6
 
 ARG PGCLI_VERSION=3.2.0
 
--- a/docker/local_postgres/Dockerfile
+++ b/docker/local_postgres/Dockerfile
@@ -8,7 +8,7 @@ ENV PIPNOCACHEDIR=1
 ENV PYTHONUNBUFFERED=1
 ENV PIP_NO_COLOR=1
 
-RUN apt-get update && apt-get -yqq upgrade && apt-get -yqq install \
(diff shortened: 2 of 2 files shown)


krysal (approved):
Good find! Thanks for adding the comment about the build step also.

stacimc (approved):
Nice, and thanks for the thorough explanation as always -- very helpful! ✨

krysal:
Let's get this merged to fix the CI workflow on the main branch 🚀

PR #458, opened 2022-03-31
Adjust timeouts for Data Refresh `wait_for_completion` step

### LINKED ISSUE #447: Override default execution timeout for data refresh DAG sensors
## Description

We ran into an issue recently where the data refresh DAG was poking the ingestion server status successfully, but was cut short prior to reaching the _sensor_ timeout. We realized that this was because we had reached the default _execution_ timeout:
https://github.com/WordPress/openverse-catalog/blob/962f1c5d9bb17bd6ea1ed5cc0674bab76d638f0b/openverse_catalog/dags/common/constants.py#L17

While this is a sensible timeout for typical tasks (e.g. the `HttpOperator` that triggers the data refresh), it's not ideal for the tasks that we know will be longer running. We still believe that having _some_ defined upper bound in general is a good policy - we'd rather add exceptions for longer tasks than never hear about a task that ran for days and days because we neglected to add an explicit timeout.

This timeout will only need to be added to the 2 sensor tasks in the data refresh DAGs.

We should also evaluate whether we should tune down the default sensor poke interval timeout, as it currently sits at one week.

## Reproduction

1. Start a data refresh DAG
2. Let it run for over an hour
3. Observe that the execution timeout gets reached

## Screenshots

**Airflow exception alert**
image

**Execution timeout value under task instance details**
image

## Resolution

stacimc:
## Fixes

Fixes #447 by @AetherUnbound 

## Description

The `data_refresh` DAGs are currently timing out after one hour of the `wait_for_completion` step, while we ping the ingestion server for the status of the data refresh. The problem was that the task was not in `reschedule` mode, so it would not free the slot after an unsuccessful ping. This caused it to timeout on its lower `execution_timeout` rather than the `timeout`. ([Relevant docs](https://airflow.apache.org/docs/apache-airflow/stable/concepts/tasks.html#timeouts))

This PR:
* Sets the `wait_for_completion` task into reschedule mode. Now it will timeout only if a single attempt at poking takes longer than the `execution_timeout` of 1hr, or if no poke is successful before the `timeout`
* Updates `timeout` to 3 days instead of the default 1 week. **Looking for confirmation that this is a reasonable value; I think it should be based on past runs**
* Updates the `poke_interval` to 15 minutes instead of the default 1 minute, as the refresh is a long running task and we don't need to check that aggressively. This poke interval can be overridden by the same `DATA_REFRESH_POKE_INTERVAL` env var for testing purposes 😄 

## Testing Instructions

**Prereqs**
* Set up an Airflow connection to the ingestion server. To hit your local server (recommended), you can set this in your `openverse-catalog/.env` (Note that it MUST be escaped/prepended like this. Use `172.17.0.1` instead of `host.docker.internal` on linux. Especially note that `http` is prepended twice 😄 ):
`AIRFLOW_CONN_DATA_REFRESH="http://http%3A%2F%2Fhost.docker.internal%3A8001"`
* For testing purposes, you'll want to set the `DATA_REFRESH_POKE_INTERVAL` environment variable in your `.env`, or else you'll have to wait 15 minutes for the second DAG to start. I set mine to `5`, so the task gets rescheduled every 5 seconds.
* Run `just up`
* Create the special `data_refresh` pool that the refresh DAGs will run in. This is done through the Airflow UI:
    * In the main menu at the top of the page, navigate to Admin > Pools
    * Click the `+` Icon to add a new Pool
    * Name it `data_refresh` and set the `Slots` to `1`.
    * Save

Try running a data refresh locally on `main`, and observe it from the Tree view (http://localhost:9090/tree?dag_id=audio_data_refresh). Notice that the `wait_for_completion` step stays in the 'running' state until it resolves.

Now checkout the branch and try again. Notice that the `wait_for_completion` step now enters the `up_for_reschedule` state after the first poke fails!

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/data_refresh/dag_factory.py
+++ b/openverse_catalog/dags/data_refresh/dag_factory.py
@@ -129,6 +129,7 @@ def create_data_refresh_dag(data_refresh: DataRefresh, external_dag_ids: Sequenc
         **data_refresh.default_args,
         "pool": DATA_REFRESH_POOL,
     }
+    poke_interval = int(os.getenv("DATA_REFRESH_POKE_INTERVAL", 60 * 15))
     dag = DAG(
         dag_id=data_refresh.dag_id,
         default_args=default_args,
@@ -147,7 +148,7 @@ def create_data_refresh_dag(data_refresh: DataRefresh, external_dag_ids: Sequenc
             external_dag_ids=external_dag_ids,
             check_existence=True,
             dag=dag,
-            poke_interval=int(os.getenv("DATA_REFRESH_POKE_INTERVAL", 60 * 15)),
+            poke_interval=poke_interval,
             mode="reschedule",
         )
 
@@ -180,6 +181,7 @@ def create_data_refresh_dag(data_refresh: DataRefresh, external_dag_ids: Sequenc
             response_check=response_check_wait_for_completion,
             dag=dag,
             mode="reschedule",
+            poke_interval=poke_interval,
             timeout=(60 * 60 * 24 * 3),  # 3 days
         )
 
--- a/openverse_catalog/dags/data_refresh/dag_factory.py
+++ b/openverse_catalog/dags/data_refresh/dag_factory.py
@@ -180,6 +180,7 @@ def create_data_refresh_dag(data_refresh: DataRefresh, external_dag_ids: Sequenc
             response_check=response_check_wait_for_completion,
             dag=dag,
             mode="reschedule",
+            timeout=(60 * 60 * 24 * 3),  # 3 days
         )
 
         wait_for_data_refresh >> trigger_data_refresh >> wait_for_completion
--- a/openverse_catalog/dags/data_refresh/dag_factory.py
+++ b/openverse_catalog/dags/data_refresh/dag_factory.py
@@ -179,6 +179,7 @@ def create_data_refresh_dag(data_refresh: DataRefresh, external_dag_ids: Sequenc
             method="GET",
             response_check=response_check_wait_for_completion,
             dag=dag,
+            mode="reschedule",
         )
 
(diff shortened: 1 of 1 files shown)


stacimc:
> I am sure I'm missing some important step(s), but I don't know what...

@obulat Ah, I think you'll need to create the `data_refresh` pool! The data refresh DAGs run in that special pool, and Airflow won't run the tasks if it can't find it. In the UI in Airflow you can go to Admin > Pools, then create a new pool. Set `Pool` to `data_refresh` and `Slots` to 1. Sorry about that, I'll add that to the testing instructions!

AetherUnbound (approved):
Tested locally, works like a charm! :star_struck:

obulat (approved):
Thank you so much for the detailed updated instructions, @stacimc !
I tested it locally, and it worked. The `audio_refresh` sets the `image_refresh` to `up_for_reschedule` if it's running.
There is probably something else missing in my setup, because the running `refresh` task still raises an error:
```
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/airflow/models/connection.py", line 410, in get_connection_from_secrets
    raise AirflowNotFoundException(f"The conn_id `{conn_id}` isn't defined")
airflow.exceptions.AirflowNotFoundException: The conn_id `data_refresh` isn't defined
```

AetherUnbound:
Ah, you may need to perform some of the testing setup steps listed here to actually get the data refresh connection set up! :smile: https://github.com/WordPress/openverse-catalog/pull/397

stacimc:
Sorry I missed this -- @AetherUnbound is right, there were some setup steps for the data refresh I neglected to add. I've copied them into the testing instructions here as well, my apologies!

PR #467, opened 2022-04-05
Re-introduce pytest-socket

### LINKED ISSUE #342: Re-investigate using pytest-socket
## Description

`pytest-socket` was removed in #276 because it was causing some inconsistencies with test completion when tests were run out of order. As discussed in #340, it might make sense to bring the package back to ensure that provider tests aren't making any unnecessary request calls during testing.

## Benefit

Where possible, all external requests in tests should be mocked so that we don't have to rely on the uptime/availability of the service in order to run our tests.

## Implementation

AetherUnbound:
## Fixes

Fixes #342 by @AetherUnbound

## Description

This PR re-introduces pytest-socket, which ensures that tests don't attempt to reach out to services via a socket unless explicitly allowed. 

In attempting to uncover which tests were failing with pytest-socket, I found that the license URL validation [makes a request to Creative Commons to ensure the URL is accurate](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/urls.py#L65). This was the cause of almost all of our external accessing and was used internally by many tests across the codebase. The only way to mock it in its current form is to auto-use a global fixture that mocks all `request.get` calls, which has undesirable side effects for other tests. So I opted to alias `requests.get` within that module specifically, which made the mock much easier to patch (even though the patch still needed to remain as a global fixture).

A few tests required specific care:
* The Smithsonian unit code test was making a request to the Smithsonian API that was immediately return with an invalid API key error :joy: Since we don't really care what the response is, I mocked this one.
* The single run external DAG sensor test was emitting a warning that had nothing to do with any code we had written and everything to do with Airflow internals. I suppressed that warning for now so we can have cleaner test output.

In addition to making sure our tests don't rely on any external services, a fun side-effect of this is that our tests are now slightly faster!

## Testing Instructions

You'll need to rebuild the container in order to test with this new dependency:
1. `just build`
1. `just test`

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/tests/dags/common/loader/test_smithsonian_unit_codes.py
+++ b/tests/dags/common/loader/test_smithsonian_unit_codes.py
@@ -1,6 +1,6 @@
 import os
 from collections import namedtuple
-from unittest.mock import patch
+from unittest.mock import Mock, patch
 
 import psycopg2
 import pytest
@@ -52,10 +52,14 @@ def test_alert_new_unit_codes():
 def test_alert_unit_codes_from_api(postgres_with_test_unit_code_table):
     postgres_conn_id = POSTGRES_CONN_ID
     unit_code_table = SI_UNIT_CODE_TABLE
+    response_mock = Mock()
+    response_mock.json.return_value = {}
 
     with patch.object(
         si, "get_new_and_outdated_unit_codes", return_value=({"d"}, {"e"})
-    ) as mock_get_unit_codes:
+    ) as mock_get_unit_codes, patch.object(
+        si.requests, "get", return_value=response_mock
+    ):
         with pytest.raises(Exception):
             si.alert_unit_codes_from_api(postgres_conn_id, unit_code_table)
 
--- a/openverse_catalog/dags/common/urls.py
+++ b/openverse_catalog/dags/common/urls.py
@@ -7,8 +7,10 @@ import re
 from functools import lru_cache
 from urllib.parse import urlparse
 
-import requests
 import tldextract
+
+# This import is aliased for easier test mocking
+from requests import get as requests_get
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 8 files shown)


AetherUnbound:
I'm working on addressing the CI failures. It turns out that:
* Docker will use different internal ports depending on what's available on the machine
* Boto3 will open a socket with the IP address directly rather than the hostname, for some reason
* Trying to get the host from the hostname dynamically isn't working with our S3 bucket fixture

AetherUnbound:
Ah, okay, here's why: https://github.com/pytest-dev/pytest/issues/1368 :upside_down_face:

tests/dags/common/loader/test_smithsonian_unit_codes.py
```diff
@@ -52,10 +52,14 @@ def test_alert_new_unit_codes():
...
     unit_code_table = SI_UNIT_CODE_TABLE
+    response_mock = Mock()
+    response_mock.json.return_value = {}
 
     with patch.object(
         si, "get_new_and_outdated_unit_codes", return_value=({"d"}, {"e"})
-    ) as mock_get_unit_codes:
+    ) as mock_get_unit_codes, patch.object(
+        si.requests, "get", return_value=response_mock
+    ):
```
stacimc, line 62: FWIW this code will be deleted in #452, but I am mocking it in the [new tests](https://github.com/WordPress/openverse-catalog/pull/452/files#diff-b25f876380bf67a37d45141dba9bf98b9e88ba7194b7e2e2e02d5678172b4840R73) so it should still be good 😄

AetherUnbound: NICE :fire:

stacimc (approved):
LGTM! Took me a minute  to reason through the changes, but I think the import aliasing is a good solution for the license url validation. The additional docs are always appreciated 😃

sarayourfriend (approved):
LGTM!

PR #480, opened 2022-04-13
Make Airflow connection variables easier to read

rwidom:
## Fixes
Fixes #433 by @AetherUnbound 

## Description
Makes the airflow connection environment variables a little easier to read and maintain, by handling the url encoding and https:// prepending automagically where necessary.

## Testing Instructions
@AetherUnbound  wrote great testing instructions here: https://github.com/WordPress/openverse-catalog/issues/433#issuecomment-1074446931 

The only (probably obvious) thing I would add is that after running `just dotenv`, I needed to manually edit `.env` to try out various values with "AIRFLOW_CONN_DATA_REFRESH=" rather than "not_set".

Here are my results:
image

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [ ] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #493, opened 2022-04-26
Recreate the audioset matview after full popularity recalculation

### LINKED ISSUE #448: Ensure audioset_view gets recreated during popularity calculation
## Description

We recently encountered an error where, after the `recreate_audio_popularity_calculation` DAG was run, the `audioset_view` was not correctly re-created. This is a result of the `DROP ... CASCADE` statement that gets run when removing the views: https://github.com/WordPress/openverse-catalog/blob/89767ec9bf6ccc499b405cd88d82e6b898c90f0a/openverse_catalog/dags/common/popularity/sql.py#L76

Because `audioset_view` [depends on](https://github.com/WordPress/openverse-catalog/blob/0c5b43fbab8cba491817ebb647b77e17114a7353/docker/local_postgres/0007_openledger_audio_view.sql#L90-L103) the `audio_view`, when the latter gets dropped so does the former. We don't have any explicit logic for recreating `audioset_view` so it never gets recreated.

## Reproduction

1. Run the `recreate_audio_popularity_calculation` DAG
2. Observe that `audioset_view` gets dropped and never recreated

## Resolution

stacimc:
## Fixes

Fixes #448 by @AetherUnbound 

### Background
Most of the time when we want to recalculate popularity data, we use the `refresh_audio_view_data` DAG to refresh the matview with popularity data, and once a month we run `refresh_all_audio_popularity_data` to recalculate popularity entirely. This workflow works to update data for newly ingested images and pick up things like new popularity metrics.

If new SQL code is deployed for the popularity calculations though, it's necessary to drop the existing matviews and functions and recreate them. We use the `recreate_audio_popularity_calculation` DAG to do this.

### Problem
The `recreate_audio_popularity_calculation` DAG removes the views in its first step, including both the `audio_view` and the dependent `audioset_view`: https://github.com/WordPress/openverse-catalog/blob/89767ec9bf6ccc499b405cd88d82e6b898c90f0a/openverse_catalog/dags/common/popularity/sql.py#L76

The `audio_view` is recreated in the final step, but not the `audioset_view`. This causes a problem during the data refresh, because we expect `audioset_view` t[o exist in the upstream db](https://github.com/WordPress/openverse-api/blob/main/ingestion_server/ingestion_server/tasks.py#L133).

This PR adds a check to the DAG to recreate the `audioset_view` automatically.

## Testing Instructions

To get set up for testing locally:
* Update your local ingestion server to connect to your local Catalog DB (as opposed to the catalog mock it spins up by default). To do this, update your [`openverse-api/ingestion_server/env.docker` ](https://github.com/WordPress/openverse-api/blob/main/ingestion_server/env.docker#L7) to point `UPSTREAM_DB_HOST="172.17.0.1"` and `UPSTREAM_DB_PORT="5434"`
* Run the ingestion server locally with `just up` from within the API repo
* Spin up the Catalog locally

Confirm the error on main:
* _With main checked out_, run the `recreate_audio_popularity_calculation` DAG from [Airflow](http://localhost:9090/tree?dag_id=recreate_audio_popularity_calculation). (This will drop the views and recreate them, but fail to recreate the `audioset_view`)
* Now run the `audio_data_refresh` DAG. It should raise an AirflowException for `Error triggering data refresh`
* If you'd like, you can check the ingestion server logs and verify that you see the error `psycopg2.errors.UndefinedTable: relation "audio_view" does not exist`

Confirm the fix:
* Check out this branch
* Re-run `recreate_audio_popularity_calculation`. This time, we expect the views to be dropped and recreated, _including_ `audioset_view`
* Re-run `audio_data_refresh`. If all goes well it should pass this time!

### Set-up troubleshooting
If you're running into errors, verify:
* You have created a `data_refresh` pool in the Airflow UI with slots set to `1`
* You've set up your `AIRFLOW_CONN_DATA_REFRESH` in `openverse-catalog/.env`.
## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/popularity/sql.py
+++ b/openverse_catalog/dags/common/popularity/sql.py
@@ -309,19 +309,19 @@ def create_audioset_view_query():
         CREATE VIEW public.{AUDIOSET_VIEW_NAME}
         AS
         SELECT DISTINCT
-            audio_set ->> 'foreign_identifier'  :: varchar(1000) :: foreign_identifier,
-            audio_set ->> 'title'               :: varchar(2000) :: title,
-            audio_set ->> 'foreign_landing_url' :: varchar(1000) :: foreign_landing_url,
-            audio_set ->> 'creator'             :: varchar(2000) :: creator,
-            audio_set ->> 'creator_url'         :: varchar(2000) :: creator_url,
-            audio_set ->> 'url'                 :: varchar(1000) :: url,
-            audio_set ->> 'filesize'            :: integer       :: filesize,
-            audio_set ->> 'filetype'            :: varchar(80)   :: filetype,
-            audio_set ->> 'thumbnail'           :: varchar(1000) :: thumbnail,
+            (audio_set ->> 'foreign_identifier')  :: varchar(1000) as foreign_identifier,
+            (audio_set ->> 'title')               :: varchar(2000) as title,
+            (audio_set ->> 'foreign_landing_url') :: varchar(1000) as foreign_landing_url,
+            (audio_set ->> 'creator')             :: varchar(2000) as creator,
+            (audio_set ->> 'creator_url')         :: varchar(2000) as creator_url,
+            (audio_set ->> 'url')                 :: varchar(1000) as url,
+            (audio_set ->> 'filesize')            :: integer       as filesize,
+            (audio_set ->> 'filetype')            :: varchar(80)   as filetype,
+            (audio_set ->> 'thumbnail')           :: varchar(1000) as thumbnail,
             provider
         FROM public.{AUDIO_VIEW_NAME}
         WHERE audio_set IS NOT NULL;
-        """
+        """  # noqa: E501
     )
 
 
--- a/openverse_catalog/dags/common/popularity/sql.py
+++ b/openverse_catalog/dags/common/popularity/sql.py
@@ -308,37 +308,19 @@ def create_audioset_view_query():
         f"""
         CREATE VIEW public.{AUDIOSET_VIEW_NAME}
         AS
-            SELECT DISTINCT
-                cast(
-                    audio_set ->> 'foreign_identifier'  as varchar(1000)
-                ) as foreign_identifier,
-                cast(
-                    audio_set ->> 'title'               as varchar(2000)
-                ) as title,
-                cast(
-                    audio_set ->> 'foreign_landing_url' as varchar(1000)
-                ) as foreign_landing_url,
-                cast(
-                    audio_set ->> 'creator'             as varchar(2000)
-                ) as creator,
-                cast(
-                    audio_set ->> 'creator_url'         as varchar(2000)
-                ) as creator_url,
-                cast(
-                    audio_set ->> 'url'                 as varchar(1000)
-                ) as url,
-                cast(
-                    audio_set ->> 'filesize'            as integer
-                )       as filesize,
-                cast(
-                    audio_set ->> 'filetype'            as varchar(80)
-                )   as filetype,
-                cast(
-                    audio_set ->> 'thumbnail'           as varchar(1000)
-                ) as thumbnail,
-                provider
-            FROM public.{AUDIO_VIEW_NAME}
-            WHERE audio_set IS NOT NULL;
+        SELECT DISTINCT
+            audio_set ->> 'foreign_identifier'  :: varchar(1000) :: foreign_identifier,
+            audio_set ->> 'title'               :: varchar(2000) :: title,
+            audio_set ->> 'foreign_landing_url' :: varchar(1000) :: foreign_landing_url,
+            audio_set ->> 'creator'             :: varchar(2000) :: creator,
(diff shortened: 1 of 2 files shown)


openverse_catalog/dags/common/popularity/sql.py
```diff
@@ -299,6 +300,49 @@ def create_standardized_media_popularity_function(
     postgres.run(query)
 
 
+def create_audioset_view_query():
+    """
+    Returns SQL to create the audioset_view.
+    """
+    return dedent(
+        f"""
+        CREATE VIEW public.{AUDIOSET_VIEW_NAME}
```
AetherUnbound, line 309: We should really have this defined somewhere as a source of truth 😓 Not necessary as part of this PR though, but perhaps we should make an issue for it! Also, do you mind changing the `cast` syntax to match what's being set up in https://github.com/WordPress/openverse-api/pull/648?

AetherUnbound (approved):
This looks great! Thanks for the awesome testing instructions, particularly the troubleshooting section - it reminded me that the data refresh pool had been dropped in my last `recreate` 😄 

I have one note about syntax but otherwise this is good!

sarayourfriend:
@stacimc what should AIRFLOW_CONN_DATA_REFRESH be set to?

AetherUnbound:
> @stacimc what should AIRFLOW_CONN_DATA_REFRESH be set to?

This is a frequent enough question, I think we should add a reasonable default to `env.template` so folks don't have to worry about it! You can set it to `AIRFLOW_CONN_DATA_REFRESH="http://172.17.0.1:8001"` if you're on Linux or `AIRFLOW_CONN_DATA_REFRESH="http://host.docker.internal:8001"` on OSX.

openverse_catalog/dags/common/popularity/sql.py
stacimc: > We should really have this defined somewhere as a source of truth

Yeah, I was thinking about that. At the moment we're doing the same here for the audio view and image view -- maybe something that can be addressed in #453 while we're refactoring things.

I forgot about the cast syntax! Updated :)

stacimc:
@sarayourfriend @AetherUnbound I added a default to `env.template` for the `AIRFLOW_CONN_DATA_REFRESH`. I thought I had already added instructions there in a previous PR, but apparently I did not commit them to the template! My bad :)

openverse_catalog/dags/common/popularity/sql.py
```diff
@@ -308,37 +308,19 @@ def create_audioset_view_query():
...
-                    audio_set ->> 'filetype'            as varchar(80)
-                )   as filetype,
-                cast(
-                    audio_set ->> 'thumbnail'           as varchar(1000)
-                ) as thumbnail,
-                provider
-            FROM public.{AUDIO_VIEW_NAME}
-            WHERE audio_set IS NOT NULL;
+        SELECT DISTINCT
+            audio_set ->> 'foreign_identifier'  :: varchar(1000) :: foreign_identifier,
```
AetherUnbound, line 312: I think these may all need to be in the following format (I talk about why a bit in https://github.com/WordPress/openverse-api/pull/648):
```suggestion
            (audio_set ->> 'foreign_identifier')  :: varchar(1000) as foreign_identifier,
```

stacimc: Sorry about that, clearly I didn't read far enough. Fixed and re-tested.

AetherUnbound (approved):
Love it!

sarayourfriend:
@AetherUnbound @stacimc why would this work with the previous casting syntax (using two `::` and no order-of-operations parens)? It seems like in the related PR that syntax caused an execution error in Postgres, why didn't it do that here?

sarayourfriend:
This seems like it should be fine to me, the code is straight forward enough. I do get an error when I try to run it locally, but I think it's probably just a configuration issue. 

```
urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='http%3a%2f%2f172.17.0.1%3a8001', port=80): Max retries exceeded with url: /task (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fd14e35c6d0>: Failed to establish a new connection: [Errno -2] Name or service not known'))
```

This is in the `trigger_data_refresh` step. Why does it try to make the request against port 80 instead of 8081 as is specified in the configuration?

AetherUnbound:
Hmm, the fact that your host is showing up as `http%3a%2f%2f172.17.0.1%3a8001` leads me to believe that something isn't quite working with the URL encoding that was added in #480 - I think it might be because I put quotes around the example value for `AIRFLOW_CONN_DATA_REFRESH` I provided above. If you remove those quotes from the `.env` file, does it work correctly?

sarayourfriend:
Ah! That explains it, sorry! I somehow was only looking at the last range of commits :facepalm:

sarayourfriend (approved):
LGTM! Tested and it works :tada:

PR #496, opened 2022-05-02
Merge popularity calculations and data refresh into a single DAG

### LINKED ISSUE #453: More tightly couple matview refresh, popularity recalculation, and data refresh
## Problem

The media type materialized view refresh, media type popularity recalculation, and media type data refresh steps all represent an implicit dependency chain, however none of the dependent steps are explicitly tied to other steps. This means that we can re-run the data refresh without having refreshed the materialized view, effectively spending hours of work to accomplish nothing. Here is the current schedule as it exists in code (or before we made modifications from ccsearch):

* Data refresh occurred weekly (scheduled by a cron job on the ingestion server)
* Popularity recalculation occurred monthly (scheduled by Airflow)
* Matview refresh was scheduled for every day except the 1st, when the popularity recalculation ran (scheduled by Airflow)

_Note: The popularity recalculation also ran a matview refresh as its final step_

## Description

Since we've moved scheduling of the data refresh from the ingestion server _into_ Airflow, we can better provide first class support for managing & scheduling these dependencies. As far as we can tell, the matview refresh is **required** for the data refresh and **only used** for the data refresh. For this reason, we believe that we can more tightly couple these steps and make the dependencies more explicit. We also need to incorporate the logic wherein once a month, the popularity calculation would need to be run prior to the matview refresh. We propose the following:

```
                  ┌──────────────────────────┐
        ┌────────►│ popularity_recalculation ├──────────┐
        │         └──────────────────────────┘          │
        │                                               ▼
 ┌──────┴────────┐                             ┌─────────────────┐          ┌──────────────┐
 │ monthly_check ├────────────────────────────►│ matview_refresh ├─────────►│ data_refresh │
 └───────────────┘                             └─────────────────┘          └──────────────┘
```

The `monthly_check` step would be a branching operator that would check if a scheduled run is the first run of the month. If so, it would run the popularity recalculation. After that step, or if it was _not_ the first run of the month, the matview refresh step would run. Once the matview refresh had been completed, the data refresh would begin.

We could either trigger an associated data refresh DAG as the last step, or move all of the data refresh tasks into this larger DAG.

This would allow us to fully monitor **all** of the necessary steps for a data refresh in one place, while also ensuring that we aren't wasting compute resources or time. It would also pave the way for moving more pieces of the data refresh into Airflow proper, as they could be added onto this DAG in lieu of the existing data refresh trigger/monitor tasks.

## Additional context

We recently ran the data refresh, only to find that no new data made it into the API! We realized this was because the matview refresh hadn't run. There's really no need to run it daily if only the data refresh needs it.

## Implementation

stacimc:
## Fixes

Fixes #453 by @AetherUnbound 

## Description

For each media type, this PR merges the following DAGs into a single DAG:
* `refresh_all_*_popularity_data`: A DAG that currently runs `@monthly`. It updates the underlying popularity DB tables 
* `refresh_*_view_data`: A DAG that currently runs `@daily`, except on the first of the month (when the previous DAG runs). It refreshes the materialized view that calculates popularity for records. 
* `*_data_refresh`: A DAG that runs `@weekly`. It syncs data (including popularity data) in the Catalog up to the API db and elasticsearch cluster, so that it is accessible via the frontend. 

The new, single DAG combines all the steps and looks like this:
Screen Shot 2022-05-02 at 11 11 41 AM

Key benefits:
* Scheduling of the popularity refresh is made much easier. With the current setup, it's not possible to schedule the data refresh step so that it always occurs _after_ the matview refresh _and_ ensure that it never collides with the monthly popularity recalculation.
* It's much easier to see how the task are related, and easily check the status of the larger process
* When we add new media types, all we have to do is add a new DataRefresh config, and the DAGs are all generated immediately. Previously we would have needed to write at least two new DAGs for the popularity calc.

### How the DAG works

Here's the full DAG, with the TaskGroups expanded so you can see all the steps:
Screen Shot 2022-05-05 at 3 21 37 PM

The steps are:

1. `month_check`: a branching operator that checks whether this is the first time this DAG has run this month. If so, it proceeds to the full popularity recalculation; if not, we go directly to the matview refresh.
2. `refresh_popularity_metrics_and_constants`: this TaskGroup is skipped if this isn't the first run of the month. In this step, we update the underlying popularity DB tables:
    1.  `update_media_popularity_metrics_table`: this adds any new popularity metrics and updates the configured percentile.
    2. `update_media_popularity_constants_view`: this completely recalculates the popularity constant for each provider.
3. `update_materialized_popularity_view`: this always runs. It refreshes the materialized popularity view, which updates popularity data for existing records and adds data for records that were ingested since the last refresh
4. `data_refresh`: Triggers the actual data refresh on the ingestion server, and awaits its completion. Read more in #397.

Only the `wait_for_data_refresh` step (first step of the remote data refresh) is restricted to the `data_refresh` pool with one worker slot. This works to ensure that the remote data refreshes cannot occur concurrently, but the popularity refresh steps preceding it _can_.

There is also an option to pass a `force_refresh_metrics` option in the DagRun config, which can be set to `true` to make the metrics refresh tasks run even if it's not the first of the month. We might use this if a new metric is added/constants need to be recalculated mid-month. We can also set the option to `false` if for some reason we want to skip this time-consuming step.

### Prereqs for setup
* Make sure your ingestion server is running locally (`just up` from the api repo)
* Set up an Airflow connection to the ingestion server in `openverse-catalog/.env`: `AIRFLOW_CONN_DATA_REFRESH="http://172.17.0.1:8001"`
* Create the special `data_refresh` pool through the Airflow UI (`Admin` -> `Pools`). Name it `data_refresh` and give it `1` Slot

### Tests

**First run of the month**
If you have any runs for `audio_data_refresh` from this month, delete them. Then run `audio_data_refresh` (encouraged because it is shorter). Ensure that all tasks, including the `refresh_all_popularity_data` tasks, run

**Subsequent runs**
Once it has finished, re-run `audio_data_refresh` and make sure the `refresh_all_popularity_data` tasks skip, but all others run successfully

**Force refresh_all_popularity_data tasks to run with config**
In Airflow, click on the run button and select "Trigger DAG with config". You'll be prompted to provide config data as json.
* Run with `{"force_refresh_metrics": true}`, and verify that the `refresh_all_popularity_data` tasks _do not skip_ even though it is not the first run of the month
* Run with `{"force_refresh_metrics": false}` and verify they skip again
* Run with an empty config and verify they skip

**Multiple DAGs**
The `data_refresh` steps should not run concurrently, but all previous steps can. To test this we can take advantage of how slow the image refresh is compared to audio.
* Start the image refresh and wait for it to get to the `wait_for_data_refresh` step. Immediately start the audio refresh. It should complete the popularity calculation steps and get to `wait_for_data_refresh` before the image refresh is complete; thus its `wait_for_data_refresh` should enter `up_for_reschedule`, and only succeed when image finishes.
* Also try starting image and audio at the same time and observe that the month_check/popularity steps can happen concurrently

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


openverse_catalog/dags/data_refresh/data_refresh_task_factory.py
```diff
@@ -0,0 +1,174 @@
+"""
```
stacimc, line 1: Note for reviewers: This file just pulls out all the old data_refresh steps into a new subfactory.

openverse_catalog/dags/common/popularity/operators.py
```diff
@@ -7,12 +7,24 @@
 logger = logging.getLogger(__name__)
 
 
+DROP_MEDIA_POPULARITY_RELATIONS_TASK_ID = "drop_media_popularity_relations"
```
obulat, line 10: Nice way to simplify the task ids! The media type is passed on anyways :)

openverse_catalog/dags/common/popularity/operators.py
```diff
@@ -7,12 +7,24 @@
...
 def drop_media_popularity_relations(
     postgres_conn_id,
     media_type="image",
 ):
     return PythonOperator(
-        task_id="drop_media_popularity_relations",
+        task_id=DROP_MEDIA_POPULARITY_RELATIONS_TASK_ID,
         python_callable=sql.drop_media_popularity_relations,
         op_args=[postgres_conn_id, media_type],
     )
```
AetherUnbound, line 30: I love this change! Almost all of the functions updated in this file are pretty simple and return `PythonOperator`s. Do you think it'd make sense to move these out into the DAG proper in this PR? We can always do that in a follow up.

openverse_catalog/dags/data_refresh/data_refresh_types.py
```diff
@@ -21,24 +21,24 @@ class DataRefresh:
...
+    doc_md:               string used for the DAG's documentation markdown
     """
 
     dag_id: str = field(init=False)
     media_type: str
     start_date: datetime = datetime(2020, 1, 1)
-    execution_timeout: timedelta = timedelta(hours=24)
     schedule_interval: Optional[str] = "@weekly"
     default_args: Optional[Dict] = field(default_factory=dict)
+    data_refresh_timeout: timedelta = 24 * 60 * 60  # 1 day
```
AetherUnbound, line 41: This is no longer a timedelta, should it be? Since this is a dataclass, we can always accept an `int` and change it to a `timedelta` in the `__post_init__` function.

openverse_catalog/dags/data_refresh/refresh_popularity_metrics_task_factory.py
```diff
@@ -0,0 +1,48 @@
...
+
+    media_type:  the type of record to refresh
+    """
+    with TaskGroup(group_id=GROUP_ID) as refresh_all_popularity_data:
+        # Update the popularity metrics table, adding any new popularity metrics
+        # and updating the configured percentile.
+        update_metrics = operators.update_media_popularity_metrics(
+            DB_CONN_ID,
+            media_type=media_type,
+        )
```
AetherUnbound, line 37: We can't quite do this here with the functional approach, but the [`BaseOperator` supports several variants of `doc_<x>` fields](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/baseoperator/index.html#airflow.models.baseoperator.BaseOperator). I think these comments would be perfect for that at the task level!

openverse_catalog/dags/data_refresh/refresh_popularity_metrics_task_factory.py
```diff
@@ -0,0 +1,48 @@
...
+
+        # Update the popularity constants view. This completely recalculates the
+        # popularity constant for each provider.
+        update_constants = operators.update_media_popularity_constants(
+            DB_CONN_ID,
+            media_type=media_type,
+        )
+
+        (update_metrics >> update_constants)
```
AetherUnbound, line 46: I _think_ these parens might be redundant:
```suggestion
        update_metrics >> update_constants
```

openverse_catalog/dags/data_refresh/refresh_view_data_task_factory.py
```diff
@@ -0,0 +1,39 @@
...
+
+This should be run every time before a data refresh is triggered.
+"""
+import os
+
+from airflow.utils.trigger_rule import TriggerRule
+from common.popularity import operators
+
+
+DB_CONN_ID = os.getenv("OPENLEDGER_CONN_ID", "postgres_openledger_testing")
```
AetherUnbound, line 19: I feel like we should centralize this somewhere 🤔 maybe in `constants.py`? It seems annoying to have to grab it from the env so frequently.

openverse_catalog/dags/data_refresh/refresh_view_data_task_factory.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def create_refresh_view_data_task(media_type: str):
```
AetherUnbound, line 22: Just a thought - we're referencing `media_type` more and more, which I love! Maybe we should make it a [`typing.Literal` with `image`/`audio` options](https://docs.python.org/3/library/typing.html#typing.Literal) for clarity?

openverse_catalog/dags/data_refresh/refresh_view_data_task_factory.py
```diff
@@ -0,0 +1,39 @@
...
+
+    Required Arguments:
+
+    media_type: the type of record to refresh
+    """
+    refresh_matview = operators.update_db_view(DB_CONN_ID, media_type=media_type)
+
+    # The upstream `refresh_popularity_metrics` task is conditional and may be skipped.
+    # The matview refresh should run regardless, as long as no upstream task failed.
+    refresh_matview.trigger_rule = TriggerRule.NONE_FAILED_MIN_ONE_SUCCESS
```
AetherUnbound, line 37: Another reason to ditch the function-based setup - we could apply the docstring/trigger rule directly in the operator constructor 😄

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -1,135 +1,143 @@
...
-ensure that no two data refresh DAGs can run concurrently, as required by
-the server.
-
-A data refresh occurs on the data refresh server in the openverse-api project.
-This is a task which imports data from the upstream Catalog database into the
-API, copies contents to a new Elasticsearch index, and makes the index "live".
-This process is necessary to make new content added to the Catalog by our
-provider DAGs available on the frontend. You can read more in the [README](
+For the given media type these DAGs will first refresh the popularity data,
+then initiate a data refresh on the data refresh server and await the
```
AetherUnbound, line 5: > `on the data refresh server`

❤️

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -1,135 +1,143 @@
...
-    to poll for the status of the triggered data refresh. This information will
-    then be available via XCom in the downstream tasks.
+@provide_session
+def _month_check(dag_id: str, session: SASession = None) -> str:
     """
-    status_check_url = response.json()["status_check"]
-    return urlparse(status_check_url).path
+    Checks whether there has been a previous DagRun this month. If so,
+    returns the task_id for the matview refresh task; else, returns the
+    task_id for refresh popularity metrics task.
```
AetherUnbound, line 71: Really good idea to check if there are any runs for the month, rather than doing a simple "if date.day == 1" check! ⭐ I was thinking the latter, this is so much better.

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -1,135 +1,143 @@
...
+        return (
+            REFRESH_POPULARITY_METRICS_TASK_ID
+            if force_refresh_metrics
+            else REFRESH_MATERIALIZED_VIEW_TASK_ID
+        )
 
-    if data["error"]:
-        raise AirflowException("Error triggering data refresh.")
+    # Get the most recent successful dagrun for this Dag
+    DR = DagRun
```
AetherUnbound, line 96: I think we've already set this above, so we don't need to set it again?
```suggestion
```

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -1,135 +1,143 @@
...
+    query = (
+        session.query(DR)
+        .filter(DR.dag_id == dag_id, DR.state == State.SUCCESS)
+        .order_by(DR.start_date.desc())
+    )
+    latest_dagrun = query.first()
```
AetherUnbound, line 102: Did we want to collapse this into a single statement like the `current_dagrun`?
```suggestion
    latest_dagrun = (
        session.query(DR)
        .filter(DR.dag_id == dag_id, DR.state == State.SUCCESS)
        .order_by(DR.start_date.desc())
        .first()
    )
```

tests/dags/data_refresh/test_dag_factory.py
```diff
@@ -0,0 +1,119 @@
...
+        start_date=start_date,
+        execution_date=start_date,
+        data_interval=(start_date, start_date),
+        state=dag_state,
+        run_type=DagRunType.MANUAL,
+        conf=conf,
+    )
+
+
+@pytest.mark.parametrize(
```
AetherUnbound, line 36: These tests are incredible 🤩

AetherUnbound (approved):
I....am SO EXCITED about this change!! It's such an excellent simplification and will make this much easier to manage and maintain 🙌🏼

For the most part my comments are trivial, fantastic work 🚀

openverse_catalog/dags/common/popularity/operators.py
stacimc: Hm I thought about this, and definitely want to do this eventually. The reason I hesitated is because a lot of these functions are used in the `recreate_*_popularity_calculation` DAGs (the ones that drop all the tables/views/functions and recreate them entirely), and those are still separate.

I actually was thinking about trying to pull those DAGs into this one as well (in which case there's definitely no need for this file). I did a bunch of work to make it possible to skip the refresh popularity steps entirely with a config option, that looked like this:
Screen Shot 2022-05-04 at 5 44 14 PM

I scrapped it (although I still have the code on a branch) because it's not useful for the majority of cases. But that use case is a valid one -- where we have SQL changes to the popularity function for example, so we need to drop and recreate all tables/functions. In that case it's unnecessary to refresh popularity data before the data refresh.

So, maybe we could do a version of what I have above, but instead of a configuration option to simply skip the refresh steps, we add a configuration option to drop and recreate tables. When enabled, we do the work currently defined in `recreate_*_popularity_calculations`, and then proceed to data refresh. If not, we go through the regular popularity flow (including the month check).

I hesitated here to over-complicate things, especially considering that recreate ought to be run quite infrequently, but there would be a couple advantages:
* When you add a new media type (like 3d models), you can just add a new DataRefreshType and generate every single popularity/data refresh DAG at once 🎉
* We save time because we don't need to refresh the matview unnecessarily if we run the data_refresh DAG after a `recreate...`

What do you think? Worthwhile as a follow-up? Or honestly, I have a lot of the work done in a local branch and could add it here 😄

openverse_catalog/dags/data_refresh/data_refresh_types.py
stacimc: You're right! It should not be a timedelta, the `timeout` on `HttpSensor` accepts an `int`. I'll update the type to match

openverse_catalog/dags/common/popularity/operators.py
stacimc: Actually it looks like `update_media_popularity_metrics` is the only one that's actually shared at the moment, so maybe we can go ahead with that part of the work at least.

openverse_catalog/dags/data_refresh/refresh_view_data_task_factory.py
stacimc: I like this idea! I'm not sure what the cleanest way is to make it extensible for additional model, though. I could define a `MEDIA_TYPES_TYPE` (😬) in `constants.py` for example. It looks like you can't dynamically create a Literal from an array, so we can't do something like `typing.Literal[MEDIA_TYPES]`. We'd need to make sure the type itself is updated when new models are added.

openverse_catalog/dags/common/popularity/operators.py
AetherUnbound: I think this is already a really solid change! I like some of the ideas you've discussed here, I think it could be a worthwhile follow up. I do worry about unnecessary complexity around the recreate steps - how often do we _really_ need to run those? If it's once a quarter, I'm not sure it makes sense to add the extra logic into this DAG. But if it's easier to maintain from a code perspective, that's a solid win...

openverse_catalog/dags/data_refresh/refresh_view_data_task_factory.py
AetherUnbound: Hmm, that sounds okay to me? I mean, we have media types defined in both the [API](https://github.com/WordPress/openverse-api/blob/270b086095ba125fa26e04bf27dff3ab6994e025/api/catalog/api/constants/media_types.py) and the [data refresh server](https://github.com/WordPress/openverse-api/blob/270b086095ba125fa26e04bf27dff3ab6994e025/ingestion_server/ingestion_server/constants/media_types.py) 🤷🏼‍♀️

openverse_catalog/dags/common/popularity/operators.py
stacimc: That was roughly my thinking, and why I ditched the original effort. The DAG gets harder to grok, especially when there are branches that can only be reached via conf.

I think as a compromise, I'll plan on a separate follow-up where I keep the `recreate_*_popularity_calculation` DAGs separate, but generate them with a factory using the `DataRefreshType` list instead of manually writing one for each model.

openverse_catalog/dags/data_refresh/refresh_popularity_metrics_task_factory.py
stacimc: I didn't realize this was available for Tasks! Very handy, especially because it was tricky writing a concise doc for the DAG overall.

openverse_catalog/dags/data_refresh/refresh_view_data_task_factory.py
stacimc: Fair enough :) I'm just annoyed something like `Literal[MEDIA_TYPES]` doesn't seem to be possible, considering we already have `MEDIA_TYPES` defined 😆

openverse_catalog/dags/common/constants.py
```diff
@@ -7,6 +9,8 @@
 IMAGE = "image"
 MEDIA_TYPES = [AUDIO, IMAGE]
 
+MEDIA_TYPES_TYPE = Literal[AUDIO, IMAGE]
```
AetherUnbound, line 12: Oh wait! [It seems like we actually can](https://stackoverflow.com/a/68474515/3277713)! 😮 

Also I should have mentioned this before, and this is a total nit, but - python types are usually declared in CamelCase, since they're not really constants and that matches the definitions in the `typing` library.
```suggestion
MEDIA_TYPES = [AUDIO, IMAGE]

MediaType = Literal[tuple(MEDIA_TYPES)]
```

openverse_catalog/dags/data_refresh/data_refresh_types.py
```diff
@@ -28,17 +30,17 @@ class DataRefresh:
     schedule_interval:    string giving the schedule on which the DAG should
                           be run.  Passed to the airflow.dag.DAG __init__
                           method.
-    data_refresh_timeout: datetime.timedelta giving the amount of time a given data
-                          pull may take.
+    data_refresh_timeout: int giving the amount of time a given data pull
```
AetherUnbound, line 33:
```suggestion
    data_refresh_timeout: int giving the amount of time in seconds a given data pull
```

openverse_catalog/dags/data_refresh/refresh_popularity_metrics_task_factory.py
```diff
@@ -29,20 +30,28 @@ def create_refresh_popularity_metrics_task_group(media_type: str):
...
+        metrics_doc = (
+            "Updates the popularity metrics table, adding any new "
+            "popularity metrics and updating the configured percentile."
+        )
+        update_metrics = PythonOperator(
+            task_id=UPDATE_MEDIA_POPULARITY_METRICS_TASK_ID,
+            python_callable=sql.update_media_popularity_metrics,
+            op_args=[POSTGRES_CONN_ID, media_type],
+            doc=metrics_doc,
         )
```
AetherUnbound, line 42: Would it be OK to define the doc in-line? Or is this separated out into a variable to give it more space before having to start a new line? 😅 
```suggestion
        update_metrics = PythonOperator(
            task_id=UPDATE_MEDIA_POPULARITY_METRICS_TASK_ID,
            python_callable=sql.update_media_popularity_metrics,
            op_args=[POSTGRES_CONN_ID, media_type],
            doc=(
                "Updates the popularity metrics table, adding any new "
                "popularity metrics and updating the configured percentile."
            )
        )
```

openverse_catalog/dags/data_refresh/refresh_view_data_task_factory.py
```diff
@@ -10,30 +10,32 @@
...
-
-    # The upstream `refresh_popularity_metrics` task is conditional and may be skipped.
-    # The matview refresh should run regardless, as long as no upstream task failed.
-    refresh_matview.trigger_rule = TriggerRule.NONE_FAILED_MIN_ONE_SUCCESS
+    refresh_matview = PythonOperator(
+        task_id=UPDATE_DB_VIEW_TASK_ID,
+        python_callable=sql.update_db_view,
+        op_args=[POSTGRES_CONN_ID, media_type],
+        trigger_rule=TriggerRule.NONE_FAILED_MIN_ONE_SUCCESS,
+        doc_md=create_refresh_view_data_task.__doc__,
```
AetherUnbound, line 38: 💖

AetherUnbound (commented):
Just a few more small things!

openverse_catalog/dags/common/constants.py
stacimc: I saw that source, but it goes on to say:
>  However, literal types cannot contain arbitrary expressions: types like Literal[my_string.trim()], Literal[x > 3], or Literal[3j + 4] are all illegal." So this is valid python syntax, but will not be understood by any type checker, which completely defies the point of adding type hints in the first place

Do you think that's still worth it?

openverse_catalog/dags/data_refresh/refresh_popularity_metrics_task_factory.py
stacimc: Yeah no reason, I liked the formatting a little better with it separated out but could go either way. Updated

openverse_catalog/dags/data_refresh/refresh_view_data_task_factory.py
sarayourfriend: Apparently you can do this, sort of, you just have to go the other way around: https://stackoverflow.com/a/64522240

openverse_catalog/dags/common/constants.py
stacimc: Actually, seems like it won't allow constants either. So while it allows this:
`MediaType = Literal('audio', 'image')`
It complains about:
`MediaType = Literal( AUDIO, IMAGE )`

AetherUnbound: Aww, that's a shame. Apologies for the confusion on my part there. No, I think it's all fine as-is!

sarayourfriend: Won't the accepted answer on that SO link Madison shared work?

stacimc: Not in a way that feels clean enough to me. We can do this:

```
# Define constants
AUDIO = 'audio'
IMAGE = 'image'

# Define the type; we cannot pass AUDIO and IMAGE in here
MediaType = typing.Literal['image', 'audio']

# Build MEDIA_TYPES constant from the type
MEDIA_TYPES: typing.Tuple[MediaType, ...] = typing.get_args(MediaType)
```

But we still have to add media types in multiple places, and I think it becomes less readable.

obulat:
I cannot test this. On the first step, when I trigger `audio_data_refresh`, I get an error on `trigger_data_refresh` task:
```
requests.exceptions.ConnectionError: HTTPSConnectionPool(host='172.17.0.1', port=8001): Max retries exceeded with url: /task (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0xffff84e84e50>: Failed to establish a new connection: [Errno 110] Connection timed out'))
[2022-05-11, 15:30:56 +03] {taskinstance.py:1272} INFO - Marking task as UP_FOR_RETRY. dag_id=audio_data_refresh, task_id=data_refresh.trigger_data_refresh, execution_date=20220501T000000, start_date=20220511T122846, end_date=20220511T123056
[2022-05-11, 15:30:56 +03] {standard_task_runner.py:89} ERROR - Failed to execute job 7 for task data_refresh.trigger_data_refresh
```

The ingestion server seems to finish the task, at least that's what I understand from all the logs, and their last lines:
```
2022-05-11 12:27:49,691 INFO indexer.py:389 - Created 'audio' index alias pointing to audio-5a96b0da35c447cea8604ecd5adbd64e
2022-05-11 12:27:49,691 INFO tasks.py:137 - Task a167d816-d173-4410-a4ad-10672404db3b exited.
```

I really love being able to run both the catalog and the API ingestion server locally! I hope I can _actually_ finish the refresh, though :)

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -1,135 +1,144 @@
...
+including the percentile values and any new popularity metrics. The DAG can also
+be run with the `force_refresh_metrics` option to run this refresh after the first
+of the month.
+
+Once this step is complete, the data refresh can be initiated. A data refresh
+occurs on the data refresh server in the openverse-api project. This is a task
+which imports data from the upstream Catalog database into the API, copies contents
+to a new Elasticsearch index, and finally makes the index "live". This process is
+necessary to make new content added to the Catalog by our provider DAGs available
+on the frontend. You can read more in the [README](
```
obulat, line 21: I think "available to the API" would be more accurate than "on the frontend" here.

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -1,135 +1,144 @@
...
-entire pool on waiting for a particular data refresh to run.
+) Importantly, the data refresh TaskGroup is also configured to handle concurrency
+requirements of the data refresh server.
 
 You can find more background information on this process in the following
 issues and related PRs:
 
 - [[Feature] Data refresh orchestration DAG](
 https://github.com/WordPress/openverse-catalog/issues/353)
+- [[Feature] Merge popularity calculations and data refresh into a single DAG](
```
obulat, line 31: Nice to have these PRs linked here

openverse_catalog/dags/data_refresh/data_refresh_types.py
```diff
@@ -17,28 +17,28 @@ class DataRefresh:
...
+    default_args:         dictionary which is passed to the airflow.dag.DAG
+                          __init__ method.
+    start_date:           datetime.datetime giving the
+                          first valid execution_date of the DAG.
+    schedule_interval:    string giving the schedule on which the DAG should
+                          be run.  Passed to the airflow.dag.DAG __init__
+                          method.
+    data_refresh_timeout: int giving the amount of time in seconds a given
+                          data pull may take.
+    doc_md:               string used for the DAG's documentation markdown
```
obulat, line 33:
```suggestion
    doc_md:               str used for the DAG's documentation markdown
```

To be consistent with the change to the media_type docstring above?

obulat:
> If you haven't updated your connection variables to the improved format, you'll see those connection errors.

Thank you, @stacimc! This fixed it, and I managed to run all of the steps you describe in the Testing instructions. It all works really well, and the workflow is much simpler now.

One thing I don't understand is this: I have not ingested any data, I was running `data_refresh` on the fresh install of the Catalog. What data was it using to run it? What was it refreshing?

obulat (approved):
Thank you for such detailed instructions for testing! It all works great and simplifies a lot in the process ❤️

openverse_catalog/dags/data_refresh/dag_factory.py
stacimc: Good call!

openverse_catalog/dags/data_refresh/data_refresh_types.py
stacimc: Oops, definitely! Good catch

stacimc:
> One thing I don't understand is this: I have not ingested any data, I was running data_refresh on the fresh install of the Catalog. What data was it using to run it? What was it refreshing?

@obulat The process syncs the catalog with the API DB, but as far as I understand it isn't smart enough to detect whether there have actually been any changes.

Actually, when you're testing locally the data refresh server doesn't even really connect to your local catalog by default. It connects to a mock `upstream_db` with sample data. You can force it to connect to your local catalog by updating the `ingestion-server/env.docker`:

```
UPSTREAM_DB_HOST="172.17.0.1"
UPSTREAM_DB_POT="5434" # Note you have to update the port too!
```

stacimc:
I just realized that we needed to update the `execution_timeout` for some of the popularity calculation steps for Image; the default of 1 hour was being applied, which certainly would not have been enough for production data 😄

@AetherUnbound, @obulat -- just wanted to give you an opportunity to object/add feedback to the last commit if you'd like before I merge. Sorry about that!

AetherUnbound (approved):
Good call on the timeouts! I love being able to pass around the data classes themselves 😄

obulat:
> Actually, when you're testing locally the data refresh server doesn't even really connect to your local catalog by default. It connects to a mock upstream_db with sample data. You can force it to connect to your local catalog by updating the ingestion-server/env.docker:

Have you tried running both the catalog and the API locally? I have wanted to try it for a year now :) I managed to get it to work, but maybe there's is a simpler way?

Simply setting the upstream_db values didn't work. I looked at the local networks (`docker network ls`) and saw that there are two bridge networks: one for API and one for Catalog, and they are not connected. So, to get the two to be in the same network locally, I added a `network` property to the API docker.compose file like this:
```
networks:
  default:
    external:
      name: openverse-catalog_default
```
Based on https://docs.docker.com/compose/networking/#use-a-pre-existing-network

And it worked, and now the local catalog and the API are talking to each other.

And none of this is related to this PR, of course, 😂

PR #506, opened 2022-05-12
Ensure SMK images don't timeout on validation

### LINKED ISSUE #494: SMK (statensmuseum) images taking too long to be validated
## Description

When we recently tried to enable the `statensmuseum` provider, searches which included results from that provider ended up timing out on the Cloudflare -> API end. After some investigation, it appears that all the `url`s we have for `statensmuseum` images are taking a significant time to return (even on `HEAD` requests). This is causing validation to fail for all images returned in a result set, and I believe that the API isn't able to return a response quick enough for Cloudflare.

We are going to reach out to SMK to see if they can help us rectify this issue.

## Reproduction

1. Attempt a HEAD request: `time curl -I https://iip.smk.dk/iiif/jp2/KKSgb5781.tif.jp2/full/max/0/default.jpg`
2. Observe that it takes longer than our timeout of 2 seconds

## Expectation

These HEAD requests should ideally be very fast, and we should ideally be able to see Statens Museum results in the API.

## Resolution

stacimc:
## Fixes

Fixes #494, #334

## Description

A little bit of background:

SMK (National Gallery of Denmark) returns both HQ and 'legacy' images. HQ images are IIIF enabled and have an `image_iiif_id` and IIIF image/thumbnail urls. Legacy images do not have this `iiif_id` field at all, and no IIIF links.

Our SMK provider script currently assumes that _all_ images are IIIF enabled, and uses an IIIF `image_url` with the size set to `max` (example URL: https://iip.smk.dk/iiif/jp2/KKSgb5781.tif.jp2/full/max/0/default.jpg). These images are very large and take a very long time to download.

This PR does a couple things:
* Renames "statensmuseum" to SMK for clarity
* Updates the `image_url` for HQ (IIIF-enabled) images to max `2048` px wide, which is what [Open SMK itself uses](https://open.smk.dk/artwork/image/KKSgb20980?q=*&page=0)
* Adds support for legacy images as well. Since these do not have IIIF links, we fallback to the `image_native` and `image_thumbnail` fields

Lots of really wonderful images coming from this provider :) I'm excited to get them back into Openverse!

## Additional Notes

Because legacy images don't have an `iiif_id`, I decided to use `id` as the foreign_identifier. According to API docs this is `the ID from collection DB`.

I chose not to use the `image_native` and `image_thumbnail` fields for HQ images as well as legacy ones, because:
* `image_native` for HQ images gives us a max size download link like this: https://api.smk.dk/api/v1/download/W3siaW1nX3VybCI6Imh0dHBzOi8vaWlwLnNtay5kay9paWlmL2pwMi9ra3MxMDM3OWMudGlmLnJlY29uc3RydWN0ZWQudGlmLmpwMi9mdWxsL2Z1bGwvMC9uYXRpdmUuanBnIiwicHVibGljX2RvbWFpbiI6dHJ1ZSwib2JqZWN0X251bWJlciI6IktLUzEwMzc5YyIsIm51bSI6IiJ9XQ==/KKS10379c.jpg
* `image_thumbnail` for HQ images gives us the same IIIF link we're currently using, but with size `1024` instead of `400`, so we get larger thumbnails than we're currently using. In either case it's unclear whether this field is being used at all.

## Testing Instructions

In your local Airflow, run the SMK provider script. Check the logs for the `pull_data` task, and wait until it has written several 100 lines from buffer to disk. We want to make sure we get some legacy images, so give it at least 30 seconds.

Find and download the TSV file that was loaded into storage through your MinIO Console (localhost:5011). In the TSV, the first column is `id`, the third is `image_url`, and the fourth is `thumbnail`.

* You should be able to see some legacy images in the TSV. These are spotted most easily by scrolling down the first column (id) and looking for an id that looks like `1170000790_object`. Its `image_url` and `thumbnail` values should **both** be non-IIIF thumbnail links, something like: "https://api.smk.dk/api/v1/thumbnail/5249f164-62b1-46e5-b099-d9ba9f596407.jpg"
* The other rows are all HQ images. Take a look at the generated `image_url` and see that the size is 2048. Example, `https://iip.smk.dk/iiif/jp2/kksgb20949.tif.jp2/full/!2048,/0/default.jpg` (specifically looking at the `full/!2048,/`, as opposed to `full/max/`)

You can also time a HEAD request for a few URLs of each type to verify the speed improvement. Example:
```
// The old URL format we were using (very slow)
$ time curl -I https://iip.smk.dk/iiif/jp2/KKSgb5781.tif.jp2/full/max/0/default.jpg

// New HQ URL format, restricts the image size (fast)
$ time curl -I https://iip.smk.dk/iiif/jp2/KKSgb5781.tif.jp2/full/!2048,/0/default.jpg

// A legacy image (fast)
$ time curl https://api.smk.dk/api/v1/thumbnail/5249f164-62b1-46e5-b099-d9ba9f596407.jpg

```

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


openverse_catalog/dags/providers/smk_workflow.py
```diff
@@ -18,12 +19,12 @@
 
 logger = logging.getLogger(__name__)
 
-DAG_ID = "staten_museum_workflow"
+DAG_ID = "smk_workflow"
```
AetherUnbound, line 22: Just a note: Airflow does not handle changing DAG IDs well 😅 We will "lose" history in the Airflow UI for `staten_museum_workflow`, and `smk_workflow` will show up as a new DAG. The DAG runs will still be present in the meta DB, they just won't show up in the UI. That said I think this change is fine 🙂 All of the runs of this DAG have been quick and successful!

AetherUnbound (approved):
What a great fix! Thanks for all the comments sprinkled throughout too 🙂 I was able to test this locally and the files indeed look correct. We'll have to keep our eye on this in production to make sure that the `2048` resolution is still passing validation.

openverse_catalog/dags/providers/smk_workflow.py
stacimc: That's true, it creates a new DAG! Very good call out, I didn't consider that meant we would lose history. But I agree -- luckily this one has much history we'd be interested in keeping 😅

openverse_catalog/dags/common/loader/provider_details.py
```diff
@@ -24,7 +24,7 @@
 NYPL_DEFAULT_PROVIDER = "nypl"
 RAWPIXEL_DEFAULT_PROVIDER = "rawpixel"
 SCIENCE_DEFAULT_PROVIDER = "sciencemuseum"
-STATENS_DEFAULT_PROVIDER = "statensmuseum"
+SMK_DEFAULT_PROVIDER = "statensmuseum"
```
zackkrida, line 27: Nit, statensmuseum basically just means 'state museum' and doesn't really represent this source well. Is there any consequence to changing this value?

zackkrida (approved):
Works great! So excited to be able to ingest and enable this source. Have we or should we deleted their existing data as part of this PR, or sometime in the future when we deploy it?

openverse_catalog/dags/common/loader/provider_details.py
AetherUnbound: This value reflects the `provider` field in our actual data I believe. If we wanted to update this, we'd need to update all associated data (and anything that references this provider in the API and the frontend).

zackkrida: I'm okay with that, as I don't believe this value is hardcoded anywhere, so we'd just need to update it in the Django Admin UI, which we need to do anyway to enable it.

zackkrida:
```bash
~/Code/openverse
❯ rg 'statensmuseum'
catalog/openverse_catalog/dags/common/loader/provider_details.py
27:SMK_DEFAULT_PROVIDER = "statensmuseum"
```

quick safety check (this is searching all our repos)

stacimc: Yeah, I kept this as is because it's the field in our actual data. If we update this and do nothing, for example, we'd get a bunch of `duplicate key` errors as soon as we try to ingest data. I assumed it would also require changes on the API/frontend, but since it seems it doesn't, I think all we'd need to do is:
* `UPDATE image SET provider='smk' WHERE provider='statensmuseum'` in production
* Update to SMK & enable in the Django admin UI

Not so bad :) I'll update it.

stacimc:
> Have we or should we deleted their existing data as part of this PR, or sometime in the future when we deploy it?

Since it's not a dated DAG, we should automatically re-consume everything and update all the URLs on the next ingestion. Now you mention it though, it would be just as easy to just delete all the `statensmuseum` data first (rather than updating the provider field). It will all get backfilled, and then we don't have to worry about any edge cases with records that may have been ingested previously and then removed from the SMK API.

openverse_catalog/dags/common/loader/provider_details.py
stacimc: Alternatively we can just [delete everything for statensmuseum](https://github.com/WordPress/openverse-catalog/pull/506#issuecomment-1129122059) and let the next data pull re-populate it with the correct provider name and urls.

PR #330, opened 2022-01-19
Freesound SSLError fix

AetherUnbound:
## Description

When running the Freesound provider API script, we kept getting the following error:

```python
Traceback (most recent call last):
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/urllib3/connectionpool.py", line 699, in urlopen
    httplib_response = self._make_request(
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/urllib3/connectionpool.py", line 382, in _make_request
    self._validate_conn(conn)
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/urllib3/connectionpool.py", line 1010, in _validate_conn
    conn.connect()
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/urllib3/connection.py", line 416, in connect
    self.sock = ssl_wrap_socket(
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/urllib3/util/ssl_.py", line 449, in ssl_wrap_socket
    ssl_sock = _ssl_wrap_socket_impl(
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/urllib3/util/ssl_.py", line 493, in _ssl_wrap_socket_impl
    return ssl_context.wrap_socket(sock, server_hostname=server_hostname)
  File "/usr/local/lib/python3.9/ssl.py", line 500, in wrap_socket
    return self.sslsocket_class._create(
  File "/usr/local/lib/python3.9/ssl.py", line 1040, in _create
    self.do_handshake()
  File "/usr/local/lib/python3.9/ssl.py", line 1309, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLEOFError: EOF occurred in violation of protocol (_ssl.c:1129)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/requests/adapters.py", line 439, in send
    resp = conn.urlopen(
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/urllib3/connectionpool.py", line 755, in urlopen
    retries = retries.increment(
  File "/usr/local/airflow/.local/lib/python3.9/site-packages/urllib3/util/retry.py", line 574, in increment
    raise MaxRetryError(_pool, url, error or ResponseError(cause))
[...]

PR #508, opened 2022-05-13
Retry flaky request when Smithsonian provider script detects no unit codes

### LINKED ISSUE #504: Smithsonian subproviders check reports all unit codes as outdated
## Description

The Smithsonian provider DAG [validates its list of subproviders](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_api_scripts/smithsonian.py#L132) (or unit codes) before attempting to pull data. Sometimes, it reports that _all_ known unit codes are outdated and must be deleted. It seems like the Smithsonian API might be flaky. We should detect this and retry.

## Reproduction

1. Run the Smithsonian provider script locally a few times, until you see the error.

You can also look at the last few runs in production.

## Resolution

stacimc:
## Fixes

Fixes #504 by @stacimc 

## Description

The Smithsonian provider script uses a list of sub-providers to map records to their sources. Before attempting to pull data, the provider script compares its own sub-provider list with the list of active sub-providers from the API, and alerts if there is a mismatch. Unfortunately this API appears to be flaky and sometimes returns an empty list of unit codes, so the Smithsonian script fails immediately and suggests that we should delete all of our subproviders 🥲 

This PR just adds a retry that detects when we get an empty list of codes from the API and tries again up to 3 times.

## Testing Instructions

This is _very flaky_ and cannot be reliably reproduced, but at minimum you can make sure the happy path still works by simply running the Smithsonian provider script a few times locally (you only need to get as far as making sure it starts to write lines to disk, then you can mark the DAG as successful).

I've personally tested this several times and could never reproduce the flakiness locally until today 🎉 Today I happened to be able to reproduce the bug, and then confirm the fix by checking the logs in my `pull_data` task and seeing that the Retry ran:

```
[2022-05-13, 22:26:54 UTC] {smithsonian.py:220} INFO - No unit codes received. Retrying.
```

And the task was able to start pulling data afterward.

If you aren't able to reproduce it naturally, you can comment out the line where we get the unit codes and manually hardcode it to receive an empty set (`unit_codes_from_api = set()`). Then you can at least verify that the retries happen as expected in the logs:

```
[2022-05-17, 17:51:30 UTC] {api.py:40} WARNING - No unit codes received., retrying in 1 seconds...
[2022-05-17, 17:51:35 UTC] {api.py:40} WARNING - No unit codes received., retrying in 2 seconds...
```

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/smithsonian.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/smithsonian.py
@@ -20,6 +20,7 @@ from common.licenses import LicenseInfo
 from common.loader import provider_details as prov
 from common.requester import DelayedRequester
 from common.storage.image import ImageStore
+from retry import retry
 
 
 logger = logging.getLogger(__name__)
@@ -202,23 +203,18 @@ def get_new_and_outdated_unit_codes(
     return new_unit_codes, outdated_unit_codes
 
 
+@retry(ValueError, tries=3, delay=1, backoff=2)
 def get_unit_codes_from_api(
     units_endpoint: str = UNITS_ENDPOINT, retries: int = 3
 ) -> set:
     query_params = {"api_key": API_KEY, "q": "online_media_type:Images"}
-    unit_codes_from_api = set()
+    response_json = delayed_requester.get_response_json(
+        units_endpoint, query_params=query_params
+    )
+    unit_codes_from_api = set(response_json.get("response", {}).get("terms", []))
 
-    if retries > 0:
-        response_json = delayed_requester.get_response_json(
-            units_endpoint, query_params=query_params
-        )
-        unit_codes_from_api = set(response_json.get("response", {}).get("terms", []))
-
-        # The API is flaky and sometimes returns an empty list of unit codes.
-        # Retry a few times if this happens.
-        if len(unit_codes_from_api) == 0:
-            logger.info("No unit codes received. Retrying.")
-            unit_codes_from_api = get_unit_codes_from_api(units_endpoint, retries - 1)
+    if len(unit_codes_from_api) == 0:
+        raise ValueError("No unit codes received.")
 
     return unit_codes_from_api
 
--- a/openverse_catalog/dags/providers/provider_api_scripts/smithsonian.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/smithsonian.py
@@ -217,6 +217,7 @@ def get_unit_codes_from_api(
         # The API is flaky and sometimes returns an empty list of unit codes.
         # Retry a few times if this happens.
         if len(unit_codes_from_api) == 0:
+            logger.info("No unit codes received. Retrying.")
             unit_codes_from_api = get_unit_codes_from_api(units_endpoint, retries - 1)
 
     return unit_codes_from_api
--- a/openverse_catalog/dags/providers/provider_api_scripts/smithsonian.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/smithsonian.py
@@ -216,7 +216,7 @@ def get_unit_codes_from_api(
 
         # The API is flaky and sometimes returns an empty list of unit codes.
         # Retry a few times if this happens.
-        if len(unit_codes_from_api) == 0 and retries > 0:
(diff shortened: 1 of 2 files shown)


openverse_catalog/dags/providers/provider_api_scripts/smithsonian.py
```diff
@@ -202,16 +202,33 @@ def get_new_and_outdated_unit_codes(
...
 
 
+def get_unit_codes_from_api(
+    units_endpoint: str = UNITS_ENDPOINT, retries: int = 3
+) -> set:
+    query_params = {"api_key": API_KEY, "q": "online_media_type:Images"}
+    unit_codes_from_api = set()
+
+    if retries > 0:
+        response_json = delayed_requester.get_response_json(
```
obulat, line 212: Does it return an otherwise correct but simply empty response?
`delayed_requester.get_response_json` already has the functionality of retrying if the response is None. It might be possible to add an optional condition to `get_response_json` so that it checks for `terms` not being empty instead of the response being non-None, and reusing that.

obulat (approved):

openverse_catalog/dags/providers/provider_api_scripts/smithsonian.py
zackkrida: > Does it return an otherwise correct but simply empty response?

That's correct Olga, it returns a healthy response of an empty array.

zackkrida (approved):
looks great!

openverse_catalog/dags/providers/provider_api_scripts/smithsonian.py
stacimc: Yep, unfortunately! We're getting a `200` status with `terms: []` 🙃 Since the response is completely healthy and even contains the `terms` array that we want, this problem is very specific to Smithsonian and I think `get_response_json` probably should not be aware of it. It is very strange!

openverse_catalog/dags/providers/provider_api_scripts/smithsonian.py
```diff
@@ -202,16 +202,33 @@ def get_new_and_outdated_unit_codes(
     return new_unit_codes, outdated_unit_codes
 
 
+def get_unit_codes_from_api(
```
AetherUnbound, line 205: We added the [`retry` library](https://pypi.org/project/retry/) a while back in #330 to address similar flakiness with the Freesound provider script. It might be worth using the `@retry` decorator and raising a specific error if the list returned is empty. That way we 1) put some automatic delay in between our requests and 2) raise an error if we _still_ get an empty list after several retries. I think the current implementation would pass the empty list if 3 retries resulted in an empty response, and leave us with the same slack notifications we're receiving now 😅

tests/dags/providers/provider_api_scripts/test_smithsonian.py
```diff
@@ -59,6 +59,17 @@ def test_alert_new_unit_codes():
...
 
+def test_get_unit_codes_from_api_retries():
+    mock_response = {"terms": []}
+    # Patch the sleep function so it doesn't take long
+    with patch.object(
+        si.delayed_requester, "get_response_json", return_value=mock_response
+    ) as request_patch, patch("time.sleep"):
+        with pytest.raises(ValueError, match="No unit codes received."):
+            si.get_unit_codes_from_api()
+            assert request_patch.call_count == 3
```
AetherUnbound, line 70: I think if `si.get_unit_codes_from_api()` raises a `ValueError`, then the assert statement won't be reached right? Maybe it should be outside the `pytest.raises` context so it occurs after the `ValueError` is caught?

sarayourfriend: Yes, Madison is correct:

https://github.com/pytest-dev/pytest/blob/main/src/_pytest/python_api.py#L838-L848

stacimc: Oops, that's right of course -- oversight, fixed :)

PR #526, opened 2022-05-20
Remove thumbnails from images

### LINKED ISSUE #519: Remove `thumbnail` field for images from the catalog
## Current Situation

We get the thumbnail URL field from the providers but do not use it because we use the thumbnail proxy instead.

## Suggested Improvement

We should stop retrieving thumbnails from providers.
There are several things that need to be addressed here:
- [ ] Remove the `thumbnail` from provider scripts.
- [ ] We probably do not need to remove `thumbnail` from the common columns (TSV and DB) as audio (and maybe other media types) use them. I think it would be easier to set the image thumbnails to NULL.
- [ ] Clean up the database?

## Benefit

Slightly-improved performance and less data to store, data consistency.

## Implementation

obulat:
## Fixes

Fixes #519 by @obulat

## Description

This PR removes `thumbnail_url` from image provider scripts, and hard-codes `thumbnail_url` to `None` in the imageStore class.

## Testing Instructions

Run the tests - they should pass. Try running the image DAGs and see that they save image thumbnails as NULL.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


openverse_catalog/dags/providers/provider_api_scripts/wordpress.py
```diff
@@ -229,7 +229,6 @@ def _extract_image_data(media_data):
     image_url, height, width, filetype = _get_file_info(image_details)
     if image_url is None:
         return None
-    thumbnail = _get_thumbnail_url(image_details)
```
stacimc, line 232: It looks like we can delete the `_get_thumbnail_url` method as well.

stacimc (changes requested):
Excited for all this code cleanup! Just a few comments. The only other `thumbnail` references I'm seeing are in the archive/commoncrawl, which is not being used at the moment and seems like a separate can of worms.

Could we add a test that the thumbnail gets hard-coded to `None`, maybe in the [image store tests](https://github.com/WordPress/openverse-catalog/blob/main/tests/dags/common/storage/test_image.py)?

openverse_catalog/dags/providers/provider_api_scripts/jamendo.py
```diff
@@ -114,7 +115,22 @@ def _get_batch_json(
         return None
     else:
         results = response_json.get("results")
-        return results
+        return force_duplicates(results)
+
+
+# Takes some results and forcibly adds some duplicates to them
+def force_duplicates(results):
```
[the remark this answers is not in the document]
obulat: Do you mean all of the `force_duplicates` function? I didn't look into it much, but it seems like it should be a part of tests, not the script itself...
Should I remove this function and logs above it?

openverse_catalog/dags/providers/provider_api_scripts/wordpress.py
obulat: Thank you for catching this :)

openverse_catalog/dags/providers/provider_api_scripts/jamendo.py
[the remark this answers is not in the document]
stacimc: I assumed all of the changes to this file were by mistake. I believe the `force_duplicates` function was actually a patch I provided a while ago in a different PR where I had made some changes to the `loader` code -- it was a function I was using to manually force there to be some duplicate results, so we could test that the loader successfully removed them. (I may be remembering the details slightly wrong 😄) I didn't intend for it to get merged -- looks like it might've accidentally got into this changeset, unless I'm missing something?

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -96,7 +96,6 @@ def _create_args(details, id_):
     args = {
         "foreign_landing_url": details[1],
         "image_url": details[2],
-        "thumbnail_url": details[3],
         "license_info": get_license_info(license_url=details[6]),
```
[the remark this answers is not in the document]
obulat: Thank you for catching this, @stacimc ! I've refactored phylopic script to use a dict instead of a list to make it easier to understand what data is coming from where.

obulat:
> Could we add a test that the thumbnail gets hard-coded to `None`, maybe in the [image store tests](https://github.com/WordPress/openverse-catalog/blob/main/tests/dags/common/storage/test_image.py)?

I tried adding a specific test for `None` in the thumbnail, but decided to simply modify existing tests:
- the public method `ImageStore.add_image` that is used by the provider scripts does not accept the `thumbnail_url` parameter. If you pass it, it throws an error.
- the inner method `ImageStore._get_image` is used by the `add_image`, and by the tests as a short-cut. It was possible to pass a non-null `thumbnail_url` there (even though only the tests would do it), so I added a line to `_get_image` that sets `thumbnail_url` to `None` anyways. That is tested in one of the tests in `test_image`.

I wonder if it was better to simply use `add_image` in tests, without ever using `_get_image`, which would make the tests more like the usage in the real code. Or maybe it doesn't make sense to call `_get_image` with arguments that `add_image` **can** pass it...

openverse_catalog/dags/providers/provider_api_scripts/brooklyn_museum.py
```diff
@@ -171,16 +170,11 @@ def _get_creators(data):
...
+def _get_image_url(image):
     image_url = image.get("largest_derivative_url")
-    if image_url:
-        if "http" not in image_url:
-            image_url = "https://" + image_url
-        thumbnail_url = image.get("thumbnail_url", "")
-        if "http" not in thumbnail_url and thumbnail_url:
-            thumbnail_url = "https://" + thumbnail_url
-    return image_url, thumbnail_url
+    if image_url and "http" not in image_url:
```
AetherUnbound, line 175: Should this maybe be a different check?
```suggestion
    if image_url and not image_url.startswith("http"):
```

openverse_catalog/dags/providers/provider_api_scripts/jamendo.py
```diff
@@ -94,7 +94,8 @@ def _get_items():
     while should_continue:
         query_params = _get_query_params(offset=offset)
         batch_data = _get_batch_json(query_params=query_params)
-        if isinstance(batch_data, list) and len(batch_data) > 0:
+        logger.info(f"batch_data: {batch_data}, query_params: {query_params}")
```
AetherUnbound, line 97: `batch_data` is going to be a pretty large list in most cases, do you think this could be a `logger.debug` line?

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -185,23 +168,23 @@ def _get_meta_data(_uuid):
...
-        img_url,
-        thumbnail,
-        str(width),
-        str(height),
-        license_url,
-        creator,
-        title,
-        meta_data,
-    ]
+    details = {
```
AetherUnbound, line 176: Nice simplification here!

AetherUnbound (commented):
This is great! I love seeing change counts like this 🤩
Screenshot_2022-05-25_07-46-22

This prevents us from acquiring _new_ thumbnail IDs - it would be awesome if we could remove `thumbnail_url` altogether from the catalog and the API. Will there be a follow up PR to remove that column from the table? I think it would be a good idea to do that shortly after this 😄

openverse_catalog/dags/providers/provider_api_scripts/jamendo.py
obulat, line 97:
```suggestion
```

obulat: It was probably left their by mistake :)

obulat:
>This prevents us from acquiring new thumbnail IDs - it would be awesome if we could remove thumbnail_url altogether from the catalog and the API. Will there be a follow-up PR to remove that column from the table? I think it would be a good idea to do that shortly after this 😄
I noted elsewhere, but will also write it here:
We do use `thumbnail_url` for other media types, and currently, it is in the list of `common` columns for all media types. We talked about video and 3D model maybe using thumbnails, too. It might be easier to simply leave it as a common column for all media types, but leave it as `NULL` for images. What do you think?

AetherUnbound (approved):
Totally makes sense! 🚀

obulat:
> Totally makes sense! 🚀 Do you think it might be worth running an update to set the field to `NULL` for all existing images?

Yes, absolutely!

openverse_catalog/dags/providers/provider_api_scripts/jamendo.py
```diff
@@ -94,7 +94,7 @@ def _get_items():
     while should_continue:
         query_params = _get_query_params(offset=offset)
         batch_data = _get_batch_json(query_params=query_params)
-        if isinstance(batch_data, list) and len(batch_data) > 0:
+        if isinstance(batch_data, list) and len(batch_data) > 0 and item_count < LIMIT:
```
stacimc, line 97: What is the purpose of this change? This still looks like testing code that was accidentally left in to me. I believe this will force the jamendo script to stop early after loading the first 200 records.

obulat: It is debug code, and I have no idea how it crept in here. Thank you for catching it, @stacimc !

obulat, line 97:
```suggestion
        if isinstance(batch_data, list) and len(batch_data) > 0:
```

stacimc (approved):
Looks good! 🎉 Look at that beautiful stat for lines changed 😄

PR #540, opened 2022-06-02
Consolidate provider workflows using dynamic DAGs and dataclasses

### LINKED ISSUE #205: Consolidate provider workflows using dynamic DAGs

stacimc:
## Fixes

Fixes #205 by @AetherUnbound 

## Description

**Provider Workflow DAGs:**
* Removes all of the (mostly identical/boilerplate) `<provider>_workflow.py` files in favor of dataclasses and a factory
* Some adjustments to defaults so that they reflect the most commonly used values
    * For example, most DAGs are not `dated`, are scheduled monthly, and have a 24 hr `execution_timeout`
    * Updates the execution_timeout for the following:
        * Freesound from 12 -> 24 hours (a recent run timed out after 12 hours)
        * Stocksnap from 12 -> 24 hours (recent runs approach the timeout)
* Automatically sets the `doc_md` to the docstring of the provider_api_script, if one exists 

**Ingestion DAGs:**
These are DAGs that basically generate a list of days to re-ingest data, and then run what amounts to just the `pull_data` task for each of those days. The `tsv_to_postgres_loader` DAG will actually load it into the catalog DB. Issue #548 discusses changing this to include loading in the same DAG. _This_ PR only seeks to preserve the current state of things:
* Removes all of the boilerplate `<provider>_ingestion_workflow.py` files in favor of using dataclasses and a factory
* Some updates to the `get_reingestion_day_list_list` helper to slightly clarify it and make setting defaults easier
* Test updates

FWIW I also double-confirmed that we will not lose DagRun history when these changes are applied 🎉 

## Testing Instructions

Make sure that all the DAGs load. Try running a few and make sure they write lines to disk. Optionally click through the DAG details and make sure that the appropriate execution_timeouts/etc have been set. 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida:
> FWIW I also double-confirmed that we will not lose DagRun history when these changes are applied 🎉

Awesome!

openverse_catalog/dags/providers/provider_ingestion_workflow_types.py
```diff
@@ -0,0 +1,75 @@
+from dataclasses import dataclass
+from datetime import datetime, timedelta
+
+from providers.provider_workflow_types import ProviderWorkflow
+
+
+@dataclass
+class ProviderIngestionWorkflow(ProviderWorkflow):
```
AetherUnbound, line 8: So great that we can use inheritance here! 💯

openverse_catalog/dags/providers/provider_workflow_types.py
```diff
@@ -0,0 +1,167 @@
+from dataclasses import dataclass
```
AetherUnbound, line 1: This and the ingestion workflow `_types.py` file both end as such but include more than types (since they also include the workflow definitions themselves). Should we rename this to just `provider_workflows.py` (and similarly for ingestion)?

openverse_catalog/dags/providers/provider_workflow_types.py
```diff
@@ -0,0 +1,167 @@
...
+                      the form 'YYYY-MM-DD'. Otherwise, the function
+                      should take no arguments.
+
+    Optional Arguments:
+
+    default_args:      dictionary which is passed to the airflow.dag.DAG
+                       __init__ method.
+    start_date:        datetime.datetime giving the first valid execution
+                       date of the DAG.
+    max_active_tasks:  integer that sets the number of tasks which can
```
AetherUnbound, line 25: This is carried over but just a note for the future - we should probably separate active tasks & active DAG runs into two separate config options. We may have cases where we want to run more than 1 task concurrently within a DAG, but still only have 1 active DAG at a given time.

openverse_catalog/dags/providers/provider_workflow_types.py
```diff
@@ -0,0 +1,167 @@
...
+    execution_timeout: datetime.timedelta giving the amount of time a given data
+                       pull may take.
+    doc_md:            string which should be used for the DAG's documentation markdown
+    media_types:       list describing the media type(s) that this provider handles
+                       (e.g. `["audio"]`, `["image", "audio"]`, etc.)
+    """
+
+    dag_id: str
+    provider_script: str
+    default_args: Optional[Dict] = None
```
AetherUnbound, line 48: Perhaps now might be a good time to tackle this - we either have a binary of "override all arguments, repeating the defaults where necessary" or "just use the defaults". Maybe instead of `dag_defaults = dag_defaults or DAG_DEFAULTS` in the factory function, we could do `dag_defaults = (dag_defaults or {}) | DAG_DEFAULTS` so that we would only have to override the ones we care about? I might be interpreting it all wrong though

openverse_catalog/dags/providers/provider_workflow_types.py
```diff
@@ -0,0 +1,167 @@
...
+    ProviderWorkflow(
+        dag_id="flickr_wokrflow",
+        provider_script="flickr",
+        schedule_string="@daily",
+        dated=True,
+    ),
+    ProviderWorkflow(
+        dag_id="freesound_workflow",
+        provider_script="freesound",
+        # doc_md = freesound.__doc__,
```
AetherUnbound, line 92: Should this be removed?

openverse_catalog/dags/providers/provider_workflow_types.py
```diff
@@ -0,0 +1,167 @@
...
+        media_types=("audio",),
+    ),
+    ProviderWorkflow(
+        dag_id="jamendo_workflow",
```
AetherUnbound, line 96: We could actually add `__post_init__` logic to the `NamedTuple` which would compute the `dag_id`! It could also be overridden for cases where it's not `f"{provider_script}_workflow"`.

openverse_catalog/dags/providers/provider_workflow_types.py
```diff
@@ -0,0 +1,167 @@
...
+    ProviderWorkflow(
+        dag_id="phylopic_workflow",
+        provider_script="phylopic",
+        schedule_string="@weekly",
+        dated=True,
+        execution_timeout=timedelta(hours=12),
+    ),
+    ProviderWorkflow(
+        dag_id="rawpixel_workflow",
+        provider_script="raw_pixel",
```
AetherUnbound, line 126: Hmm, perhaps we should rename the rawpixel script to be `rawpixel` instead for consistency 😅

AetherUnbound (commented):
I LOVE THIS SO MUCH 😭 ❤️ It's seriously such a simplification 💯 

I have a few comments, mostly just to think about further ways we could pair this down!

openverse_catalog/dags/providers/provider_workflow_types.py
stacimc: That's a good idea! We'd just have to rename a few of the provider scripts to avoid losing DAG history. I think rawpixel, Metropolitan, and Walters. But I see no reason not to!

openverse_catalog/dags/providers/provider_workflow_types.py
stacimc: That's a good point! I went ahead and separated them out while I'm here 😄

openverse_catalog/dags/providers/provider_workflow_types.py
stacimc: Ah yes, I didn't notice those factory methods were completely overriding all arguments. It doesn't look like we're actually overriding anything in any of the DAG definitions yet, but that will definitely be a pain when we need to! Done 👍

openverse_catalog/dags/providers/provider_workflow_types.py
stacimc: Yep, oops 😄 Thanks!

openverse_catalog/dags/providers/provider_workflow_types.py
stacimc: Oh actually wait, I fell into the mutable default trap 😅 _Not_ done, I'll push an update

AetherUnbound (approved):
Woohoo! This looks fantastic 🔥

openverse_catalog/dags/common/helpers.py
```diff
@@ -1,27 +1,40 @@
...
 
     `get_reingestion_day_list_list((x_0, y_0), ..., (x_n, y_n))`
 
     The return will be a list of lists of integers. The zeroth inner
     list will be a list of integers counting by x_0, of length y_0. The
     first will be a list of integers counting by x_1, of length y_1,
     and in general, the ith list will be a list of integers counting by
-    x_i of length y_i. The (i-1)th and ith lists are separated by the
-    value x_i.
+    x_i of length y_i. The zeroth list begins counting from 0, and each
```
zackkrida, line 21: TIL `zeroth` is a real word :)

zackkrida (approved):
LGTM! I ran a few DAGs successfully locally. So nice to have the provider workflows defined in one place.

openverse_catalog/dags/common/helpers.py
AetherUnbound: MATHS 😁

Issue #536, opened 2022-05-26
Audit provider scripts to collect `filetype` and `filesize`


obulat:
## Current Situation

Currently, a lot of image provider scripts are not collecting the `filetype` and `filesize` information. This information can improve the frontend performance and make Openverse friendlier to providers by not requiring a `head` request for each image that lacks this information.

## Suggested Improvement

Image DAGs and what data they collect (the list is updated when the PRs are created or merged):

- [x] `smithsonian`: `filetype` ➖ ,`filesize` ➖ The API does not return `filetype`/`filesize`
- [x] `raw_pixel`: `filetype` ➖ ,`filesize` ➖ The API does not return `filetype`/`filesize`
- [x] `museum_victoria`: `filetype` ✅  ,`filesize` ✅  #600
- [x] `nypl`: `filetype` ✅  ,`filesize` ➖ #630 The API does not return `filesize`
- [x] `phylopic`: `filetype` ✅  ,`filesize` ➖ #547 The API does not return `filesize`
- [x] `science_museum`: `filetype` ✅  ,`filesize` ✅  #576
- [x] `smk`: `filetype` ✅  ,`filesize` ✅  #542
- [x] `cleveland_museum_of_art`: `filetype` ✅  ,`filesize` ✅  #537 
- [x] `metropolitan_museum_of_art`: `filetype` ✅  ,`filesize` ➖ #568 The API does not return `filesize`
- [x] `finnish_museums`: `filetype` ➖ ,`filesize` ➖ The API does not return `filetype`/`filesize`

Temporarily disabled DAGs that will need to be fixed later:
- `walters_art_museum`: `filetype` ➖ ,`filesize` ➖ Reason: no API key, #386
- `brooklyn_museum`: `filetype` ➖ ,`filesize` ➖ Reason: no API token, #385
- `europeana`: `filetype` ➖ ,`filesize` ➖ Reason: needs to be refactored for v2, #109 

Scripts that were already collecting `filetype` and `filesize` data:
- ✅  `flickr`: `filetype` ✅ ,`filesize`  ✅  - Uses image GET request
- ✅ `stocksnap`: `filetype` ✅ ,`filesize` ✅  Uses image GET request for filesize, and only has JPG
- ✅ `wikimedia_commons`: `filetype` ✅ ,`filesize` ✅  Has an ad-hoc function to compute filetype, needs to be updated to use the common util one
[...]

PR #543, opened 2022-06-02
Add `filetype` to Science Museum script

obulat:
## Fixes

Related to #536 by @obulat 

## Description

This PR adds `filetype` and `filesize` to the list of data collected from the Science Museum API.

## Testing Instructions

The tests for the script should pass.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/science_museum.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/science_museum.py
@@ -17,7 +17,7 @@ RETRIES = 3
 PROVIDER = prov.SCIENCE_DEFAULT_PROVIDER
 ENDPOINT = "https://collection.sciencemuseumgroup.org.uk/search/"
 
-delay_request = DelayedRequester(delay=DELAY)
+delayed_requester = DelayedRequester(delay=DELAY)
 image_store = ImageStore(provider=PROVIDER)
 
 HEADERS = {"Accept": "application/json"}
@@ -99,7 +99,7 @@ def _get_batch_objects(
         headers = HEADERS.copy()
     data = None
     for retry in range(retries):
-        response = delay_request.get(endpoint, query_param, headers=headers)
+        response = delayed_requester.get(endpoint, query_param, headers=headers)
         try:
             response_json = response.json()
             if "data" in response_json.keys():
@@ -135,7 +135,7 @@ def _handle_object_data(batch_data):
                 continue
             processed = image_data.get("processed")
             source = image_data.get("source")
-            image_url, height, width = _get_image_info(processed)
+            image_url, height, width, filetype, filesize = _get_image_info(processed)
             if image_url is None:
                 continue
 
@@ -151,6 +151,8 @@ def _handle_object_data(batch_data):
                 image_url=image_url,
                 height=height,
                 width=width,
+                filetype=filetype,
+                filesize=filesize,
                 license_info=license_info,
                 creator=creator,
                 title=title,
@@ -171,19 +173,28 @@ def _get_creator_info(obj_attr):
     return creator_info
 
 
+def _get_filesize(image_url):
+    """
+    Get the size of the image in bytes.
+    """
+    resp = delayed_requester.get(image_url)
+    if resp:
+        filesize = int(resp.headers.get("Content-Length", 0))
+        return filesize if filesize != 0 else None
+
+
 def _get_image_info(processed):
-    if processed.get("large"):
-        image = processed.get("large").get("location")
-        measurements = processed.get("large").get("measurements")
-    elif processed.get("medium"):
-        image = processed.get("medium").get("location")
-        measurements = processed.get("medium").get("measurements")
-    else:
-        image = None
-        measurements = None
+    image_data = processed.get("large")
+    if image_data is None:
+        image_data = processed.get("medium", {})
+
+    image = image_data.get("location")
+    measurements = image_data.get("measurements")
+    filetype = image_data.get("format")
     image = check_url(image)
     height, width = _get_dimensions(measurements)
-    return image, height, width
+    filesize = _get_filesize(image)
+    return image, height, width, filetype, filesize
 
 
 def check_url(image_url):
--- a/openverse_catalog/dags/providers/provider_api_scripts/science_museum.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/science_museum.py
@@ -1,4 +1,5 @@
 import logging
+from typing import Dict, Tuple
 
 from common.licenses import get_license_info
 from common.loader import provider_details as prov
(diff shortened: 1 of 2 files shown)


openverse_catalog/dags/providers/provider_api_scripts/science_museum.py
```diff
@@ -171,19 +173,28 @@ def _get_creator_info(obj_attr):
...
 
 
+def _get_filesize(image_url):
+    """
+    Get the size of the image in bytes.
+    """
+    resp = delayed_requester.get(image_url)
+    if resp:
+        filesize = int(resp.headers.get("Content-Length", 0))
+        return filesize if filesize != 0 else None
```
AetherUnbound, line 183: If I'm understanding this correctly, this will make a request for every image we receive from their API. Should this maybe go the way of NYLP and not report `filesize` since it's not provided by their API? And if not, should this maybe be doing a `HEAD` request rather than a full `GET`?

tests/dags/providers/provider_api_scripts/test_science_museum.py
```diff
@@ -156,41 +156,69 @@ def test_creator_info_fail():
...
+        actual_filesize,
+    ) = sm._get_image_info(large_image)
     expected_image = (
         "https://coimages.sciencemuseumgroup.org.uk/images/3/563/"
         "large_1999_0299_0001__0002_.jpg"
     )
     expected_height = 1022
     expected_width = 1536
+    expected_filetype = "jpeg"
+    expected_filesize = None
```
AetherUnbound, line 173: Would it be possible to add a test case where filesize is a non-`None` value?

openverse_catalog/dags/providers/provider_api_scripts/science_museum.py
obulat: I agree with not reporting `filesize` at all, this is a more sustainable approach.
By the way, I was sure I was using the `HEAD` requests everywhere, and not `GET`'s 😵‍💫

openverse_catalog/dags/providers/provider_api_scripts/science_museum.py
```diff
@@ -173,60 +172,51 @@ def _get_creator_info(obj_attr):
...
-    if measurements:
-        dimensions = measurements.get("dimensions")
-        if dimensions:
-            for dim in dimensions:
-                height_width[dim.get("dimension")] = dim.get("value")
-    return height_width.get("height"), height_width.get("width")
+def _get_dimensions(image_data: Dict) -> Tuple[int | None, int | None]:
+    """
+    Returns the height and width of the image from a list of dictionaries
+    with keys of "dimension", "units", "value".
```
AetherUnbound, line 202: This is not super important here, but we could use a [`TypedDict`](https://peps.python.org/pep-0589/) for cases like this! Also, it seems like it has a `measurements` key with a `dimensions` key inside the dictionary there?

openverse_catalog/dags/providers/provider_api_scripts/science_museum.py
```diff
@@ -173,60 +172,51 @@ def _get_creator_info(obj_attr):
...
+def _get_license_version(source: Dict | None) -> str | None:
     license_version = None
     if source:
-        legal = source.get("legal")
-        if legal:
-            rights = legal.get("rights")
-            if type(rights) == list:
-                license_version = rights[0].get("usage_terms")
+        rights = source.get("legal", {}).get("rights")
+        if type(rights) == list and len(rights) > 0:
```
AetherUnbound, line 218: We should use `isinstance` here because it's [faster and will respect inheritance](https://switowski.com/blog/type-vs-isinstance)!

openverse_catalog/dags/providers/provider_api_scripts/science_museum.py
```diff
@@ -171,51 +172,51 @@ def _get_creator_info(obj_attr):
...
-        if dimensions:
-            for dim in dimensions:
-                height_width[dim.get("dimension")] = dim.get("value")
-    return height_width.get("height"), height_width.get("width")
+def check_url(image_url: str | None) -> str | None:
+    if not image_url:
+        return None
+    if image_url.startswith("http"):
+        return image_url
+    else:
```
AetherUnbound, line 195: This `else` could probably be dropped in this case!

openverse_catalog/dags/providers/provider_api_scripts/science_museum.py
obulat: Thank you for the link, it looks interesting! Can we leave it for the next refactor 😉

I've changed the signature of this function. Instead of trying to get the `measurements` in one of two places using `measurements = processed.get("large / medium").get("measurements")`, I call the `_get_dimensions` with `image_data` we get using ``measurements = processed.get("large / medium")`, and then extract the measurements inside this function:
`height, width = _get_dimensions(image_data)`

AetherUnbound: Yes, another time on that one! Just wanted to mention it :) 

And that works great! But I think the documentation should still read `with keys of "measurement"."dimensions", "units", and "value".` (since the top level key is still `measurements` rather than `dimensions`).

openverse_catalog/dags/providers/provider_api_scripts/science_museum.py
```diff
@@ -173,60 +173,50 @@ def _get_creator_info(obj_attr):
...
+def _get_license_version(source: Dict | None) -> str | None:
     license_version = None
     if source:
-        legal = source.get("legal")
-        if legal:
-            rights = legal.get("rights")
-            if type(rights) == list:
-                license_version = rights[0].get("usage_terms")
+        rights = source.get("legal", {}).get("rights")
+        if isinstance(rights, list) and len(rights) > 0:
```
AetherUnbound, line 218: Nice simplification! Maybe this could even be?
```suggestion
        rights = source.get("legal", {}).get("rights", [])
        if rights:
```

This defaults to a list and just does a truthy check on the list to see if it has items. I guess the `isinstance` check would be important if `rights` could ever be an truthy object that's _not_ a list, like a dict or something 🤔

AetherUnbound (approved):
Just a couple more notes, otherwise this looks good!

obulat:
> And that works great! But I think the documentation should still read with keys of "measurement"."dimensions", "units", and "value". (since the top level key is still measurements rather than dimensions).

Ah, I didn't even see the docstring! 🫢

openverse_catalog/dags/providers/provider_api_scripts/science_museum.py
```diff
@@ -171,51 +173,50 @@ def _get_creator_info(obj_attr):
...
+    if not image_url:
+        return None
+    if image_url.startswith("http"):
+        return image_url
+    return f"https://coimages.sciencemuseumgroup.org.uk/images/{image_url}"
+
+
+def _get_dimensions(image_data: Dict) -> Tuple[int | None, int | None]:
+    """
+    Returns the height and width of the image from a list of dictionaries
```
obulat, line 201:
```suggestion
    Returns the height and width of the image from "image_data"."measurements"
```

obulat:
I am going to draft this for now so that the reviewers are not bothered, and if the new Provider Data Ingester is merged today, I'll move the work from here to #576.

PR #545, opened 2022-06-03
Add a helper function to extract extension from the media URL

obulat:
## Fixes

Related to #536 by @obulat

## Description

This PR extracts the function to get an extension (`filetype`) from the media URL. It also adds `filetype` validation to make sure that we use a single variant of the `filetype`s that have several valid variants: `jpg` for `jpeg/jpg` and `tiff` for `tif/tiff`.

## Testing Instructions

This PR is the basis for the other data normalization PRs, and it is tested there :)

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/extensions.py
+++ b/openverse_catalog/dags/common/extensions.py
@@ -1,0 +1,18 @@
+from typing import Literal
+
+
+EXTENSIONS = {
+    "image": ["jpg", "jpeg", "png", "gif", "bmp", "webp", "tiff", "tif", "svg", "webp"],
+    "audio": ["mp3", "ogg", "wav", "aiff", "flac", "aac", "m4a", "wma", "mp4", "m4b"],
+}
+
+
+def extract_filetype(url: str, media_type: Literal["image", "audio"]) -> str | None:
+    """
+    Extracts the filetype from a media url extension.
+    """
+    possible_filetype = url.split(".")[-1]
+    if possible_filetype in EXTENSIONS[media_type]:
+        return possible_filetype
+    else:
+        return None
--- a/openverse_catalog/dags/common/extensions.py
+++ b/openverse_catalog/dags/common/extensions.py
@@ -1,5 +1,5 @@
 EXTENSIONS = {
-    "image": {"jpg", "jpeg", "png", "gif", "bmp", "webp", "tiff", "tif", "svg", "webp"},
+    "image": {"jpg", "jpeg", "png", "gif", "bmp", "webp", "tiff", "tif", "svg"},
     "audio": {"mp3", "ogg", "wav", "aiff", "flac", "wma", "mp4", "aac", "m4a", "m4b"},
 }
 
--- a/openverse_catalog/dags/common/extensions.py
+++ b/openverse_catalog/dags/common/extensions.py
@@ -1,17 +1,14 @@
-from typing import Literal
-
-
 EXTENSIONS = {
     "image": {"jpg", "jpeg", "png", "gif", "bmp", "webp", "tiff", "tif", "svg", "webp"},
-    "audio": {"mp3", "ogg", "wav", "aiff", "flac", "aac", "m4a", "wma", "mp4", "m4b"},
+    "audio": {"mp3", "ogg", "wav", "aiff", "flac", "wma", "mp4", "aac", "m4a", "m4b"},
 }
 
 
-def extract_filetype(url: str, media_type: Literal["image", "audio"]) -> str | None:
+def extract_filetype(url: str, media_type: str) -> str | None:
     """
     Extracts the filetype from a media url extension.
     """
     possible_filetype = url.split(".")[-1]
-    if possible_filetype in EXTENSIONS[media_type]:
+    if possible_filetype in EXTENSIONS.get(media_type, {}):
         return possible_filetype
     return None
--- a/openverse_catalog/dags/common/extensions.py
+++ b/openverse_catalog/dags/common/extensions.py
@@ -2,8 +2,8 @@ from typing import Literal
 
 
 EXTENSIONS = {
-    "image": ["jpg", "jpeg", "png", "gif", "bmp", "webp", "tiff", "tif", "svg", "webp"],
-    "audio": ["mp3", "ogg", "wav", "aiff", "flac", "aac", "m4a", "wma", "mp4", "m4b"],
+    "image": {"jpg", "jpeg", "png", "gif", "bmp", "webp", "tiff", "tif", "svg", "webp"},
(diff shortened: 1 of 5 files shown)


openverse_catalog/dags/common/extensions.py
```diff
@@ -0,0 +1,17 @@
+from typing import Literal
+
+
+EXTENSIONS = {
+    "image": {"jpg", "jpeg", "png", "gif", "bmp", "webp", "tiff", "tif", "svg", "webp"},
+    "audio": {"mp3", "ogg", "wav", "aiff", "flac", "aac", "m4a", "wma", "mp4", "m4b"},
+}
+
+
+def extract_filetype(url: str, media_type: Literal["image", "audio"]) -> str | None:
```
AetherUnbound, line 10: This looks great! Do you mind adding tests for these specific functions? Or do those tests exist in other PRs?

openverse_catalog/dags/common/extensions.py
```diff
@@ -0,0 +1,17 @@
+from typing import Literal
+
+
+EXTENSIONS = {
+    "image": {"jpg", "jpeg", "png", "gif", "bmp", "webp", "tiff", "tif", "svg", "webp"},
```
sarayourfriend, line 5:
```suggestion
    "image": {"jpg", "jpeg", "png", "gif", "bmp", "webp", "tiff", "tif", "svg"},
```

openverse_catalog/dags/common/extensions.py
```diff
@@ -0,0 +1,17 @@
...
+    "audio": {"mp3", "ogg", "wav", "aiff", "flac", "aac", "m4a", "wma", "mp4", "m4b"},
```
sarayourfriend, line 6: I wonder if treating `m4a` and `mp4` as distinct file types for _audio_ is safe. `mp4` files that are audio only are _supposed_ to be rare, m4a is really the correct extension for those types of files. `mp4` is generally used for audio + video (and maybe subtitles and other things), i.e., an AV container format.

It makes me wonder if our search would be a lot more useful if we didn't use just file extensions but actually merged into _format_, which is more useful information than filetype. It probably requires more work though than just checking out the URL (for example, actually downloading each file and checking the headers manually or with tools). Wikimedia has some good code we can check out in this regard.

It'd probably be more useful, for example, to merge `m4a` and audio-only `mp4`s into a general MPEG-4 Part 14 "format". We'd get even more mileage out of this if we extracted container information like the audio format in ogg files or the type of wma, wma uses the same extension for [wma's mp3 equivalent (lossy), wma's lossless implementation, and two other lesser used wma codecs](https://en.wikipedia.org/wiki/Windows_Media_Audio#Codecs). In a sense the `wma` "filetype" is actually not useful in many cases... maybe you only care to check out lossless files, in that case you'd have to comb through Oggs, WMAs and FLACs, and only one of those is definitely lossless! The others are only _potentially_ lossless, but our current approach won't help anyone find that information out without manually downloading and inspecting each one.

I'm not sure it is something we can solve today, maybe we'd need a lot more work to do this, but I think it is important to consider the meaning behind these filetypes and how we convey them to users and what users could be looking for. If we just extract the "filetype" from the extension, then we probably shouldn't call it filetype and should call it "extension", it's more accurate that way, at least for audio.

Thankfully images (I think) are more straightforward (though there are multiple JPEG standards) and the extension does generally correspond 1-to-1 to a specific format.

obulat: Thank you for all the context about the `filetype`/`format`/`extension` meanings, @sarayourfriend!
Extracting the `format` information can be a project for some time in the future as it would require a lot of changes in how we download the data.
I think a good interim solution could be to change the frontend wording for audio to use `extension` instead of `filetype`, and add a note about extensions not always corresponding to the format of the API.

AetherUnbound (commented):
This looks great! Do you know why the tests are still failing?

Yes :( I've changed the MockImageStore's `add_item` so that it's returning the item instead of the number of items. And there are several tests that are expecting the number.

obulat:
> This looks great! Do you know why the tests are still failing?


Yes :( I've changed the MockImageStore's `add_item` so that it's returning the item instead of the number of items. And there are several tests that are expecting the number.

openverse_catalog/dags/common/storage/image.py
```diff
@@ -194,10 +197,19 @@ def __init__(
...
-        return self._get_image(**image_data)
+        image = self._get_image(**image_data)
+        if image is not None:
+            self.media_buffer.append(image)
+        return len(self.media_buffer)
+
+    def get_item(self):
+        if len(self.media_buffer):
+            return self.media_buffer[0]
+        return None
```
AetherUnbound, line 215: I wonder if it might be best to just use `media_buffer` directly in tests, rather than a proxy method. That way if multiple items are added, the test can index into each one (or whichever one it needs). If an `IndexError` is raised, that'll fail the test too 😄

obulat: Thank you, this will be much better!
This actually should be out of scope for this PR. Maybe I should actually create a new one.
My main goal was to test the ImageStore class the way it works, and not its inner implementation. Currently, many tests rely on the internal `_get_image` method to test that the image data is handled correctly. However, it is much more robust to test the available API methods, and not the internal implementation (which can and should be able to change).
We could probably add some helper methods to the MockImageStore class that would extract the image details from the `_media_buffer` list of TSV rows. I wonder if it would be better to make the `_media_buffer` in the `ImageStore` a public field...

obulat: @AetherUnbound , I've decided that it's better to improve the tests in a different PR, and used `_clean_metadata` for the specific test.

obulat:
I set the priority for this PR as high because there are several other PRs for adding filetype to provider scripts that depend on it.

openverse_catalog/dags/common/storage/image.py
AetherUnbound: Sounds great!

AetherUnbound (approved):
🚀

stacimc (approved):
LGTM, and sounds good about [deferring further test cleanup](https://github.com/WordPress/openverse-catalog/pull/545#discussion_r893354770) to a future PR 👍

obulat:
> LGTM, and sounds good about [deferring further test cleanup](https://github.com/WordPress/openverse-catalog/pull/545#discussion_r893354770) to a future PR 👍

Thank you, @stacimc! I've opened an issue for that.

PR #547, opened 2022-06-03
Add `filetype` to Phylopic script

obulat:
## Fixes

Related to #536 by @obulat 

## Description

This PR adds `filetype`  to the list of data collected from the Phylopic API.

## Testing Instructions

The tests for the script should pass.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
@@ -13,6 +13,7 @@ Notes:                  http://phylopic.org/api/
 import argparse
 import logging
 from datetime import date, timedelta
+from typing import Dict
 
 from common.licenses import get_license_info
 from common.requester import DelayedRequester
@@ -83,15 +84,22 @@ def main(date_start: str = "all", date_end: str = None):
 def _add_data_to_buffer(**kwargs):
     endpoint = _create_endpoint_for_IDs(**kwargs)
     IDs = _get_image_IDs(endpoint)
+    if IDs is None:
+        return
 
     for id_ in IDs:
-        if id_ is not None:
-            details = _get_meta_data(id_)
-            if details is not None:
-                image_store.add_item(**details)
+        item_data = _get_object_json(id_)
+        if item_data:
+            _process_item(item_data)
 
 
-def _get_total_images():
+def _process_item(item_data: Dict) -> None:
+    details = _get_meta_data(item_data)
+    if details is not None:
+        image_store.add_item(**details)
+
+
+def _get_total_images() -> int:
     # Get the total number of PhyloPic images
     total = 0
     endpoint = "http://phylopic.org/api/a/image/count"
@@ -124,7 +132,7 @@ def _create_endpoint_for_IDs(**kwargs):
     return endpoint
 
 
-def _get_image_IDs(_endpoint):
+def _get_image_IDs(_endpoint) -> list | None:
     result = delayed_requester.get_response_json(_endpoint, retries=2)
     image_IDs = []
 
@@ -132,31 +140,35 @@ def _get_image_IDs(_endpoint):
         data = list(result.get("result"))
 
         if len(data) > 0:
-            for i in range(len(data)):
-                image_IDs.append(data[i].get("uid"))
+            for item in data:
+                image_IDs.append(item.get("uid"))
 
     if not image_IDs:
         logger.warning("No content available!")
-        return [None]
+        return None
 
     return image_IDs
 
 
-def _get_meta_data(_uuid):
+def _get_object_json(_uuid: str) -> Dict | None:
     logger.info(f"Processing UUID: {_uuid}")
 
-    base_url = "http://phylopic.org"
-    meta_data = {}
     endpoint = (
         f"http://phylopic.org/api/a/image/{_uuid}?options=credit+"
         "licenseURL+pngFiles+submitted+submitter+taxa+canonicalName"
         "+string+firstName+lastName"
     )
+    result = None
     request = delayed_requester.get_response_json(endpoint, retries=2)
     if request and request.get("success") is True:
         result = request["result"]
-    else:
-        return None
+    return result
+
+
+def _get_meta_data(result: Dict) -> Dict:
+    base_url = "http://phylopic.org"
+    meta_data = {}
+    _uuid = result.get("uid")
     license_url = result.get("licenseURL")
 
     meta_data["taxa"], title = _get_taxa_details(result)
--- a/openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
@@ -14,7 +14,6 @@ import argparse
 import logging
 from datetime import date, timedelta
 
-import requests
 from common.licenses import get_license_info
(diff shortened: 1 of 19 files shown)


openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -220,11 +221,20 @@ def _get_taxa_details(result):
     return taxa_list, title
 
 
+def _get_filesize(img_url):
+    filesize = requests.head(img_url).headers.get("content-length", "")
+    try:
+        return int(filesize)
+    except ValueError:
+        return None
```
AetherUnbound, line 229: Same question here - should we neglect this for now since we have to request each image? I think if we're going to be doing this in other cases, we may want to add a `DelayedRequester::head` function if it doesn't already have one.

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -220,11 +221,20 @@ def _get_taxa_details(result):
     return taxa_list, title
 
 
+def _get_filesize(img_url):
+    filesize = requests.head(img_url).headers.get("content-length", "")
+    try:
+        return int(filesize)
```
AetherUnbound, line 227: I think if this returns a string, you can use [`str::isdigit`](https://docs.python.org/3/library/stdtypes.html#str.isdigit) here rather than a try/except block for doing this conversion!

tests/dags/providers/provider_api_scripts/test_phylopic.py
```diff
@@ -1,23 +1,44 @@
...
+        yield json.load(f)
+
+
+@pytest.fixture(autouse=True)
+def file_size_patch():
+    with patch(
+        "providers.provider_api_scripts.phylopic._get_filesize"
+    ) as _get_filesize_mock:
+        _get_filesize_mock.return_value = IMAGE_FILESIZE
+        yield
```
sarayourfriend, line 35: Doesn't this kind just mask over the testing for the filesize logic a bit? Is there a way for us to mock the actual HEAD request response being made so that we can exercise the logic in `_get_filesize` instead of just mocking it (thereby never actually covering that function's implementation)?

tests/dags/providers/provider_api_scripts/test_phylopic.py
```diff
@@ -129,8 +118,6 @@ def test_get_meta_data_correct(image_data):
             "foreign_landing_url": "http://phylopic.org/image/e9df48fe-68ea-419e-b9df-441e0b208335",
             "width": 847,
             "height": 1024,
-            "filesize": IMAGE_FILESIZE,
-            "filetype": "png",
```
AetherUnbound: Should we still be checking filetype in this test?

AetherUnbound (approved):
Looks great! 🥳

tests/dags/providers/provider_api_scripts/test_phylopic.py
```diff
@@ -1,33 +1,42 @@
...
 
 import pytest
 from common.licenses import get_license_info
 from providers.provider_api_scripts import phylopic as pp
 
 
-RESOURCES = os.path.join(
-    os.path.abspath(os.path.dirname(__file__)), "resources/phylopic"
-)
+RESOURCES = Path(__file__).parent.resolve() / "resources/phylopic"
```
sarayourfriend, line 12: Just for my understanding, why is the `.resolve()` necessary here versus just using `.parent` by itself?

tests/dags/providers/provider_api_scripts/test_phylopic.py
```diff
@@ -1,33 +1,42 @@
...
     format="%(asctime)s - %(name)s - %(levelname)s:  %(message)s",
     level=logging.DEBUG,
 )
 
 
+@pytest.fixture
+def image_data():
+    IMAGE_DATA_EXAMPLE = RESOURCES / "correct_meta_data_example.json"
+    with open(IMAGE_DATA_EXAMPLE) as f:
+        yield json.load(f)
```
sarayourfriend, line 24: Can this fixture be defined using the `get_json` helper?

```suggestion
    yield get_json("correct_meta_data_example.json")
```

tests/dags/providers/provider_api_scripts/test_phylopic.py
```diff
@@ -97,10 +104,10 @@ def test_get_meta_data_for_none_response():
...
 
 
-def test_get_meta_data_correct():
-    with open(os.path.join(RESOURCES, "correct_meta_data_example.json")) as f:
-        r = json.load(f)
-    with patch.object(pp.delayed_requester, "get_response_json", return_value=r):
+def test_get_meta_data_correct(image_data):
+    with patch.object(
+        pp.delayed_requester, "get_response_json", return_value=image_data
+    ):
```
sarayourfriend, line 110: Just a suggestion (because this was already not done in an optimal way) but we could use the [`monkeypatch` fixture helper](https://docs.pytest.org/en/6.2.x/monkeypatch.html) to create a fixture that automatically patches the image_data as the result of `get_response_json`.

Something like this I think?

```
@pytest.fixture(autouse=True)
def return_response_json(monkeypatch, image_data):
    monkeypatch.setattr(pp.delayed_requester, "get_response_json", value=lambda *args: image_data)
```

I don't remember if the lambda will work like that, it may be necessary to define the function separately as a `def` closure before monkeypatching.

Up to you whether to do this though :slightly_smiling_face:

sarayourfriend (commented):
I don't actually see `filetype` being added here. Am I missing something in the diff?

obulat:
> I don't actually see `filetype` being added here. Am I missing something in the diff?

For the providers who don't specifically send filetype, we use the ImageStore to extract the `filetype` from the image URL. So, for Phylopic provider, the `filetype` will be added. But because the tests only look at what the script sends to `ImageStore`, and not what the `ImageStore` saves, this PR doesn't have it. I'll adjust the tests to include the check for `filetype`, too!

tests/dags/providers/provider_api_scripts/test_phylopic.py
obulat: I don't think it's necessary. I think I just copied it from a different file.

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -83,15 +84,22 @@ def main(date_start: str = "all", date_end: str = None):
...
-            details = _get_meta_data(id_)
-            if details is not None:
-                image_store.add_item(**details)
+        item_data = _get_object_json(id_)
+        if item_data:
+            _process_item(item_data)
 
 
-def _get_total_images():
+def _process_item(item_data: Dict) -> None:
```
obulat, line 96: I extracted this method so that I can use it in tests with the saved json responses.

tests/dags/providers/provider_api_scripts/test_phylopic.py
obulat: This will make the tests cleaner, but I think it is a change for a different PR :)

sarayourfriend (approved):
LGTM!

PR #555, opened 2022-06-16
Add base class for Provider API scripts

### LINKED ISSUE #189: [Feature] Auto-generate DAGs-by-schedule-interval list in documentation
## Problem
The DAG lists in the `README.md` are tedious to maintain and might fall out of sync with the actual DAGs we have defined.

## Description
It would be ideal to have a way to auto-generate the documentation (by parsing the `DagBag`, restructuring how we define DAGs, etc) or have a tool to auto-generate it for us.

## Additional context
Came about as part of a discussion in #187 

## Implementation

### LINKED ISSUE #229: [Quality] Delay object creation until runtime

### LINKED ISSUE #248: Ability to load 'small batches' of images
## Problem

Running Provider DAGs locally takes a lot of time, and can download a lot of images to a user's work machine.

## Description

It would be cool to have some configuration flag or other way of getting all provider scripts to download a _small subset_ of images and exit early. This can currently be done by manually killing DAG runs (like marking the `pull_image` task as successful in the ui for the current run) but that's not reasonable. It would be cool if this was abstracted so it automatically works for all provider scripts, like if X tsv lines are written, kill the DAG sort of thing.

## Alternatives

I don't know! There might be an even cooler way to do this I haven't thought of.

## Implementation

stacimc:
## Fixes

Fixes #229, fixes #248

## Description

This PR adds an abstract `ProviderDataIngester` class which encapsulates the `DelayedRequester` and `MediaStore` objects needed for a provider script, and pulls out reusable code for requesting and processing batches of records. It also refactors the Cleveland Museum provider script to use the new class, as a simple initial example.

In doing so it attempts to accomplish a couple things:
* Delays the creation of `DelayedRequester` and `MediaStore` objects until the DAG actually _runs_, rather than creating expensive objects at import time. This is discussed in #229 
* Attempts to pull out as much reusable code as possible, related to making and processing requests. The goal is for individual provider scripts to only need to provide code for API-specific handling, and make it easier to write new provider scripts
* Makes it easier to add new features that apply to all provider scripts. For example, this PR adds an Airflow variable `ingestion_limit` which can be used to restrict the number of records ingested for testing purposes (#248). All provider scripts that use the base class will automatically get this, as opposed to needing to implement it in each one individually.

There is a lot of variability in the way that our provider scripts are written which makes this difficult. I've done an audit of our provider scripts to try to nail down what we'll need to support. For this reason, the code aims to follow the Single Responsibility Principle and is broken into many individual methods. The goal is for each provider to be able to override exactly the parts they need, where they differ from the 'normal' flow. This PR is intentionally kept as small in size as possible, but I've been testing locally with Wikimedia and Finnish as well to make sure it accommodates some of our more divergent APIs.

### What it doesn't do

It does not rigidly define types for the `record` objects as they are added to the `MediaStore`, nor does it do any validation. At present, what validation there is happens in the `MediaStore` itself. I'd love to type more explicitly, but this is challenging without making changes to the `MediaStore` -- which must continue to support all of the current provider scripts.

In all, I think this should be considered **Step One** in a much longer refactoring process. Once all (or at least most) of the provider scripts use the ProviderDataIngester, it will be _much_ easier to pull out and refactor even more -- particularly since doing so at the moment simply requires duplicating code so that it will continue to work for legacy providers. My attempts to do much more here have risked making the PR far too large/complex, so I've pulled back.

**Importantly**, this does not address new providers like iNaturalist that process a data dump instead of data scraping via the API. As we add more providers with this kind of workflow, I think what will likely make the most sense is creating a second `Ingester` base class to encompass shared code for those providers, with a common parent. 

## Testing Instructions

**Ensure that all tests pass**
```
just test
```

**Ensure that you can run an unaffected provider script**
Run the jamendo DAG for example. Wait until you see lines written to buffer and then mark the `pull_data` task as successful.

**Ensure that the Cleveland DAG runs**
Run the Cleveland DAG locally. Wait until you see several 100 lines written to buffer and then mark the `pull_data` task as successful. You can go to the MinIo console and inspect the generated tsvs to see they look correct (for bonus points 😄).

**Ensure the Cleveland DAG runs with an `ingestion_limit`**
Create an Airflow variable through the Admin -> Variables UI. Name it `ingestion_limit` and set it to some reasonably small value (I set `2000`). Now run the Cleveland DAG again, but let it run to completion: it should complete quickly and only ingest the specified number of records. 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/provider_dag_factory.py
+++ b/openverse_catalog/dags/common/provider_dag_factory.py
@@ -56,6 +56,7 @@ import logging
 import os
 import time
 from datetime import datetime, timedelta
+from types import FunctionType
 from typing import Callable, Dict, List, Optional, Sequence
 
 from airflow import DAG
@@ -80,7 +81,7 @@ DATE_RANGE_ARG_TEMPLATE = "{{{{ macros.ds_add(ds, -{}) }}}}"
 
 
 def _push_output_paths_wrapper(
-    func: Callable,
+    ingestion_callable: Callable,
     media_types: List[str],
     ti: TaskInstance,
     args: Sequence = None,
@@ -96,43 +97,65 @@ def _push_output_paths_wrapper(
     """
     args = args or []
     logger.info("Pushing available store paths to XComs")
-    module = inspect.getmodule(func)
-    stores = {}
 
-    # Stores exist at the module level, so in order to retrieve the output values we
-    # must first pull the stores from the module.
-    for media_type in media_types:
-        if not (store := getattr(module, f"{media_type}_store", None)):
-            continue
-        stores[media_type] = store
+    if isinstance(ingestion_callable, FunctionType):
+        # Stores exist at the module level, so in order to retrieve the output values we
+        # must first pull the stores from the module.
+        module = inspect.getmodule(ingestion_callable)
+        stores = {}
+        for media_type in media_types:
+            if not (store := getattr(module, f"{media_type}_store", None)):
+                continue
+            stores[media_type] = store
 
-    if len(stores) != len(media_types):
-        raise ValueError(
-            f"Expected stores in {module.__name__} were missing: "
-            f"{list(set(media_types) - set(stores))}"
-        )
+        if len(stores) != len(media_types):
+            raise ValueError(
+                f"Expected stores in {module.__name__} were missing: "
+                f"{list(set(media_types) - set(stores))}"
+            )
 
-    for media_type, store in stores.items():
-        logger.info(f"{media_type.capitalize()} store location: {store.output_path}")
-        ti.xcom_push(key=f"{media_type}_tsv", value=store.output_path)
+        for media_type, store in stores.items():
+            logger.info(
+                f"{media_type.capitalize()} store location: {store.output_path}"
+            )
+            ti.xcom_push(key=f"{media_type}_tsv", value=store.output_path)
 
-    logger.info("Running provider function")
+        logger.info("Running provider function")
 
-    start_time = time.perf_counter()
-    # Not passing kwargs here because Airflow throws a bunch of stuff in there that none
-    # of our provider scripts are expecting.
-    data = func(*args)
-    end_time = time.perf_counter()
+        start_time = time.perf_counter()
+        # Not passing kwargs here because Airflow throws a bunch of stuff in there that
+        # none of our provider scripts are expecting.
+        data = ingestion_callable(*args)
+        end_time = time.perf_counter()
 
+    else:
+        # A ProviderDataIngester class was passed instead. First we initialize the
+        # class, which will initialize the media stores and DelayedRequester.
+        logger.info(f"Initializing ProviderIngester {ingestion_callable.__name__}")
+        ingester = ingestion_callable()
+
+        # Push the media store output directories to XComs.
+        for store in ingester.media_stores.values():
+            logger.info(
+                f"{store.media_type.capitalize()} store location: {store.output_path}"
+            )
+            ti.xcom_push(key=f"{store.media_type}_tsv", value=store.output_path)
+
+        # Run the `ingest_records` function to perform ingestion.
+        logger.info("Running ingestion function")
+        start_time = time.perf_counter()
+        data = ingester.ingest_records(*args)
+        end_time = time.perf_counter()
+
+    # Report duration
     duration = end_time - start_time
     ti.xcom_push(key="duration", value=duration)
-
     return data
 
 
 def create_provider_api_workflow(
     dag_id: str,
-    main_function: Callable,
+    ingestion_callable: Callable,
     default_args: Optional[Dict] = None,
     start_date: datetime = datetime(1970, 1, 1),
     max_active_runs: int = 1,
@@ -150,12 +173,12 @@ def create_provider_api_workflow(
 
     Required Arguments:
 
-    dag_id:         string giving a unique id of the DAG to be created.
-    main_function:  python function to be run. If the optional argument
-                    `dated` is True, then the function must take a
-                    single parameter (date) which will be a string of
-                    the form 'YYYY-MM-DD'. Otherwise, the function
-                    should take no arguments.
+    dag_id:             string giving a unique id of the DAG to be created.
+    ingestion_callable: python function to be run, or a ProviderDataIngester class
+                        whose `ingest_records` method is to be run. If the optional
+                        argument `dated` is True, then the function must take a
+                        single parameter (date) which will be a string of
+                        the form 'YYYY-MM-DD'.
 
     Optional Arguments:
 
@@ -204,7 +227,10 @@ def create_provider_api_workflow(
     )
 
     with dag:
-        pull_kwargs = {"func": main_function, "media_types": media_types}
+        pull_kwargs = {
+            "ingestion_callable": ingestion_callable,
+            "media_types": media_types,
+        }
         if dated:
             pull_kwargs["args"] = [DATE_RANGE_ARG_TEMPLATE.format(day_shift)]
 
--- a/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Dict, Optional
+from typing import Dict
 
 from common.licenses import get_license_info
 from common.loader import provider_details as prov
@@ -12,17 +12,10 @@ CC0_LICENSE = get_license_info(license_="cc0", license_version="1.0")
 
 
 class ClevelandDataIngester(ProviderDataIngester):
-    def __init__(
-        self,
-        providers={"image": prov.CLEVELAND_DEFAULT_PROVIDER},
-        endpoint="http://openaccess-api.clevelandart.org/api/artworks/",
-        batch_limit=1000,
-        delay=5,
-        retries=3,
-        headers: Optional[Dict] = {},
-    ):
-        # Initialize the DataIngester with appropriate values
-        super().__init__(providers, endpoint, batch_limit, delay, retries, headers)
+    providers = {"image": prov.CLEVELAND_DEFAULT_PROVIDER}
+    endpoint = "http://openaccess-api.clevelandart.org/api/artworks/"
+    batch_limit = 1000
+    delay = 5
 
     def get_next_query_params(self, old_query_params, **kwargs):
         if not old_query_params:
--- a/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
@@ -77,7 +77,8 @@ class ClevelandDataIngester(ProviderDataIngester):
                 return keyed_image
         return None
 
-    def _get_int_value(self, data: Dict, key: str) -> int | None:
+    @staticmethod
+    def _get_int_value(data: Dict, key: str) -> int | None:
         """
         Converts the value of the key `key` in `data` to an integer.
         Returns None if the value is not convertible to an integer, or
--- a/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
@@ -102,3 +102,13 @@ class ClevelandDataIngester(ProviderDataIngester):
         }
         metadata = {k: v for k, v in metadata.items() if v is not None}
         return metadata
+
+
+def main():
+    logger.info("Begin: Cleveland Museum data ingestion")
+    ingester = ClevelandDataIngester()
+    ingester.ingest_records()
+
+
+if __name__ == "__main__":
+    main()
--- a/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
@@ -69,7 +69,8 @@ class ClevelandDataIngester(ProviderDataIngester):
             "meta_data": self._get_metadata(record),
         }
 
-    def _get_image_type(self, image_data):
+    @staticmethod
+    def _get_image_type(image_data):
         # Returns the image url and key for the image in `image_data` dict.
         for key in ["web", "print", "full"]:
             if keyed_image := image_data.get(key):
@@ -90,7 +91,8 @@ class ClevelandDataIngester(ProviderDataIngester):
                 return value
         return None
 
-    def _get_metadata(self, data):
+    @staticmethod
+    def _get_metadata(data):
         metadata = {
             "accession_number": data.get("accession_number", ""),
             "technique": data.get("technique", ""),
--- a/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
@@ -39,6 +39,11 @@ class ClevelandDataIngester(ProviderDataIngester):
         # This provider only supports Images.
         return "image"
 
+    def get_batch_data(self, response_json):
+        if response_json:
+            return response_json.get("data")
+        return None
+
     def get_record_data(self, record):
         license_ = record.get("share_license_status", "").lower()
         if license_ != "cc0":
--- a/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
@@ -1,161 +1,112 @@
 import logging
-from typing import Dict, List
+from typing import Dict, List, Optional
 
 from common.licenses import get_license_info
 from common.loader import provider_details as prov
-from common.requester import DelayedRequester
-from common.storage.image import ImageStore
+from providers.provider_api_scripts.provider_data_ingester import ProviderDataIngester
 
 
-LIMIT = 1000
-DELAY = 5.0
-RETRIES = 3
-PROVIDER = prov.CLEVELAND_DEFAULT_PROVIDER
-ENDPOINT = "http://openaccess-api.clevelandart.org/api/artworks/"
-CC0_LICENSE = get_license_info(license_="cc0", license_version="1.0")
-
-delay_request = DelayedRequester(delay=DELAY)
-image_store = ImageStore(provider=PROVIDER)
-
-DEFAULT_QUERY_PARAMS = {"cc": "1", "has_image": "1", "limit": LIMIT, "skip": 0}
-
-logging.basicConfig(
-    format="%(asctime)s - %(name)s - %(levelname)s:  %(message)s", level=logging.INFO
-)
 logger = logging.getLogger(__name__)
 
+CC0_LICENSE = get_license_info(license_="cc0", license_version="1.0")
 
-def main():
-    logger.info("Begin: Cleveland Museum API requests")
-    condition = True
-    offset = 0
 
-    while condition:
-        query_param = _build_query_param(offset)
-        response_json, total_images = _get_response(query_param)
-        if response_json is not None and total_images != 0:
-            batch = response_json["data"]
-            image_count = _handle_response(batch)
-            logger.info(f"Total images till now {image_count}")
-            offset += LIMIT
+class ClevelandDataIngester(ProviderDataIngester):
+    def __init__(
+        self,
+        providers={"image": prov.CLEVELAND_DEFAULT_PROVIDER},
+        endpoint="http://openaccess-api.clevelandart.org/api/artworks/",
+        delay=5,
+        batch_limit=1000,
+        retries=3,
+        headers: Optional[Dict] = {},
+    ):
+        # Initialize the DataIngester with appropriate values
+        super().__init__(providers, endpoint, delay, batch_limit, retries, headers)
+
+    def get_next_query_params(self, old_query_params, **kwargs):
+        if not old_query_params:
+            # Return default query params on the first request
+            return {"cc": "1", "has_image": "1", "limit": self.batch_limit, "skip": 0}
         else:
-            logger.error("No more images to process")
-            logger.info("Exiting")
-            condition = False
-    image_count = image_store.commit()
-    logger.info(f"Total number of images received {image_count}")
+            # Increment `skip` by the batch limit.
+            return {
+                **old_query_params,
+                "skip": old_query_params["skip"] + self.batch_limit,
+            }
 
+    def get_media_type(self, record):
+        # This provider only supports Images.
+        return "image"
 
-def _build_query_param(offset=0, default_query_param=None):
-    if default_query_param is None:
-        default_query_param = DEFAULT_QUERY_PARAMS
-    query_param = default_query_param.copy()
-    query_param.update(skip=offset)
-    return query_param
+    def get_record_data(self, record):
+        license_ = record.get("share_license_status", "").lower()
+        if license_ != "cc0":
+            logger.error("Wrong license image")
+            return None
 
+        foreign_id = record.get("id")
+        if foreign_id is None:
+            return None
 
-def _get_response(query_param, endpoint=ENDPOINT, retries=RETRIES):
-    response_json, total_images, tries = None, 0, 0
-    for tries in range(retries):
-        response = delay_request.get(endpoint, query_param)
-        if response is not None and response.status_code == 200:
-            try:
-                response_json = response.json()
-                total_images = len(response_json["data"])
-            except Exception as e:
-                logger.warning(f"response not captured due to {e}")
-                response_json = None
-            if response_json is not None and total_images is not None:
-                break
+        image = self._get_image_type(record.get("images", {}))
+        if image is None or image.get("url") is None:
+            return None
 
-        logger.info(
-            "Retrying \n"
-            f"endpoint -- {endpoint} \t"
-            f" with parameters -- {query_param} "
-        )
-    if tries == retries - 1 and ((response_json is None) or (total_images is None)):
-        logger.warning("No more tries remaining. Returning Nonetypes.")
-        return None, 0
-    else:
-        return response_json, total_images
+        if record.get("creators"):
+            creator_name = record.get("creators")[0].get("description", "")
+        else:
+            creator_name = ""
+        
+        return {
+            "foreign_identifier": f"{foreign_id}",
+            "foreign_landing_url": record.get("url"),
+            "title": record.get("title", None),
+            "creator": creator_name,
+            "image_url": image["url"],
+            "width": self._get_int_value(image, "width"),
+            "height": self._get_int_value(image, "height"),
+            "filesize": self._get_int_value(image, "filesize"),
+            "license_info": CC0_LICENSE,
+            "meta_data": self._get_metadata(record),
+        }
 
+    def _get_image_type(self, image_data):
+        key, image_url = None, None
+        if image_data.get("web"):
+            key = "web"
+            image_url = image_data.get("web").get("url", None)
+        elif image_data.get("print"):
+            key = "print"
+            image_url = image_data.get("print").get("url", None)
+        elif image_data.get("full"):
+            key = "full"
+            image_url = image_data.get("full").get("url", None)
+        return image_url, key
 
-def get_int_value(data: Dict, key: str) -> int | None:
-    """
-    Converts the value of the key `key` in `data` to an integer.
-    Returns None if the value is not convertible to an integer, or
-    if the value doesn't exist.
-    """
-    value = data.get(key)
-    if bool(value):
-        if isinstance(value, str) and value.isdigit():
-            return int(value)
-        elif isinstance(value, int):
-            return value
-    return None
-
-
-def _handle_batch_item(data: Dict) -> Dict | None:
-    license_ = data.get("share_license_status", "").lower()
-    if license_ != "cc0":
+    def _get_int_value(self, data: Dict, key: str) -> int | None:
+        """
+        Converts the value of the key `key` in `data` to an integer.
+        Returns None if the value is not convertible to an integer, or
+        if the value doesn't exist.
+        """
+        value = data.get(key)
+        if bool(value):
+            if isinstance(value, str) and value.isdigit():
+                return int(value)
+            elif isinstance(value, int):
+                return value
         return None
-    foreign_id = data.get("id")
-    if foreign_id is None:
-        return None
-    image = _get_image_type(data.get("images", {}))
-    if image is None or image.get("url") is None:
-        return None
 
-    if data.get("creators"):
-        creator_name = data.get("creators")[0].get("description", "")
-    else:
-        creator_name = ""
-
-    return {
-        "foreign_identifier": f"{foreign_id}",
-        "foreign_landing_url": data.get("url"),
-        "title": data.get("title", None),
-        "creator": creator_name,
-        "image_url": image["url"],
-        "width": get_int_value(image, "width"),
-        "height": get_int_value(image, "height"),
-        "filesize": get_int_value(image, "filesize"),
-        "license_info": CC0_LICENSE,
-        "meta_data": _get_metadata(data),
-    }
-
-
-def _handle_response(batch: List):
-    total_images = 0
-    for data in batch:
-        item = _handle_batch_item(data)
-        if item is not None:
-            total_images = image_store.add_item(**item)
-
-    return total_images
-
-
-def _get_image_type(image_data):
-    # Returns the image url and key for the image in `image_data` dict.
-    for key in ["web", "print", "full"]:
-        if keyed_image := image_data.get(key):
-            return keyed_image
-    return None
-
-
-def _get_metadata(data):
-    metadata = {
-        "accession_number": data.get("accession_number", ""),
-        "technique": data.get("technique", ""),
-        "date": data.get("creation_date", ""),
-        "credit_line": data.get("creditline", ""),
-        "classification": data.get("type", ""),
-        "tombstone": data.get("tombstone", ""),
-        "culture": ",".join([i for i in data.get("culture", []) if i is not None]),
-    }
-    metadata = {k: v for k, v in metadata.items() if v is not None}
-    return metadata
-
-
-if __name__ == "__main__":
-    main()
+    def _get_metadata(self, data):
+        metadata = {
+            "accession_number": data.get("accession_number", ""),
+            "technique": data.get("technique", ""),
+            "date": data.get("creation_date", ""),
+            "credit_line": data.get("creditline", ""),
+            "classification": data.get("type", ""),
+            "tombstone": data.get("tombstone", ""),
+            "culture": ",".join([i for i in data.get("culture", []) if i is not None]),
+        }
+        metadata = {k: v for k, v in metadata.items() if v is not None}
+        return metadata
--- a/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Dict, List, Optional
+from typing import Dict, Optional
 
 from common.licenses import get_license_info
 from common.loader import provider_details as prov
@@ -57,7 +57,7 @@ class ClevelandDataIngester(ProviderDataIngester):
             creator_name = record.get("creators")[0].get("description", "")
         else:
             creator_name = ""
-        
+
         return {
             "foreign_identifier": f"{foreign_id}",
             "foreign_landing_url": record.get("url"),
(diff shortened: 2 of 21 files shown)


openverse_catalog/dags/common/provider_dag_factory.py
```diff
@@ -96,43 +97,65 @@ def _push_output_paths_wrapper(
...
-        ti.xcom_push(key=f"{media_type}_tsv", value=store.output_path)
+    if isinstance(ingestion_callable, FunctionType):
+        # Stores exist at the module level, so in order to retrieve the output values we
+        # must first pull the stores from the module.
+        module = inspect.getmodule(ingestion_callable)
+        stores = {}
+        for media_type in media_types:
+            if not (store := getattr(module, f"{media_type}_store", None)):
+                continue
+            stores[media_type] = store
```
sarayourfriend, line 109: Oof, this walrus operator hurts my brain, with `store` being used outside the if body :face_with_head_bandage: 

Can we split the assignment and check into two lines? It would make it clearer how store is intended to be used.

```suggestion
            store = getattr(module, f"{media_type}_store", None)
            if not store:
                continue
            stores[media_type] = store
```

Maybe it's just me but any time I see a walrus operator for an if block I read it as a signal that the variable being assigned (especially if it's also being declared) is only relevant to the if body and not the larger function. Maybe this isn't a pythonic way of thinking considering it layers on block scoping that doesn't exist? I'm not sure.

Update: I'm noticing now that this isn't new code in this PR, just moved around, so feel free to ignore.

sarayourfriend (commented):
This is extremely cool! I'm curious how things like `endpoint` are meant to be used for providers like iNaturalist that won't be interacting with an HTTP API? Would those have a separate base class?

obulat (commented):
This is a great start of a really nice clean-up of our provider scripts. It would be really nice to make them more on-par with each other!

I have opened several PRs for adding `filetype` and `filesize` to the provider scripts. While at it, I also did some refactoring and clean up of the scripts. At first I thought that it might be better to wait for this PR to get merged before merging mine. But then, this PR will probably be a part of the larger process, so it might be a better solution to do it the other way around. In which case, we will probably need to do a lot of merge-conflict resolution work :( Sorry about it.

stacimc:
> I'm curious how things like endpoint are meant to be used for providers like iNaturalist that won't be interacting with an HTTP API? Would those have a separate base class?

Yes, exactly! I'll update the PR description to make that clear, but I do not expect iNaturalist to fit well into this workflow. Ideally in the future we'll have more similar providers, and a separate base class with that flow, possibly with a new common parent. My goal is not to put any restrictions on how that work should look -- I expect it will make more sense to look at adding new base classes after we have a concrete example.

> But then, this PR will probably be a part of the larger process, so it might be a better solution to do it the other way around. In which case, we will probably need to do a lot of merge-conflict resolution work :( Sorry about it.

No worries @obulat, you go ahead! Doesn't look like the conflict resolution should be too bad :)

openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
```diff
@@ -1,97 +1,57 @@
...
-            logger.info(f"Total images till now {image_count}")
-            offset += LIMIT
+class ClevelandDataIngester(ProviderDataIngester):
+    def __init__(
+        self,
+        providers={"image": prov.CLEVELAND_DEFAULT_PROVIDER},
+        endpoint="http://openaccess-api.clevelandart.org/api/artworks/",
+        delay=5,
+        batch_limit=1000,
+        retries=3,
```
AetherUnbound, line 22: What do you think about having these fields as class level attributes rather than parameters for the `__init__` function? Having them here makes me think that we might have cases where we'd override this for a certain run of the Cleveland workflow, which I don't think will be the case. That also prevents us from having to call `super().__init__()` on all of the subclasses!

openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
```diff
@@ -1,97 +1,57 @@
...
+            return {
+                **old_query_params,
+                "skip": old_query_params["skip"] + self.batch_limit,
+            }
+
+    def get_media_type(self, record):
+        # This provider only supports Images.
+        return "image"
+
+    def get_record_data(self, record):
```
AetherUnbound, line 43: I seriously love this!! This is going to make the patterns so much easier to follow 🤩

tests/dags/providers/provider_api_scripts/test_provider_data_ingester.py
```diff
@@ -0,0 +1,181 @@
+import json
+import os
+import unittest
+from unittest.mock import call, patch
+
+from common.storage.audio import AudioStore, MockAudioStore
+from common.storage.image import ImageStore, MockImageStore
+
+from tests.dags.providers.provider_api_scripts.resources.provider_data_ingester.mock_provider_data_ingester import (
```
AetherUnbound, line 9: That's a mouthful! 😂

tests/dags/providers/provider_api_scripts/test_provider_data_ingester.py
```diff
@@ -0,0 +1,181 @@
...
+        assert isinstance(ingester.media_stores["image"], ImageStore)
+
+    def test_batch_limit_is_capped_to_ingestion_limit(self):
+        with patch(
+            "providers.provider_api_scripts.provider_data_ingester.Variable"
+        ) as MockVariable:
+            MockVariable.get.side_effect = ["200"]
+
+            ingester = MockProviderDataIngester()
+            assert ingester.batch_limit == 200
```
AetherUnbound, line 54: Should we also check `ingester.limit` here too? Do you think that's redundant?

tests/dags/providers/provider_api_scripts/test_provider_data_ingester.py
```diff
@@ -0,0 +1,181 @@
...
+            assert image_store_mock.call_count == 2
+
+    def test_ingest_records(self):
+        with (
+            patch.object(self.ingester, "get_batch") as get_batch_mock,
+            patch.object(
+                self.ingester, "process_batch", return_value=3
+            ) as process_batch_mock,
+            patch.object(self.ingester, "commit_records") as commit_mock,
+        ):
```
AetherUnbound, line 80: I'm so glad that you can now have multiple with statements separated by a comma! It's a recent feature as of 3.10[^1] and it make this scenario look so much cleaner 💯 

[^1]: [Changed in version 3.10: Support for using grouping parentheses to break the statement in multiple lines.](https://docs.python.org/3/reference/compound_stmts.html#the-with-statement)

AetherUnbound (commented):
This is _so_ rad! I can't wait to refactor a provider 😁

openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
```diff
@@ -1,97 +1,57 @@
...
-            continue
+            return None
         license_version = "1.0"
 
-        foreign_id = data.get("id")
-        foreign_landing_url = data.get("url", None)
-        image_data = data.get("images", None)
+        foreign_id = record.get("id")
+        foreign_landing_url = record.get("url", None)
+        image_data = record.get("images", None)
```
obulat, line 52: Do you think it's better to explicitly state the default fallback value of `None` for `get` as is done here, or keep it shorter with `record.get("images")`.   I think I prefer the second option, but it's not a very strong preference.

openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
```diff
@@ -101,56 +61,50 @@ def _handle_response(batch):
...
+            "license_info": license_info,
+            "foreign_identifier": foreign_id,
+            "width": width,
+            "height": height,
+            "title": title,
+            "creator": creator_name,
+            "meta_data": metadata,
+        }
+
+    def _get_image_type(self, image_data):
```
obulat, line 86: Pycharm is complaining that this method (and the `_get_metadata` below it) can be `static` because it doesn't use `self`. But I'm not sure what difference it makes.

openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
```diff
@@ -1,97 +1,57 @@
...
-    while condition:
-        query_param = _build_query_param(offset)
-        response_json, total_images = _get_response(query_param)
-        if response_json is not None and total_images != 0:
-            batch = response_json["data"]
-            image_count = _handle_response(batch)
-            logger.info(f"Total images till now {image_count}")
-            offset += LIMIT
+class ClevelandDataIngester(ProviderDataIngester):
+    def __init__(
```
obulat, line 16: Would the `date` parameter be passed to the `Ingester` class?

openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
AetherUnbound: Good find! It seems like it's [generally good practice to use static methods where possible](https://stackoverflow.com/a/22589883/3277713), though it really wouldn't matter too much for this case since we're usually only instantiating one instance anyway.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -0,0 +1,236 @@
...
+        self.providers = providers
+        self.endpoint = endpoint
+        self.delay = delay
+        self.retries = retries
+        self.headers = headers
+
+        # An airflow variable used to cap the amount of records to be ingested.
+        # This can be used for testing purposes to ensure a provider script
+        # processes some data but still returns quickly.
+        self.limit = Variable.get("ingestion_limit", None)
```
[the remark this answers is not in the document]
stacimc: Good idea!

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -0,0 +1,236 @@
...
+        """
+        A factory method to create the appropriate MediaStore for the given
+        media type.
+        """
+        if media_type == "image":
+            return ImageStore(provider)
+        if media_type == "audio":
+            return AudioStore(provider)
+
+        raise Exception(f"No MediaStore is configured for type: {media_type}")
```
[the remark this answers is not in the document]
stacimc: I actually toyed with adding something in `MediaStore` and backed out due to circular imports :) Let me take another look!

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -0,0 +1,236 @@
...
+        """
+        return True
+
+    def get_batch_data(self, response_json):
+        """
+        Takes an API response and returns the list of records. This method
+        can be overridden to accomodate different response formats.
+        """
+        if response_json:
+            return response_json.get("data")
```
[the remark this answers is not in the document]
stacimc: It is a bit arbitrary -- there are a couple other providers with this response format, but it's far from common. I think you're right that this would be better to simply leave as abstract.

openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
stacimc: I'm glad you brought this up, because I've actually gone back and forth on this a lot 😓  The fields were originally implemented as class level attributes, but I wasn't completely sold because I wanted to enforce that `providers` and `endpoint`, at least, be defined in the subclasses, and also because of the overriding of `batch_limit`. I did play around with abstract properties, but ultimately went with this structure because it's also consistent with [the way we've implemented the MediaStores](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/storage/audio.py#L37).

You also suggested adding a `__main__` block to [run the data ingester](https://github.com/WordPress/openverse-catalog/pull/555#discussion_r904174641). I wonder if an option might be to define constants for the Cleveland (or whatever provider) defaults within its file, and add a `__main__` block that initializes the Ingester using those defaults and runs ingestion. We could use that from the `provider_dag_factory` and avoid all the new code paths to deal with the difference between the class/function approach, and then the individual providers won't need to `super().__init__()` just to set new defaults 🤔

openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
stacimc: Hm do you mean in general, the parameter that gets passed through `dated` DAGs like Metropolitan? For dated DAGs, you _can_ override the `__init__` to add the parameter. Good point, though -- I think it's worth adding `date` as an optional param, and providers that don't care about it can just ignore it 👍

openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
stacimc: I guess the downside would be a slight disparity in the timing reported for the `pull_data` task, which would include the time to initialize the Ingester class along with stores/DelayedRequester in addition to ingestion. The logic to[ push the Store output paths to Xcoms](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/provider_dag_factory.py#L115) also still needs to be different for the old type of script and new ones that subclass the DataIngester, and that'll be harder to key off of 🤔

AetherUnbound: Oh, those are good points! I don't think that initialization time will contribute _too much_ to the overall time, but it's a good thing to be thinking about 🤔 

For your other points, I guess I'm not sure why having those as class attributes would affect the aspects you mention 😮 I like the idea of having `providers` and `endpoint` be abstract properties! `batch_limit` could still be set at the class level - I think even if we change that Variable while a run is happening, the next run will respect whatever the new limit is since Airflow re-imports everything on each DAG run step. And I was thinking the main block could look something like:

```python
if __name__ == "__main__":
    ingester = ClevelandDataIngester()
    ingester.ingest_records(DEFAULT_ARGS)
```

Or something similar. I might be missing initialization steps in that though 😅 and FWIW I don't think we would need to hook the `__main__` section up to publish XComs or interact with any of the rest of the workflow steps - that's strictly used for testing the ingester locally without having to run the DAG!

stacimc: For sure, the extra steps I suggested in `main` and using `main` instead of the class would only be useful in the scenario where we were still using instance variables -- I was thinking we could get a cleaner transition between the old way of invoking provider scripts and the new way, but realized we can't quite pull it off because of the XCom issue. It's definitely not useful outside of that!

openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
obulat: I was trying to rewrite one of the dated provider scripts using the ingester class to understand if there are problems that I'm not seeing simply from the code review, and this is where I stumbled, @stacimc :)

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -0,0 +1,236 @@
...
+
+        Return:
+        batch:           list of records in the batch
+        should_continue: boolean indicating whether ingestion should continue after
+                         this batch has been processed
+        """
+        batch = None
+        should_continue = True
+
+        try:
```
obulat, line 152: Some provider scripts also set `headers`:
https://github.com/WordPress/openverse-catalog/blob/dc8a68ba3b7784bc460b5e4758acfd14c42ddc15/openverse_catalog/dags/providers/provider_api_scripts/science_museum.py#L98-L99

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -0,0 +1,236 @@
...
+    """
+    An abstract base class that initializes media stores and ingests records
+    from a given provider.
+
+    Class variables of note:
+    providers:   a dictionary whose keys are the supported `media_types`, and values are
+                 the `provider` string in the `media` table of the DB for that type.
+                 Used to initialize the media stores.
+    endpoint:    the URL with which to request records from the API
+    delay:       integer giving the minimimum number of seconds to wait between
```
obulat, line 24:
```suggestion
    delay:       integer giving the minimum number of seconds to wait between
```

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
obulat: Sorry, I found the answer: You can use the class attribute on a child DataIngester class :)

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -0,0 +1,236 @@
...
+    def get_media_type(self, record):
+        """
+        For a given record, return the media type it represents (eg "image", "audio",
+        etc.) If a provider only supports a single media type, this may be hard-coded
+        to return that type.
+        """
+        pass
+
+    @abstractmethod
+    def get_record_data(self, record):
```
obulat, line 228: One more consideration:
Science museum has 'Objects' that can have several images described inside the `multimedia` dictionary. What is the best way of handling such nested items?

obulat: Maybe 
`record_data = self.get_record_data(record)`
could return a list of items?

openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
```diff
@@ -1,160 +1,115 @@
...
+    def _get_image_type(image_data):
+        # Returns the image url and key for the image in `image_data` dict.
+        for key in ["web", "print", "full"]:
+            if keyed_image := image_data.get(key):
+                return keyed_image
         return None
-    image = _get_image_type(data.get("images", {}))
-    if image is None or image.get("url") is None:
+
+    def _get_int_value(self, data: Dict, key: str) -> int | None:
```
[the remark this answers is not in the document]
stacimc: Oops, missed one! Thanks

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
stacimc: Oh, that's a good point -- SMK also does something similar, with `alternative_images`. 👀

stacimc: Hm, I think the vast majority of providers will be returning information for a single record. I pushed a change in [6112969](https://github.com/WordPress/openverse-catalog/pull/555/commits/61129690bee1a72a8587c226516af925d0c8274f) that allows `get_record_data` to return a single dict _or_ a list, but I'm not 100% sold. How do you feel about that change, or do you think it would be better to just require all providers to return a list here?

obulat: This is exactly what I wanted to suggest! For the same reasons as you state: it doesn't make sense to make it a list only because a couple out of all providers require it, it is better to handle it within the `ingester` class.
In fact, I added similar changes locally, and my refactored Science museum script works, and it is so much cleaner!!! It works well except for one small piece that I'll describe in a different comment.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -0,0 +1,250 @@
...
+
+        logger.info(f"Begin ingestion for {self.__class__.__name__}")
+
+        while should_continue:
+            query_params = self.get_next_query_params(query_params, **kwargs)
+
+            batch, should_continue = self.get_batch(query_params)
+
+            if batch and len(batch) > 0:
+                record_count = self.process_batch(batch)
```
obulat, line 109:
```suggestion
                record_count += self.process_batch(batch)
```

obulat: I'm not sure if this is the best coding style for this, but currently we are resetting the record_count with each batch.

tests/dags/providers/provider_api_scripts/test_cleveland_museum.py
```diff
@@ -35,197 +35,177 @@ def validate_url_string():
...
-
-
-def test_handle_response():
-    response_json = _get_resource_json("handle_response_data.json")
-    data = response_json["data"]
-    actual_total_images = clm._handle_response(data)
-    expected_total_images = 100
-
-    assert actual_total_images == expected_total_images
+class TestClevelandDataIngester(unittest.TestCase):
```
obulat, line 38: I am not entirely sure, but I think we're mostly using pytest in the catalog. And I keep hearing in podcasts how great it is compared to other testing frameworks. Is unittest better for this case?

tests/dags/providers/provider_api_scripts/test_cleveland_museum.py
```diff
@@ -35,197 +35,177 @@ def validate_url_string():
...
+
+        assert actual_image == expected_image
+
+    def test_get_image_type_full(self):
+        image_data = self._get_resource_json("image_type_full.json")
+        actual_image = self.clm._get_image_type(image_data)
+
+        expected_image = {
+            "url": "https://openaccess-cdn.clevelandart.org/1335.1917/1335.1917_full.tif",
+            "width": "6280",
```
obulat, line 94: I think `width` and `height` should be ints. Or does the image store convert them anyways?

stacimc: They get [cast to int](https://github.com/WordPress/openverse-catalog/pull/555/files/61129690bee1a72a8587c226516af925d0c8274f#diff-5fa4a32973ff2b61bd0ce8750c964c488aa0c4613da88f6f76f807dab3099214R65) in `get_record_data` when we pass them to the ImageStore, yep! This is just testing `get_image_type`, which just selects which image type to use when parsing record data (preferring web, print, full in that order). What's being returned at this point is just json.

tests/dags/providers/provider_api_scripts/test_cleveland_museum.py
stacimc: Not really, but we're not doing much with it either way -- just formatting, basically 😆

stacimc:
> > Do you think we should merge this PR and then adjust it if necessary when we are converting other scripts, or should we try to find the edge cases among the scripts and see what we might need to update before merging this PR?
> 
> I personally think the former. That might be the best way to get some 'hands on' experience about what changes would benefit this script, instead of trying to guess by looking at edge cases.

I agree with this! I think we'll naturally encounter additional edge cases or even just places where this can be improved as we work through additional provider scripts.

Thanks for opening #576 @obulat, that's an above-and-beyond way to code review 😄 ❤️

krysal (approved):
This is such a great improvement to providers' scripts, I love the modularity! and it will reduce a lot of duplicated code. Looks like everything has been said and this is ready to go! 👏

Also makes sense to me to merge this and look at the edge cases while converting other providers one by one. Otherwise, this PR will be huge, we can already take advantage of it.

obulat (approved):
I love this! :)

Issue #450, opened 2022-03-29
Investigate & mitigate Finnish Museums duplicates


AetherUnbound:
## Description

The Finnish Museums provider appears to be providing us with two kinds of "duplicates":
1. Straight duplicates with no differences in the data provided
1. Duplicates on `foreign_identifier` which have different `url` values (namely with different `index` parameters, e.g. `index=0`, `index=1`, etc).

The latter appears to be a set of "alt images" for a specific foreign identifier. Here's an example _(note the `index` values)_:
1. https://api.finna.fi/Cover/Show?id=museovirasto.AC3BCA665DB5AB5D26B9F0E2A0CC6E49&index=0&size=large&source=Solr
1. https://api.finna.fi/Cover/Show?id=museovirasto.AC3BCA665DB5AB5D26B9F0E2A0CC6E49&index=1&size=large&source=Solr
1. https://api.finna.fi/Cover/Show?id=museovirasto.AC3BCA665DB5AB5D26B9F0E2A0CC6E49&index=2&size=large&source=Solr

In these cases, the `index=0` value seems to be the highest value entry. Unfortunately, the logic that we have for cleaning duplicates will **always** select the last/highest index:
https://github.com/WordPress/openverse-catalog/blob/962f1c5d9bb17bd6ea1ed5cc0674bab76d638f0b/openverse_catalog/dags/common/loader/sql.py#L173-L178

We may want to simply chose `index=0` images only, since those seem to be the highest value.

As for the first duplicate case, we're not sure where these are coming from. In the last run, we received ~500k records, but the duplicate cleaning step removed ~400k entries. After some investigation, we found that most of the duplicates were coming from a few specific foreign identifiers. Here is a mapping of the number of duplicates for a given foreign identifier to the number of times that many duplicates occurred:
```python
[ins] In [27]: dict(zip(unique, counts))
Out[27]: 
{1: 99066,
 2: 251,
 3: 18,
 4: 10,
 5: 1,
 7: 1,
 14: 1,
 18: 1,
 40: 1,
 44: 1,
 50: 1,
 3993: 36,
 3994: 64}
 ```

 Said another way, we had **64 different foreign identifiers that each had 3994 total duplicates for that identifier**. This _might_ be due to the way API ranges are calculated, or it may be a quirk of the Finnish Museums API. Further investigation is needed.

## Reproduction
Using the most recently produced Finnish TSV, here's the analysis steps I took:
```python
import pandas as pd
import numpy as np
[...]

PR #579, opened 2022-06-30
Update Finnish Museums to use base class

stacimc:
## Description

Quick PR to update Finnish Museums to use the new `ProviderDataIngester` base class.

## Testing Instructions

`just test` and confirm tests pass

Run the Finnish Museums provider script and watch the `pull_data` task. Make sure it at least pulls records for a few `buildings` before marking as success; alternatively you can set an `ingestion_limit` admin variable to cap the number of records pulled.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
```diff
@@ -1,167 +1,128 @@
...
+    batch_limit = 100
+    delay = 5
+    format_type = "0/Image/"
+
+    def ingest_records(self, **kwargs):
+        for building in BUILDINGS:
+            logger.info(f"Obtaining images of building {building}")
+            super().ingest_records(building=building)
+
+    def get_next_query_params(self, old_query_params, **kwargs):
```
AetherUnbound, line 34: Gah, sorry I didn't recognize this in the original PR - do you think this should be `prev_query_params` instead of `old_query_params`? "Old" makes me think "legacy", which isn't quite the case here. Just a thought, not critical!

tests/dags/providers/provider_api_scripts/test_finnish_museums.py
```diff
@@ -54,70 +39,72 @@ def test_build_query_param_default():
...
-
-    mock_add_item.assert_called_once_with(
-        license_info=(
-            LicenseInfo(
-                "by",
-                "4.0",
-                "https://creativecommons.org/licenses/by/4.0/",
-                "http://creativecommons.org/licenses/by/4.0/",
-            )
+    # with patch.object(fm.image_store, "add_item", return_value=100) as mock_add_item:
```
AetherUnbound, line 80: Should this be removed?

AetherUnbound (approved):
This ran just fine locally, and everything looks good! 🎉 

One thing to note (which you may have already identified): the `for building in BUILDINGS` loop which calls `super().ingest_records` means that the `ingestion_limit` is checked _for every building_ rather than globally. This means that every building is still run, but only for 200 records. As a result, I had about 800 records loaded. Not a huge deal, but was unexpected on my end! It does show that `ingestion_limit` works though, which is excellent 😁

openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
stacimc: Thank you for this -- now you mention it, I actually agree really strongly 😆 `prev` feels much better, I'll update this.

openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
```diff
@@ -1,167 +1,128 @@
...
-            source=source,
-            raw_tags=raw_tags,
+class FinnishMuseumsDataIngester(ProviderDataIngester):
+    providers = {"image": PROVIDER}
+    endpoint = "https://api.finna.fi/api/v1/search"
+    batch_limit = 100
+    delay = 5
+    format_type = "0/Image/"
+
+    def ingest_records(self, **kwargs):
```
obulat, line 29: It's not really that important, but when we override `ingest_records`, the logging can be a bit misleading. After logging that we start 'Obtaining images for buildings', the parent Ingester class logs that it starts ingestion:
```
[2022-07-27, 21:35:17 +03] {provider_data_ingester.py:112} INFO - Batch complete.
[2022-07-27, 21:35:17 +03] {media.py:213} INFO - Writing 5 lines from buffer to disk.
[2022-07-27, 21:35:17 +03] {finnish_museums.py:31} INFO - Obtaining images of building 0/Museovirasto/
[2022-07-27, 21:35:17 +03] {provider_data_ingester.py:101} INFO - Begin ingestion for FinnishMuseumsDataIngester
```

openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
```diff
@@ -1,167 +1,128 @@
...
+    def get_batch_data(self, response_json):
+        if (
+            response_json is None
+            or str(response_json.get("status")).lower() != "ok"
+            or response_json.get("records") is None
+            or len(response_json.get("records")) == 0
+        ):
+            return None
+
+        return response_json.get("records")
```
obulat, line 56:
```suggestion
        return response_json["records"]
```
Tiny nit: we have already tested that `response_json` has a key of `records`, and the value's length is not 0.

openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
```diff
@@ -1,167 +1,128 @@
...
+        foreign_landing_url = self._get_landing(data)
+        raw_tags = self._get_raw_tags(data)
+
+        image_list = data.get("images")
+        for img in image_list:
+            image_url = self._get_image_url(img)
+            records.append(
+                {
+                    "license_info": get_license_info(license_url),
+                    "foreign_identifier": foreign_identifier,
```
obulat, line 80: Foreign identifier is supposed to be unique. Here, we can be collecting several images for a single record. They will all have the same foreign_identifier, resulting in many duplicates. We need to either add the image index to the `foreign_identifier` here, or only use the first image from the `image_list` here that would be the single image for the record in question.

I tried looking into the records that have several images, and I think I'm in favor of using only the first image for each record.
Here's an example of record with 2 images that are quite different and we _might_ want to add both: https://www.finna.fi/Record/museovirasto.AA6BD6DA6AB5F4C526C79DC74C7A58D7?imgid=1

But there are many more records that are like this: https://www.finna.fi/Record/museovirasto.D24CDFCAA0324B50C830BD43133CA3F1
The first image is the actual image, and all others are pictures of the back of the picture frame.

openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
```diff
@@ -1,167 +1,128 @@
...
+        # (eg `licenses/by/4.0/deed.fi`), but the license validation logic expects
+        # links to the license page (eg `license/by/4.0`).
+        return license_url.removesuffix("deed.fi")
+
+    @staticmethod
+    def _get_raw_tags(obj):
+        tag_lists = obj.get("subjects")
+        if tag_lists is None:
+            return None
+        return list(chain(*tag_lists))
```
obulat, line 106: TIL!

openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
```diff
@@ -1,167 +1,128 @@
...
+
+    @staticmethod
```
obulat, line 108: Do we really need to have these two methods separate? I remember reading that Python's function calls can be expensive, and these two functions could be replaced with a couple of lines?

obulat (changes requested):
I left a note inline about the foreign id duplication, I think we should fix it before merging this PR.

AetherUnbound:
> I left a note inline about the foreign id duplication, I think we should fix it before merging this PR. And I removed sara from the reviewers list because I reviewed the PR. I hope it's OK.

We plan on addressing this in https://github.com/WordPress/openverse-catalog/issues/450 I believe! This is just a straight refactor of existing code, and I think we can work to solve that issue in another PR.

obulat (approved):
This is a great improvement!
Addressing the `foreign_identifier` problem separately (due to the fact that it requires us to make a decision on whether to keep all images) makes sense.

openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
stacimc: Yep, we ran into this awhile ago, including the funny back-of-frame issue 😭 -- there's a separate ticket for it here https://github.com/WordPress/openverse-catalog/issues/450 but I think your assessment is exactly right. Since this involves making a decision about how to handle the duplicates I'd still prefer to keep that discussion/work separate.

openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
stacimc: True!

openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
stacimc: It's definitely not _necessary_ to split them, just a slight readability concern to keep `get_record_data` as simple as possible. These are simple enough that I wouldn't be opposed to getting rid of them, though.

Function calls _are_ relatively expensive in Python, but normally I'd consider that a micro-optimization. In the case of Finnish Museums though we've been running up against execution timeouts for awhile now, so we may as well! Good looking out.

PR #576, opened 2022-06-29
Refactor Science museum to use ProviderDataIngester

obulat:
## Fixes

Fixes #582

## Description

This was PR created to better review the new Ingester class and to show how awesome it is :)
Now it's the actual PR to refactor the Science museum script to use ProviderDataIngester class. It also should fix the issue when the license information was incorrect and caused the script to stop.

While refactoring this script, I thought that we are probably extracting the creator here incorrectly:

Science museum uses the "Maker" of the object as the author for the image. Shouldn't we normally put the **author** of the **image**, not the **object** being photographed? 

For example, for this telescope replica(https://collection.sciencemuseumgroup.org.uk/objects/co56202/telescope-by-galileo-replica-telescope-galilean-telescope-refracting-replica) should we put _Galileo_ or  _The Board of Trustees of the Science Museum_  (based on the note below the image: Science Museum Group Collection © The Board of Trustees of the Science Museum )? 

## Testing Instructions

`just test` and confirm tests pass.

Run the Science Museum provider script and watch the pull_data task. Make sure it at least pulls records for a few buildings before marking as success; alternatively you can set an ingestion_limit admin variable to cap the number of records pulled.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
[...]

PR #600, opened 2022-07-03
Refactor Museum Victoria to use ProviderDataIngester

### LINKED ISSUE #589: Refactor Museum Victoria to use ProviderDataIngester
## Suggested Improvement

Refactor the Museum Victoria provider script to use the new ProviderDataIngester base class

## Benefit

More details in #229

## Implementation

obulat:
## Fixes

Fixes #589 by @stacimc

## Description

This was originally a PR to add filetype and filesize to the script, but then I incorporated those changes into the ProviderDataIngester refactor.

## Testing Instructions

Check the CI logs. Run Museum Victoria DAG.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


openverse_catalog/dags/common/requester.py
```diff
@@ -79,7 +79,9 @@ def get_response_json(self, endpoint, retries=0, query_params=None, **kwargs):
                 logger.warning(f"Could not get response_json.\n{e}")
                 response_json = None
 
-        if response_json is None or response_json.get("error") is not None:
+        if response_json is None or (
```
obulat, line 82: Victoria returns a list instead of an object in `json()` call, so we cannot use `response_json.get('error')` with it. `response_json` is a list.

openverse_catalog/dags/providers/provider_api_scripts/museum_victoria.py
```diff
@@ -1,203 +1,175 @@
...
+    ]
+
+    def ingest_records(self, **kwargs):
+        for license_ in VictoriaDataIngester.LICENSE_LIST:
+            super().ingest_records(license_=license_)
+
+    def get_batch_data(self, response_json):
+        if response_json:
+            return response_json
+        return None
```
AetherUnbound, line 49: Nit: this could be 
```suggestion
        return response_json or None
```

openverse_catalog/dags/providers/provider_api_scripts/museum_victoria.py
```diff
@@ -1,203 +1,175 @@
...
+
+    def get_next_query_params(self, old_query_params: Optional[Dict], **kwargs) -> Dict:
```
AetherUnbound, line 51: Note that this will soon be `prev_query_params` so it might be good to preemptively change that here: https://github.com/WordPress/openverse-catalog/pull/579#discussion_r919333596

openverse_catalog/dags/providers/provider_api_scripts/museum_victoria.py
```diff
@@ -1,203 +1,175 @@
...
-    if type(keywords) == list:
-        metadata["keywords"] = ",".join(keywords)
-
-    classifications = obj.get("classifications")
-    if type(classifications) == list:
-        metadata["classifications"] = ",".join(classifications)
-
-    return metadata
+            license_info = VictoriaDataIngester._get_license_info(media)
+            if image_url is None or image_id is None or license_info is None:
```
AetherUnbound, line 103: I think this is fine but in case this gets any larger, we could do:
```suggestion
            if any(val is None for val in [image_url, image_id, license_info]):
```

openverse_catalog/dags/providers/provider_api_scripts/museum_victoria.py
```diff
@@ -1,203 +1,175 @@
...
+    def get_record_data(self, data: Dict):
+        object_id = data.get("id")
+        if object_id in self.RECORDS_IDS:
+            return None
+        self.RECORDS_IDS.add(object_id)
+        foreign_landing_url = f"{self.LANDING_PAGE}{object_id}"
+
+        if (media_data := data.get("media")) is None:
+            return None
+        images = VictoriaDataIngester._get_images(media_data)
```
AetherUnbound, line 74: This doesn't matter too much but all the static reference calls on `VictoriaDataIngester` can be replaced with `self` (it at least makes them a bit shorter for formatting haha)

openverse_catalog/dags/providers/provider_api_scripts/museum_victoria.py
```diff
@@ -1,203 +1,175 @@
...
+        image_url = media_data.get("uri")
+        if image_url is not None:
+            height = media_data.get("height")
+            width = media_data.get("width")
+            filesize = media_data.get("size")
+        return image_url, height, width, filesize
+
+    @staticmethod
+    def _get_creator(media) -> str | None:
+        creators = media.get("creators", None)
```
AetherUnbound, line 140: This can be called without a default, since the default is already `None`
```suggestion
        creators = media.get("creators")
```

openverse_catalog/dags/providers/provider_api_scripts/museum_victoria.py
```diff
@@ -1,203 +1,175 @@
...
+        if isinstance(creators, list):
+            creators = ",".join(media.get("creators"))
+        return creators
+
+    @staticmethod
+    def _get_metadata(obj):
+        def join_string_list(object_key, obj):
+            data = obj.get(object_key)
+            return ",".join(data) if isinstance(data, list) else None
```
AetherUnbound, line 149: Could this become a separate static method as well? Or maybe even a utility function somewhere, if we're likely to re-use it! I'm concerned that having `obj` at both the inner and the outer scope can be a bit confusing, since they ultimately reference the same object but the inner function gets handed it anyway.

tests/dags/providers/provider_api_scripts/test_museum_victoria.py
```diff
@@ -1,182 +1,200 @@
...
+        image_data
+    )
 
     assert actual_image_url == (
         "https://collections.museumsvictoria.com.au/content/media/45/"
         "329745-small.jpg"
     )
     assert actual_height == 500
     assert actual_width == 540
+    assert actual_filesize == 20109
```
AetherUnbound, line 169: Should we be checking `filetype` in any of these tests as well?

AetherUnbound (commented):
This is so much more organized! 🤩 I have just a few questions/comments

AetherUnbound:
DAG works great locally!

openverse_catalog/dags/providers/provider_api_scripts/museum_victoria.py
obulat: I made the same suggestion on one of stacimc's PR's 😆 Thank you!

openverse_catalog/dags/providers/provider_api_scripts/museum_victoria.py
obulat: Much better (shorter) with `self` :)

tests/dags/providers/provider_api_scripts/test_museum_victoria.py
obulat: I've added a test for filetype that mocks image_store's `save_item` method. It passes if I run that test alone, but fails when I run all the tests together (and in CI). I need help :)

openverse_catalog/dags/providers/provider_api_scripts/museum_victoria.py
```diff
@@ -137,23 +137,26 @@ def _get_image_data(
...
+    @staticmethod
+    def _get_metadata(obj):
         meta_data = {
             "datemodified": obj.get("dateModified"),
             "category": obj.get("category"),
             "description": obj.get("physicalDescription"),
-            "keywords": join_string_list("keywords", obj),
-            "classifications": join_string_list("classifications", obj),
+            "keywords": VictoriaDataIngester.join_string_list("keywords", obj),
+            "classifications": VictoriaDataIngester.join_string_list(
```
AetherUnbound, line 157: Note for `self` here as well 😄

tests/dags/providers/provider_api_scripts/test_museum_victoria.py
```diff
@@ -88,6 +89,16 @@ def test_get_record_data():
...
 
 
+def test_filetype_gets_added_by_image_store():
+    media = _get_resource_json("record_data.json")
+    with patch.object(mv.media_stores["image"], "save_item") as mock_save:
+        mv.process_batch([media])
+
+    actual_image = mock_save.call_args[0][0]
+
+    assert "jpg" == actual_image.filetype
```
AetherUnbound, line 99: Gah, I'm so sorry - I keep forgetting that this is effectively covered by the media store class and probably doesn't need to be tested at the ingester level! I'll still try to figure out why this test is failing in certain cases, but I think barring any weird discoveries there we can probably drop this test.

openverse_catalog/dags/providers/provider_api_scripts/museum_victoria.py
```diff
@@ -1,203 +1,178 @@
...
+
+
+class VictoriaDataIngester(ProviderDataIngester):
+    providers = {"image": prov.VICTORIA_DEFAULT_PROVIDER}
+    endpoint = "https://collections.museumsvictoria.com.au/api/search"
+    headers = {"User-Agent": prov.UA_STRING, "Accept": "application/json"}
+    batch_limit = 100
+    delay = 5
+    # This set is used to prevent duplicate images of the same items
+    RECORDS_IDS = set()
```
AetherUnbound, line 31: Looking at this again, I think it may be ill-advised to have a mutable object attached to the class instance (same for #576) - do you think we could move this to an `__init__` override and set it once the instance is initialized instead?

AetherUnbound (commented):
Very weird about the failing tests! I'll look into that now

openverse_catalog/dags/providers/provider_api_scripts/museum_victoria.py
stacimc: There's so many little things that can be cleaned up while we're in here refactoring :) It's a good opportunity!

openverse_catalog/dags/providers/provider_api_scripts/museum_victoria.py
AetherUnbound: It turns out this is the _exact_ cause of the test failures! Multiple tests ingest that `record_data`, and so when more than one test is run the object already exists and isn't added 😅 Although even moving this into the `__init__` method wouldn't immediately fix this, we'd probably need to add a fixture called `mv` which returns a new `VictoriaDataIngester` instance for each test.

openverse_catalog/dags/common/requester.py
obulat: @stacimc , do you think this is a good solution here for cases where `response_json` is a list, not a dictionary?

openverse_catalog/dags/providers/provider_api_scripts/museum_victoria.py
obulat: This is a `static` method that doesn't have a `self` 😢

openverse_catalog/dags/providers/provider_api_scripts/museum_victoria.py
obulat: Thank you so much for figuring this out, @AetherUnbound !!!
I've moved `RECORD_IDS` to `__init__`, and in the tests file added a fixture that runs after each test and replaces `RECORD_IDS` with a blank set.

openverse_catalog/dags/providers/provider_api_scripts/museum_victoria.py
AetherUnbound: D'oh! Sorry about that 😭

AetherUnbound (approved):
🎉 Woohoo! Looks great!

openverse_catalog/dags/common/requester.py
stacimc: I guess the alternative would be to override `get_response_json` on `VictoriaDataIngester` to use `delayed_requester.get` instead of `.get_response_json`, like the original provider script -- in which case we miss out on error handling and retries. I don't love that the API is returning json without a top level object, but I think this is the best solution on our end. At any rate we're actually doing more error handling now than we were before! :)

stacimc (approved):
🎉 LGTM and tests well locally!

openverse_catalog/dags/providers/provider_api_scripts/museum_victoria.py
```diff
@@ -1,203 +1,183 @@
...
-def _get_media_info(media_data):
-    image_data = []
-    for media in media_data:
-        media_type = media.get("type")
-        if media_type == "image":
-            image_id = media.get("id")
-            image_url, height, width = _get_image_data(media)
-            license_url = _get_license_url(media)
-            if image_url is None or image_id is None or license_url is None:
+class ImageDetails(TypedDict, total=False):
```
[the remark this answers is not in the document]
obulat: True! :)

PR #601, opened 2022-07-03
Refactor StockSnap to use ProviderDataIngester

### LINKED ISSUE #595: Refactor Stocksnap to use ProviderDataIngester
## Suggested Improvement

Refactor the Stocksnap provider script to use the new ProviderDataIngester base class

## Benefit

More details in #229

## Implementation

rwidom:
## Fixes
Fixes #595 by @stacimc 

## Description
Moves StockSnap provider API script to use the new ProviderDataIngester class. (More detail in #299 and https://github.com/WordPress/openverse-catalog/pull/555 ).

Questions / notes for reviewers:
- I couldn't test with the actual API, because [StockSnap doesn't have a fully public API](https://stocksnap.io/faq). So my testing is based on the test resources and existing test cases in the repo.
- `ProviderDataIngester` intentionally does not handle validation, but I thought `_get_int_value` ([from the Cleveland script](https://github.com/WordPress/openverse-catalog/blob/1c9fd4cf46f7c90f8e9db6eaacb31d6ad031fcf5/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py#L80)) was so nice... I know all validation comes with a performance/resource cost, though, and all of the test data had easy to use ints. OK to use the function in this way?
- StockSnap doesn't use query parameters, but [reflects the incremental page in the endpoint url](https://github.com/WordPress/openverse-catalog/blob/1c9fd4cf46f7c90f8e9db6eaacb31d6ad031fcf5/openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py#L63). It's possible that parameters would work too, but I didn't have a quick way to find out, so I used `endpoint` and `get_should_continue` to increment the endpoint the way the existing code does.
 
## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py
@@ -15,11 +15,6 @@ import logging
 
 from common.licenses import get_license_info
 from common.loader import provider_details as prov
-
-# There are a couple of things that seem really broadly applicable from the Cleveland
-# sample script: the method _get_int_value and the constant CC0_LICENSE.
-# I'm not using the latter, because it fails some of the existing StockSnap tests,
-# because of a difference in trailing / in the license url.
 from providers.provider_api_scripts.cleveland_museum import ClevelandDataIngester
 from providers.provider_api_scripts.provider_data_ingester import ProviderDataIngester
 
--- a/openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py
@@ -3,20 +3,25 @@ Content Provider:       StockSnap
 
 ETL Process:            Use the API to identify all CC-licensed images.
 
-Output:                 TSV file containing the image, the respective
-                        meta-data.
+Output:                 TSV file containing the image, the respective meta-data.
 
-Notes:                  https://stocksnap.io/api/
-                        No rate limit specified. No authentication required.
+Notes:                  https://stocksnap.io/faq
                         All images are licensed under CC0.
+                        Seems like their API has been suspended, but this assumes
+                        that there is a special agreement with Openverse.
 """
 import json
 import logging
 
 from common.licenses import get_license_info
 from common.loader import provider_details as prov
-from common.requester import DelayedRequester
-from common.storage.image import ImageStore
+
+# There are a couple of things that seem really broadly applicable from the Cleveland
+# sample script: the method _get_int_value and the constant CC0_LICENSE.
+# I'm not using the latter, because it fails some of the existing StockSnap tests,
+# because of a difference in trailing / in the license url.
+from providers.provider_api_scripts.cleveland_museum import ClevelandDataIngester
+from providers.provider_api_scripts.provider_data_ingester import ProviderDataIngester
 
 
 logging.basicConfig(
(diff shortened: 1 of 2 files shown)


openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py
```diff
@@ -28,175 +28,165 @@
...
+            "width": width,
+            "license_info": lic,
+            "meta_data": metadata,
+            "raw_tags": tags,
+            "category": prov.DEFAULT_IMAGE_CATEGORY[PROVIDER],
+        }
+
+    @staticmethod
+    def _get_image_info(item):
+        width = ClevelandDataIngester._get_int_value(item, "img_width")
```
stacimc, line 124: `_get_int_value` is useful, but I'd prefer to pull it out of `ClevelandDataIngester` and into a shared `util` module if it's going to be reused elsewhere. Another alternative would be to add this cleaning step to the `MediaStore` (maybe [here](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/storage/media.py#L98)?) so that all providers get it automatically.

openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py
```diff
@@ -28,175 +28,165 @@
...
-def _get_batch_json(
-    endpoint=ENDPOINT, headers=None, retries=RETRIES, query_params=None
-):
-    if headers is None:
-        headers = HEADERS.copy()
-    response_json = delayed_requester.get_response_json(
-        endpoint, retries, query_params, headers=headers
-    )
-    if response_json is None:
+    def get_next_query_params(self):
```
stacimc, line 48: This will throw an error when `ingest_records` attempts to invoke it with `old_query_params`
```suggestion
    def get_next_query_params(self, old_query_params, **kwargs):
```

openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py
```diff
@@ -28,175 +28,165 @@
...
-        return None
-    title = _get_title(media_data)
-    if title is None:
-        logger.info("Found no image title.")
-        logger.info(f"{json.dumps(media_data, indent=2)}")
+
+    def get_media_type(self):
+        return "image"
+
+    def endpoint(self):
```
stacimc, line 56: We should annotate this as a property so that we can continue to access it as `self.endpoint`.
```suggestion
    @property
    def endpoint(self):
```

openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py
```diff
@@ -28,175 +28,165 @@
...
+        return f"{ENDPOINT_BASE}/{self._page_counter}"
+
+    def get_should_continue(self, response_json):
+        if isinstance(response_json, list) and len(response_json) > 0:
+            self._page_counter += 1
+            return True
+        else:
+            return False
```
stacimc, line 64: This works, but I'm not sure `get_should_continue` is the best place to increase the counter 🤔 The intention of this method is to allow additional logic to short-circuit ingestion (so for example, you might place a check for some continue token here). The check here just duplicates the logic in `get_batch_data`, and the counter increment is a side effect that's not really logically related.

We could also increment the counter in `get_next_query_params`. Since it's called before making the first request, you'd have to make sure that it's initialized correctly for the first request. What do you think?

openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py
```diff
@@ -28,175 +28,165 @@
 DELAY = 1  # in seconds
 RETRIES = 3
```
stacimc, line 29: We can remove `DELAY` and `RETRIES` here

stacimc (changes requested):
🎉  I am so excited to see this @rwidom, thanks for picking this up!

I was able to run this locally with real data, but first you'll need to update the [StockSnap provider workflow definition](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_workflows.py#L144) in `provider_workflows.py` to configure the ingester class:

```
ProviderWorkflow(
    provider_script="stocksnap",
    ingester_class=StockSnapDataIngester,
)
```

I left a couple comments for some errors being thrown at the moment, but then you should be able to run this locally and actually pull data. I don't think an API key is necessary.

> StockSnap doesn't use query parameters, but [reflects the incremental page in the endpoint url](https://github.com/WordPress/openverse-catalog/blob/1c9fd4cf46f7c90f8e9db6eaacb31d6ad031fcf5/openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py#L63)

All of these providers seem to have some little quirk! 😄 I like your idea of updating the endpoint with a page counter.

openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py
rwidom: :face_palm: That makes total sense. I'll do it.

rwidom: That makes total sense. I'll do it.

openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py
stacimc: Ah, you're right, since those fields are media type specific it shouldn't go in `MediaStore`! But maybe we could extend `clean_media_metadata` in the `ImageStore`?

If we go that route, we'd have the advantage of simplifying even more provider script code and automatically getting the validation for all image providers (even ones that haven't been refactored yet). But for the same reason it's probably best to pull that into a separate PR so we don't increase the scope of this one. If you think that's a good idea, I think we can pull the logic out of this script for now (since the code currently in production does not do this validation) and wait to add it for all providers.

rwidom: Yeah, that makes total sense to me. I'll do that.

rwidom: By "that" I mean just taking out the validation for now, though extending `clean_media_metadata` in the `ImageStore` makes total sense, and I wish I had thought of it. :)

rwidom:
I don't seem to be able to run this with real data locally (at least, when I do `just down -v`, `just recreate`, and then go to my local airflow site and click run, it runs twice "successfully" but skips writing to or pulling from S3 both times, and I can't find any tmp directory with tsv files on my local machine after a very quick look), but that might be about separate issues with my local set-up, which I'm going to delve into more with @sarayourfriend later this week. 

I think everything else is resolved?

stacimc (changes requested):
Just a couple quick comments I forgot as I was re-testing this -- thanks for your perseverance despite not being able to run this locally! I do think your problems must be related to your environment setup, since the API does not require authentication 😞 

After those comments though, I was able to run StockSnap with and without an `ingestion_limit` and pull/load data perfectly 🎉

openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py
```diff
@@ -25,178 +24,160 @@
...
+    def endpoint(self):
+        return f"{ENDPOINT_BASE}/{self._page_counter}"
+
+    def get_batch_data(self, response_json):
+        """
+        Take an API response and return the list of records.
+        """
+        # Based on the sample test file, using json.load on the file will return a list.
+        if isinstance(response_json, list) and len(response_json) > 0:
+            return response_json
```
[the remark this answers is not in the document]
rwidom: Wow, thank you so much! Yes, will definitely do that. I swear I tried that url and got 404, but I see it is working today, so I will definitely adjust accordingly. Thanks so much!

openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py
```diff
@@ -3,20 +3,19 @@
...
-Output:                 TSV file containing the image, the respective
-                        meta-data.
+Output:                 TSV file containing the image, the respective meta-data.
 
-Notes:                  https://stocksnap.io/api/
-                        No rate limit specified. No authentication required.
+Notes:                  https://stocksnap.io/faq
                         All images are licensed under CC0.
+                        Seems like their API has been suspended, but this assumes
+                        that there is a special agreement with Openverse.
```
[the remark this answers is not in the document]
rwidom: OK, I'm happy to put back the no rate limit, no authentication, but when I go to https://stocksnap.io/api/ I still get 404 today. I think that may be why I was confused about the end point stuff. So, I mean, I could put the full endpoint there, or the link to the faq that I provided or some other in between language?

zackkrida (approved):
I can't provide as many in-depth suggestions for the actual code as other folks on the team, but I'm able to run this locally without issue. It's great to see the new `ProviderDataIngester` in action!

openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py
```diff
@@ -25,178 +25,162 @@
...
+        if title is None:
+            logger.info("Found no image title.")
+            logger.info(f"{json.dumps(data, indent=2)}")
+            return None
+
+        creator, creator_url = self._get_creator_data(data)
+        metadata = self._get_metadata(data)
+        tags = self._get_tags(data)
+        filesize = self._get_filesize(image_url)
+        lic = get_license_info("https://creativecommons.org/publicdomain/zero/1.0/")
```
krysal, line 97: This will be processed for each image so declaring it at the beginning as a class attribute would be slightly more efficient :)

```suggestion
        lic = self.license
```

krysal (commented):
This is great @rwidom! I hope you have been able to run it, it's working well. I have left just a few comments to improve the data we are getting from this provider, but really this is almost nearly complete. Excellent work!

openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py
```diff
@@ -25,178 +25,162 @@
...
+        """
+        Parse out the necessary information (license info, urls, etc) from the record
+        data (single image dict-like thing) into a dictionary.
+        """
+        try:
+            foreign_id = data["img_id"]
+        except (TypeError, KeyError, AttributeError):
+            return None
+
+        foreign_landing_url = f"https://{HOST}/photo/{foreign_id}"
```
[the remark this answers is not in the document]
stacimc: Cool idea. I wonder if they changed the link structure at some point -- nice find 💯

tests/dags/providers/provider_api_scripts/test_stocksnap.py
```diff
@@ -17,42 +19,131 @@
...
+
+
+#     def _get_image_info(item):
+#     def _get_title(item):
+#     def _get_metadata(item):
+# def ingest_records(self, **kwargs) -> None:
+# def get_batch(self, query_params: Dict) -> Tuple[Optional[List], bool]:
+# def get_should_continue(self, response_json):
+# def get_batch_data(self, response_json):
+# def process_batch(self, media_batch):
```
stacimc, line 34: Looks like this was accidentally left in :)

openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py
rwidom: Of course! Sorry, I think I was in inaturalist head where different images have different licenses. Will do!

tests/dags/providers/provider_api_scripts/test_stocksnap.py
rwidom: Ugh, yes, thank you!

openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py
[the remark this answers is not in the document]
stacimc: Yeah, it's definitely an unusual case! My thinking is that `https://stocksnap.io/api` is the correct base url, while _just_ the faq without context can actually be misleading. I like what you ended up doing here with linking to both and adding a comment about it being undocumented 👍

stacimc (approved):
💯 This looks great! Thanks so much for your perseverance with local testing issues :) Everything looks good, and running locally works great.

rwidom:
Thanks so much @zackkrida , @krysal  and @stacimc!

rwidom:
> 💯 This looks great! Thanks so much for your perseverance with local testing issues :) Everything looks good, and running locally works great.

Thanks so much! I didn't see it had been approved when I just edited the overview. Oh well. Either way-- Yay! Onward!

PR #608, opened 2022-07-14
Elasticsearch metrics DAG

### LINKED ISSUE #564: Elasticsearch metrics reporting DAG
## Description

It would be great if we could have a DAG that would regularly (possibly daily) generate some graphs of Elasticsearch metrics and report them to Slack. We've used our ES metrics as a proxy for determining when our API is being hit hard, and it would be helpful to have insight into those metrics in a form that doesn't require folks to log into AWS.

I think we should at least include **Network Data** and **CPU** in the metrics that are reported.

**Example**:
image

## Alternatives

We plan on having better metrics and alerting as part of our [monitoring project](https://github.com/WordPress/openverse/pull/193), but this will provide a solid stopgap until then.

## Additional context

We can use boto3's [`Cloudwatch::get_metric_widget_image`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudwatch.html#CloudWatch.Client.get_metric_widget_image) function to generate the images. We will then have to either upload them to Slack or upload them somewhere that's publicly accessible so Slack can reference them. Our [Slack module](https://github.com/WordPress/openverse-catalog/blob/c4ffd95ddbb085eae3512622e82394ac39c7fb14/openverse_catalog/dags/common/slack.py) does not currently have the capacity to upload images, so if we went that route we'd need to add that functionality. Alternatively @sarayourfriend has suggested we use [pre-signed URLs in a private bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ShareObjectPreSignedURL.html) - this will allow us to set expiration times to the links provided in Slack (up to 7 days).

I have some example code for how to potentially accomplish this here: https://github.com/WordPress/openverse-catalog/compare/main...feature/elasticsearch-metrics

## Implementation

AetherUnbound:
## Fixes

Fixes #564 by @AetherUnbound

## Description

This PR adds a DAG which captures Elasticsearch metrics for the 72 hours prior to the schedule interval and reports them to slack.

I also took this opportunity to use the [TaskFlow API](https://airflow.apache.org/docs/apache-airflow/stable/tutorial_taskflow_api.html) for Airflow - it made the actual DAG definition so much simpler! Super useful when all of your steps would otherwise be `PythonOperators`. 

### Additional notes
- This adds a few new variables/connections. With the way this is set up though, nothing (should) need to be modified in production besides adding the `ES_INSTANCE_IDS` Variable.
- This adds a new python package ([PIL](https://pillow.readthedocs.io/en/stable/)) in order to perform some image manipulation. This means we'll need to publish and deploy a new Docker image in order to actually run this DAG. We could do this as a specific commit, or we could cut and release v1.3.1.
- This adds a new bucket to S3, `openverse-airflow`. This bucket can be used for Airflow entities that don't explicitly relate to the catalog (since `openverse-catalog` almost exclusively has TSVs in it). I've added this bucket by hand for testing, but I recognize that it should probably have a follow-up in the infrastructure repo. @sarayourfriend do you have any insight on where it'd be best to put that?
- I tried to make the DAG agnostic to the number of metrics we're gathering, in case we want to add more in the immediate future.

### Screenshots
**DAG**
image

**Slack Message**
Screenshot_2022-07-14_15-48-42

## Testing Instructions

In order to test this DAG locally, you will need to set the following variables in your `.env` file:
```bash
AIRFLOW_VAR_ES_INSTANCE_IDS=i-00e708a227372b51b,i-0e609c909948a7a02,i-075774d799a0da4b3
AIRFLOW_CONN_AWS_METRICS=aws://<your-key>:<your-secret-url-encoded>@?region_name=us-east-1
AWS_METRICS_CONN_ID=aws_metrics
```

**Note**: The metrics connection must use your actual AWS credentials, otherwise the DAG will not be able to retrieve metrics from CloudWatch. The secret key **must** be URL encoded as well, you can do so with this command:

```bash
python -c 'import urllib.parse; safe_string = urllib.parse.quote_plus("<your-secret>"); print(safe_string)'
```

Once this is set up, run `just up` and turn on the DAG! The DAG should kick off a run on its own. This will generate a PNG from the metrics and upload it to your local Minio instance. You can view the generated PNG once the DAG is complete by navigating to http://localhost5011 and going to `openverse-airflow/metrics/elasticsearch/<date>/es-metrics.png` and then click "Preview".

Screenshot_2022-07-18_14-40-27

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


AetherUnbound:
Sure! Elasticsearch is consistently my greatest question mark and source of unknowns. Having anomaly detection is great, but the sustained erratic behavior we're seeing sort of constitutes a persistent anomaly. I am hoping that these graphs will let us see if/when that situation changes, and a snapshot into the health of our cluster. I'm foreseeing that most of the posts won't necessitate any intervention, and (ideally!) we'll turn this off once we've stabilized ES performance at a consistently lower level. It's not my desire to have this in perpetuity, but this make me personally feel a lot better because I get to see a picture of "what's happening with Elasticsearch today" (and so does the rest of the team). I could certainly go and check the dashboard every day for that information; for myself, I find the "push" aspect of slack notifications to be more beneficial for this particular case because it happens without me having to think about it.

sarayourfriend:
So, I have lots of questions about this still and I wrote a long comment asking them all. I don't think that's necessarily the most helpful, but I can share those questions later if necessary.

My primary questions relates to these sentences:

> [...] but this make me personally feel a lot better because I get to see a picture of "what's happening with Elasticsearch today" (and so does the rest of the team). I could certainly go and check the dashboard every day for that information; for myself, I find the "push" aspect of slack notifications to be more beneficial for this particular case because it happens without me having to think about it.

One of the daily tasks of the mini-support rotation role that we've recently implemented is to check the CloudWatch overview dashboard. That dashboard includes the graph in question in this PR.

Is the MSR's attention insufficient for the coarse grained monitoring we're promoting with this? If so, why?

Part of my understanding of the MSR role is to reduce the noise and ambiguity for other folks on the team. Pushing an image of a graph into an already heavily used notifications channel seems to go against the intentions of the MSR role.

sarayourfriend:
A firehose channel would make something like this less worrisome to me. I would probably silence it with the assumption that the MSR is monitoring it for any aberrations. However, then I still don't understand the point as it seems like a real-time and up-to-date dashboard is more relevant than this kind of thing.

I will leave it to other folks to discuss further. I've shared my thoughts, I don't think I have more to add :+1:

AetherUnbound:
I think that the circumstances we find ourselves in regarding ES usage are exceptional, and I think this would be good information for the whole team in addition to MSR monitoring. I don't see this being a permanent fixture and I fully agree that alarms would be more ideal.

I also wonder if an Airflow-specific notifications channel might make more sense as well, particularly as we get more reporting. Or perhaps even something like `#openverse-data-notifications` - this could serve as a place for the ingestion load messages as well as data refresh updates and any other reporting we want to add that doesn't fit alarms/general notifications. I'm interested in hearing what other folks think 🙂

krysal:
I like the idea of having a `#openverse-airflow-notification` or `#openverse-data-notifications` channel for this kind of thing. I also think we can benefit from having these metrics in Slack during this exceptional period, it's more accessible there, and maybe even other experienced folks can see a trend and give some hints.

> Is the MSR's attention insufficient for the coarse grained monitoring we're promoting with this? If so, why?
>
> Part of my understanding of the MSR role is to reduce the noise and ambiguity for other folks on the team. Pushing an image of a graph into an already heavily used notifications channel seems to go against the intentions of the MSR role.

The MSR role is to catch (almost) all the glue work, it makes certain things easier for the rest as a side effect, but ES CPU usage requires further investigation that is out of the scope of this role, will be convenient that more than one person from the team is monitoring.

Moreover, it's only an image, so I'm less concerned given this is temporary. Will it be running daily? Maybe every other day is a better schedule, given it reports the last 3 days?

obulat:
I don't know how to interpret the images, so I would prefer for this message to go to a separate channel such #openverse-data-notifications. Or at least to document the criteria for when this DAG becomes obsolete and can be turned off.

obulat (approved):

AetherUnbound:
I'm going to close this, we've got some new alerting thanks to @sarayourfriend which will tell us when ES CPU drops. That should cover the case this DAG was trying to cover well enough, and we can always revisit this down the line as well. If nothing else, it can be a good reference for the TaskFlow API in the future!

PR #614, opened 2022-07-20
Refactor Wikimedia Commons to use ProviderDataIngester

### LINKED ISSUE #598: Refactor Wikimedia Commons to use ProviderDataIngester
## Suggested Improvement

Refactor the Wikimedia commons provider script to use the new ProviderDataIngester base class

## Benefit

More details in #229

## Implementation

stacimc:
## Fixes

Fixes #598 by @stacimc 

## Description

Refactors Wikimedia Commons to use the ProviderDataIngester base class

### Notes
For Wikimedia Commons, the `batch_limit` applies to the number of `image_pages` to process, which is not necessarily a 1:1 relationship with records ingested. The result is that using a global `ingestion_limit` still works to limit the processing time but you may see slightly more records ingested.

## Testing Instructions

Run the Wikimedia Commons DAG. Also try setting an `ingestion_limit`

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
@@ -315,7 +315,9 @@ class ProviderDataIngester(ABC):
 
         return batch, should_continue
 
-    def get_response_json(self, query_params: dict, endpoint: str | None = None, **kwargs):
+    def get_response_json(
+        self, query_params: dict, endpoint: str | None = None, **kwargs
+    ):
         """
         Make the actual API requests needed to ingest a batch. This can be overridden
         in order to support APIs that require multiple requests, for example.
@@ -325,7 +327,7 @@ class ProviderDataIngester(ABC):
             self.retries,
             query_params,
             headers=self.headers,
-            **kwargs
+            **kwargs,
         )
 
     def get_should_continue(self, response_json):
--- a/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
@@ -315,7 +315,7 @@ class ProviderDataIngester(ABC):
 
         return batch, should_continue
 
-    def get_response_json(self, query_params: dict, endpoint: str | None = None):
+    def get_response_json(self, query_params: dict, endpoint: str | None = None, **kwargs):
         """
         Make the actual API requests needed to ingest a batch. This can be overridden
         in order to support APIs that require multiple requests, for example.
@@ -325,6 +325,7 @@ class ProviderDataIngester(ABC):
             self.retries,
             query_params,
             headers=self.headers,
+            **kwargs
         )
 
     def get_should_continue(self, response_json):
--- a/openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
@@ -222,18 +222,19 @@ class WikimediaCommonsDataIngester(ProviderDataIngester):
     def parse_audio_file_meta_data(self, media_info):
         """Parse out audio file metadata"""
         metadata = media_info.get("metadata", [])
+
         streams = self.get_value_by_name(metadata, "streams")
-        try:
-            if not streams:
-                audio = self.get_value_by_name(metadata, "audio")
-                streams = self.get_value_by_name(audio, "streams")
-            streams_data = [stream["value"] for stream in streams][0]
+        if not streams:
+            audio = self.get_value_by_name(metadata, "audio")
+            streams = self.get_value_by_name(audio, "streams")
+
+        if streams:
+            streams_data = streams[0].get("value", [])
             file_data = self.get_value_by_name(streams_data, "header")
-            if not file_data:
-                file_data = streams_data
-        except (IndexError, TypeError):
-            file_data = []
-        return file_data
+            # Fall back to streams_data
+            return file_data or streams_data
+
+        return []
 
     def get_media_info_dict(self, media_data):
         media_info_list = media_data.get("imageinfo")
@@ -245,7 +246,10 @@ class WikimediaCommonsDataIngester(ProviderDataIngester):
 
     def get_value_by_name(self, key_value_list: list, prop_name: str):
         """Gets the first value for the given `prop_name` in a list of
-        key value pairs"""
+        key value pairs."""
+        if key_value_list is None:
+            key_value_list = []
+
         prop_list = [
             key_value_pair
             for key_value_pair in key_value_list
--- a/openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
@@ -56,11 +56,11 @@ class WikimediaCommonsDataIngester(ProviderDataIngester):
     # the exact limit may not be respected.
     batch_limit = 250
 
-    def __init__(self, date: str | None = None):
+    def __init__(self, conf: dict = None, date: str | None = None):
         self.start_timestamp, self.end_timestamp = self.derive_timestamp_pair(date)
         self.continue_token = {}
 
-        super().__init__(date)
+        super().__init__(conf, date)
 
     def get_next_query_params(self, prev_query_params, **kwargs):
         return {
--- a/openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
@@ -400,14 +400,14 @@ class WikimediaCommonsDataIngester(ProviderDataIngester):
             merged_pages = self.get_media_pages(merged_json)
             merged_pages.update(
                 {
-                    k: self.merge_image_pages(left_pages[k], right_pages[k])
+                    k: self.merge_media_pages(left_pages[k], right_pages[k])
                     for k in left_pages
                 }
             )
 
         return merged_json
 
-    def merge_image_pages(self, left_page, right_page):
+    def merge_media_pages(self, left_page, right_page):
         merged_page = deepcopy(left_page)
         merged_globalusage = left_page.get("globalusage", []) + right_page.get(
             "globalusage", []
--- a/openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
@@ -106,7 +106,7 @@ class WikimediaCommonsDataIngester(ProviderDataIngester):
                 # Update continue token for the next request
                 self.continue_token = response_json.pop("continue", {})
                 query_params.update(self.continue_token)
-                logger.info(f"new_continue_token: {self.continue_token}")
+                logger.info(f"New continue token: {self.continue_token}")
 
                 # Merge this response into the batch
                 batch_json = self.merge_response_jsons(batch_json, response_json)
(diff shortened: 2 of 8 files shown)


openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -21,450 +21,424 @@
...
+        return batch_json
+
+    def get_should_continue(self, response_json):
+        # Should not continue if continue_token is Falsy
+        return self.continue_token
+
+    def get_batch_data(self, response_json):
+        image_pages = self.get_image_pages(response_json)
+        if image_pages is not None:
+            return image_pages.values()
```
AetherUnbound, line 127: Does this cover both images and audio? Should it be `media_pages`?

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -21,450 +21,424 @@
...
+            if not streams:
+                audio = self.get_value_by_name(metadata, "audio")
+                streams = self.get_value_by_name(audio, "streams")
+            streams_data = [stream["value"] for stream in streams][0]
+            file_data = self.get_value_by_name(streams_data, "header")
+            if not file_data:
+                file_data = streams_data
+        except (IndexError, TypeError):
+            file_data = []
+        return file_data
```
AetherUnbound, line 236: I know this is unaltered from the OG script, but could we alter this so that it checks for keys/indices rather than a try/except block?

tests/dags/providers/provider_api_scripts/test_wikimedia_commons.py
```diff
@@ -185,54 +184,56 @@ def test_merge_image_pages_both_have_gu():
...
+    record_data = wmc.get_record_data(media_data)
+
     expected_license_info = get_license_info(
         license_url="https://creativecommons.org/licenses/by-sa/4.0"
     )
-    assert mock_add.call_count == 1
-    assert mock_add.call_args == call(
-        foreign_landing_url=(
+    # assert mock_add.call_count == 1
+    # assert mock_add.call_args == call(
```
AetherUnbound, line 218: Did we still want to run similar checks?

tests/dags/providers/provider_api_scripts/test_wikimedia_commons.py
```diff
@@ -185,54 +184,56 @@ def test_merge_image_pages_both_have_gu():
...
     """
     Converts sample json data to correct image metadata,
     and calls `add_item` once for a valid image.
     """
     with open(RESOURCES / "image_data_example.json") as f:
         media_data = json.load(f)
 
-    with patch.object(wmc.image_store, "add_item", return_value=1) as mock_add:
-        wmc._process_media_data(media_data)
+    # with patch.object(wmc.image_store, "add_item", return_value=1) as mock_add:
```
AetherUnbound, line 211: Can this be removed?

AetherUnbound (commented):
Whew, this was a big one! 😄

tests/dags/providers/provider_api_scripts/test_wikimedia_commons.py
stacimc: I don't think it's necessary -- checking that the records are added to the correct media store should be covered in the tests for the base class. I should've removed these after commenting them out, fixed!

sarayourfriend:
@stacimc Is this PR still in progress? Is there anything you need to be able to continue the work on this?

AetherUnbound:
This is in draft pending the completion of the v1.3.0 milestone - once that's complete this can be rebased and undrafted!

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
stacimc: This bit of logic is actually somewhat fiddly because of the structure of the data we get back from the API, which involves lists of dicts with "name" and "value" keys ([example file](https://github.com/WordPress/openverse-catalog/blob/main/tests/dags/providers/provider_api_scripts/resources/wikimedia/audio_filedata_ogg.json#L7)). I pushed a refactor to remove the try/catch safely but I can see why they were added 😄

stacimc:
Changing this to high priority as it blocks our ability to turn on reingestion workflows.

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -21,450 +21,428 @@
...
+            ).strip()
+            meta_data["description"] = description_text
+        meta_data["global_usage_count"] = global_usage_length
+        meta_data["date_originally_created"] = date_originally_created
+        meta_data["last_modified_at_source"] = last_modified_at_source
+        meta_data["categories"] = categories_list
+        meta_data.update(self.get_geo_data(media_data))
+        return meta_data
+
+    def cleanse_url(self, url_string):
```
obulat, line 377: I believe this function is a left-over from before we had the ImageStore classes. We should probably use `validate_url_string` from https://github.com/WordPress/openverse-catalog/blob/2a9647abec0c78734fdd9b22f3be47450e79bbc4/openverse_catalog/dags/common/urls.py

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -21,450 +21,428 @@
...
+        artist_url = self.cleanse_url(url_list[0][2]) if url_list else None
+        return artist_text, artist_url
+
+    def extract_category_info(self, media_info):
+        categories_string = self.get_ext_value(media_info, "Categories") or ""
+
+        categories_list = categories_string.split("|")
+        return categories_list
+
+    def extract_file_type(self, media_info):
```
obulat, line 328: File extraction from the media URL field is done in the MediaStore if the `filetype` is None, so it's better to remove this ad-hoc function.

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
stacimc: Well spotted! In fact it looks like `validate_url_string` gets automatically applied to the [URLColumn](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/storage/columns.py#L503) so I don't think we need to call it here at all. I verified this by removing the url cleaning and then re-running Wikimedia locally. I added a log on `artist_url` and verified that we're getting urls like `//commons.wikimedia.org/wiki/User...` from the API, but that they are all formatted correctly in the generated TSV and in the database.

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
stacimc: I think we need this because for wikimedia we later do some extra handling on `filetype` to set it to [illustration for svgs](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py#L333).

AetherUnbound (commented):
Added a few more thoughts!

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
obulat: We were previously automatically setting all of the SVG files to `illustration`, so it makes sense to move this to the `ImageStore` class, and categorize the images after we extract their file types.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -320,13 +320,13 @@ def get_batch(self, query_params: Dict) -> Tuple[Optional[List], bool]:
 
         return batch, should_continue
 
-    def get_response_json(self, query_params: Dict):
+    def get_response_json(self, query_params: Dict, **kwargs):
```
AetherUnbound, line 323: Ooo, this is smart! Much more robust than my approach in #701

AetherUnbound (approved):
This looks great!

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
stacimc: To clarify -- you're saying that we can set media type to `illustration` for _all_ svg files regardless of provider? I'm wondering if it would be worth making that change separately from the refactor if it will potentially have a broader effect.

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -21,450 +21,428 @@
...
+    def get_media_type(self, record):
+        """Get the media_type of a parsed Record"""
+        return record["media_type"]
+
+    def get_response_json(self, query_params):
+        """
+        Overrides the parent function to make multiple requests until
+        we see "batchcomplete", rather than a single request to the
+        endpoint. This ensures that global usage data used for calculating
+        popularity is tabulated correctly.
```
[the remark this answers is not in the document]
stacimc: I went back and forth on this in the initial implementation and decided to override `get_response_json` because I thought it was more likely for `get_batch` to have future changes (for example we're looking at addressing error handling there right now). Good point about using `super().get_response_json` here! Done ✔️

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -21,450 +21,428 @@
...
+        }
+
+        # Extend record_data with media-type specific fields
+        funcs = {
+            IMAGE: self.get_image_record_data,
+            AUDIO: self.get_audio_record_data,
+        }
+        return funcs[valid_media_type](record_data, media_info)
+
+    def get_image_record_data(self, record_data, media_info):
```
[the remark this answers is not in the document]
stacimc: You're right! I think I was avoiding it just because most of them use `extract_ext_value` and accessing it statically is a bit clunky looking, but we probably should :)

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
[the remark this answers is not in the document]
AetherUnbound: Sounds great, and I love the changes you made there 💯

krysal (commented):
This looks like a very solid refactor! I ran the DAG and from a brief view, I have only one question. I'm seeing many titles including HTML tags, stuff like `<div class='fn'> Silver penny of Henry I</div>`. I haven't run this DAG before so I don't know if it was a bug before the refactor or if it was a change on the Wikimedia API or waht, though I definitely haven't seen those tags on the frontend.

Do we want to address that bug here?

stacimc:
@krysal Good catch! I just confirmed that the issue appears on `main`, so it's not a regression. I opened a new issue for this in #715.

krysal (approved):
Great! Thanks for confirming @stacimc. LGTM!

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
stacimc: #716 opened to track this.

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -21,450 +21,428 @@
...
+
+        media_url = media_info.get("url")
+        if media_url is None:
+            return None
+
+        creator, creator_url = self.extract_creator_info(media_info)
+        title = self.extract_title(media_info)
+        filesize = media_info.get("size", 0)  # in bytes
+        filetype = self.extract_file_type(media_info)
+        meta_data = self.create_meta_data_dict(record)
```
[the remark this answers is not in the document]
obulat: Could you open an issue for this, if you haven't already, @AetherUnbound ? It's nice to track it somewhere.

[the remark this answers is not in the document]
stacimc: Won't this be partially dependent on the class? `extract` and `get` have slightly different connotations to me. In this class `extract` felt useful because many of the utilities were a little more complex than simply pulling data out of a response.

If we really want to aim for consistency with/without `_` prefixing on the subclasses then I imagine we'd want a style guide or some sort of linting 🤔

PR #549, opened 2022-06-03
Add iNaturalist.org metadata

rwidom:
## Fixes
Fixes #439  by @sarayourfriend 

## Description
The [iNaturalist API](https://api.inaturalist.org/v1/docs/) is not intended for data scraping, but there is a [full dump](https://github.com/inaturalist/inaturalist-open-data/tree/documentation/Metadata) intended for sharing on S3. Those files cover over 120 million images, with information about the person who shared the photo and the species in the photo represented across four normalized tables. 

This PR loads *all* of the inaturalist metadata at once. The source files do not include any date or time fields to support differentiating among duplicate IDs or performing incremental updates. However, image files are in the same S3 bucket, prefixed by the photo_id in the metadata files. So, future dagruns could use [s3Hook.list_keys](https://airflow.apache.org/docs/apache-airflow-providers-amazon/3.3.0/_api/airflow/providers/amazon/aws/hooks/s3/index.html#airflow.providers.amazon.aws.hooks.s3.S3Hook.list_keys) to pull photo ID prefixes with files that have been updated since the [prev_start_date_success](https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html) and process only those photo IDs each month.

## Testing Instructions
Check out the branch. And then this PR includes changes to the docker dev setup, so it's a good idea to use `just down -v` and `just recreate` when you're testing. 

`just test` will cover basic tests of the provider data ingestion class functions, but not the DAG steps that make the metadata available .

There are very small subsets of the real files in /tests/s3-data/inaturalist-open-data, that will be loaded to the dev environment equivalent of s3, so that you can go kick off the inaturalist_workflow dag in Airflow, and inspect the logs and the resulting tsv file and the `image` table in the db. The Airflow log will give you the exact file name, and you can see it by running `just shell` and `more /var/workflow_output/inaturalist_image_v001_20220709220913.tsv` (or whatever the file is called). Run the dag a second time to confirm that the skips work when there is no new data.

I tried putting the real fullsize inaturalist files in /tests/s3-data/inaturalist-open-data, running `just down -v` and `just recreate`, and waiting a little bit for minio to finish loading the files. I've not been able to process the full dataset in my dev environment, but for now I am assuming that it is due to limitations of the environment. 

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
[...]

PR #620, opened 2022-07-21
Generate TSV filenames in separate step

### LINKED ISSUE #619: Generate TSV filename in separate DAG task
## Description

Presently the TSV filename used for a provider ingestion run is generated and pushed as part of a wrapper around the `pull_media_data` task. This has been working fine for now, but is beginning to cause problems in more atypical runs (e.g. reingestion workflows, iNaturalist pre-ingestion steps, etc.). It would be ideal for us to have an initial, separate step where the TSV filename for a run is generated. The loader steps would then reference the XCom created from that task rather than the `pull_media_data` task, which would prevent any issues with a case where the pull data task failed before the XCom was pushed (or in iNaturalist's case, where the pre-ingestion tasks failed before the pull data task).

## Implementation

AetherUnbound:
## Fixes

Fixes #619 by @AetherUnbound

## Description

This PR adds a separate task at the very beginning of the provider DAG factory which generates the TSV filenames up front. The pull data task now pulls the generated filenames and sets the appropriate store output path(s) prior to execution.

This should help us with #549 and #618 I believe! Ultimately I believe we should be able to completely alter our approach to TSV filenames down the road once we have all the provider scripts unified.

In doing this refactor, I split some functions out into a `factory_utils.py` module. I'm not sold on the name, so let me know if you think of any other names :slightly_smiling_face: I also moved some test files around to reflect their current position in the file tree.

### Screenshots
**Image DAG**
Screenshot_2022-07-21_14-32-52

**Audio DAG**
Screenshot_2022-07-21_14-32-30

**Mixed DAG**
Screenshot_2022-07-21_14-32-40

## Testing Instructions

Run an image DAG, and audio DAG, and the wikimedia DAG and ensure that each of the `copy_to_s3` and `pull_media` steps are using the filenames generated in generate TSV filenames step. The "Rendered" tab for a certain task in the UI can help quickly determine what values the downstream tasks received for the filenames.
Screenshot_2022-07-21_14-38-36

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


openverse_catalog/dags/providers/factory_utils.py
```diff
@@ -0,0 +1,137 @@
...
+            f"Setting {store.media_type.capitalize()} store location "
+            f"to {tsv_filename}"
+        )
+        store.output_path = tsv_filename
+
+    logger.info("Beginning ingestion")
+    start_time = time.perf_counter()
+    # Not passing kwargs here because Airflow throws a bunch of stuff in there that
+    # none of our provider scripts are expecting.
+    data = run_func(*args)
```
stacimc, line 132: This is unfortunate because I love how nice the refactor here looks -- but while we *do* need to pass `*args` here for legacy scripts, we don't actually want to pass them in here if we're using an ingester class.

For those, we pass `*args` to the `__init__` but we don't want to pass them to `ingest_records` as well, or else it causes problems with dated DAGs when the `date` is unexpectedly passed in (I found/fixed this separately in #618).

openverse_catalog/dags/providers/factory_utils.py
```diff
@@ -0,0 +1,137 @@
...
+
+    else:
+        # A ProviderDataIngester class was passed instead. First we initialize the
+        # class, which will initialize the media stores and DelayedRequester.
+        logger.info(
+            f"Initializing ProviderIngester {ingestion_callable.__name__} in"
+            f"order to generate store filenames."
+        )
+        ingester = ingestion_callable(*args)
+        stores = ingester.media_stores
```
stacimc, line 73: You reference refactoring this down the line in the PR description so this is probably already the plan, but -- I'd love to pull the filename generation out of the media store and actually perform the logic here instead of initializing the stores twice. This would make it easier to add conditional logic (in my case, like adding a `date` suffix for tsvs generated during reingestion) without having to make the MediaStore aware of those parameters.

Totally reasonable to wait until the provider scripts are refactored though, to avoid duplicating the logic. 👍

tests/dags/providers/test_provider_dag_factory.py
```diff
@@ -0,0 +1,48 @@
+from providers import provider_dag_factory
+
```
stacimc, line 2: Appreciate this, I had _just_ made a note this morning to move this file when I went looking for it under this directory and couldn't find it! Nice ✨

stacimc (commented):
This looks great! Tested with Cleveland (image provider using the new base class), Jamendo (audio provider, legacy), and Wikimedia (image and audio provider). Wikimedia weirdly generates filenames for the audio tsv that look like this: `wikimedia_audio_audio_v001_20220721223643.tsv` (not sure why audio is duplicated?) but that's definitely happening on main so not a problem here.

Tests look great as always :) Do you think we could add coverage for the ingester class path? That was an oversight on my part when the base class was added.

AetherUnbound:
> Wikimedia weirdly generates filenames for the audio tsv that look like this: `wikimedia_audio_audio_v001_20220721223643.tsv` (not sure why audio is duplicated?) but that's definitely happening on main so not a problem here.

I noticed that as well 😅 I think it's because the actual provider is defined as `wikimedia_audio`:

https://github.com/WordPress/openverse-catalog/blob/cd7ca961c169320af1a3c69c514d880b9ac0dd85/openverse_catalog/dags/common/loader/provider_details.py#L18
 
> Tests look great as always :) Do you think we could add coverage for the ingester class path? That was an oversight on my part when the base class was added.

Can do!

openverse_catalog/dags/providers/factory_utils.py
AetherUnbound: Ah, okay I was a little thrown off when I first saw that! I was only copying what was present in the original code: https://github.com/WordPress/openverse-catalog/blob/cd7ca961c169320af1a3c69c514d880b9ac0dd85/openverse_catalog/dags/providers/provider_dag_factory.py#L138-L150

I think I should be able to adjust things here to make that case work!

openverse_catalog/dags/providers/factory_utils.py
AetherUnbound: 100% agree! And yea, I'd rather have the paths look similar now while we still have two, and then totally shift stuff around when we only have the one to worry about.

openverse_catalog/dags/providers/factory_utils.py
```diff
@@ -117,6 +117,9 @@ def pull_media_wrapper(
         ingester = ingestion_callable(*args)
         stores = ingester.media_stores
         run_func = ingester.ingest_records
+        # args have already been passed into the ingester, we don't need them passed
+        # in again to the ingest_records function, so we clear the list
+        args = []
```
AetherUnbound, line 122: @stacimc do you think this is sufficient to cover that case?

AetherUnbound:
@stacimc I've added tests for the ingester class code path as part of 98ec3a96ade2e6abd1d43d3745a2132c6158902a. It was a little funky...just because of the way we initialize and call the different pieces. Let me know if it's super confusing and I can try to add more documentation/split things out. The good news is that this now also tests to make sure that only the ingester initialization receives `*args` and that `ingest_records` receives nothing!

openverse_catalog/dags/providers/factory_utils.py
stacimc: I'm fine with that -- it'll be easy to clean up anyway once the refactors are done :)

openverse_catalog/dags/providers/factory_utils.py
stacimc: Yes, exactly! I meant I noticed it on main as well when working on #618, but that's still in draft. It's wasn't found before because none of the dated DAGs have been refactored yet. I like your fix!

tests/dags/providers/test_factory_utils.py
```diff
@@ -0,0 +1,210 @@
...
+        ),
+        # Empty case, no media types provided
+        (fake_provider_module.main, [], []),
+        ##################################################
+        # Class-based
+        ##################################################
+        # Happy path
+        (FakeDataIngesterClass, ["image", "audio"], list(fdi.media_stores.values())),
+        # No media types provided, ingester class still supplies stores
+        (FakeDataIngesterClass, 2, list(fdi.media_stores.values())),
```
stacimc, line 110: This isn't a reachable code path, right? Just proving that when an IngesterClass is supplied it doesn't matter what value is passed in for the `media_stores`, since it isn't used?

stacimc (approved):
LGTM! Fast work on this one 🚀

tests/dags/providers/test_factory_utils.py
AetherUnbound: That's right! I assume these tests will be altered when we can simplify all this down and these will probably go away 🙂

rwidom:
> This looks great! Tested with Cleveland (image provider using the new base class), Jamendo (audio provider, legacy), and Wikimedia (image and audio provider). Wikimedia weirdly generates filenames for the audio tsv that look like this: `wikimedia_audio_audio_v001_20220721223643.tsv` (not sure why audio is duplicated?) but that's definitely happening on main so not a problem here.

I wonder if the weird name is because of [these details](https://github.com/WordPress/openverse-catalog/blob/cd7ca961c169320af1a3c69c514d880b9ac0dd85/openverse_catalog/dags/common/loader/provider_details.py#L18)? Is that the kind of thing that would have a million tentacles and be impossible to change, or might it be easy to generalize naming for multi-type providers?

rwidom:
This is just to say looks good to me. :) 

I ran Brooklyn Museum, and `generate_image_filename` sent this to xcoms: /var/workflow_output/brooklynmuseum_image_v001_20220725215300.tsv. Then `copy_to_s3` had the same in the rendered template. 

I ran Freesound, and `generate_audio_filename` sent this to xcoms: /var/workflow_output/freesound_audio_v001_20220725220933.tsv. Then authorization failed, so `pull_audio_data` failed too, but `copy_to_s3` had the right filename in the rendered template.

I ran Wikimedia Commons and wondered to myself a little (not enough to look it up though) how the downstream tasks would know which filename to use. Based on searching for the media type in the file name?  Regardless, the copy to s3 tasks for each media type did get the specific file name generated for them in their rendered template.

Yay! :)

AetherUnbound:
> I ran Wikimedia Commons and wondered to myself a little (not enough to look it up though) how the downstream tasks would know which filename to use. Based on searching for the media type in the file name? Regardless, the copy to s3 tasks for each media type did get the specific file name generated for them in their rendered template.

That's right! The generation task publishes multiple named XComs which are referenced downstream by media type:

https://github.com/WordPress/openverse-catalog/blob/9b6e788ad1f4c32cca46721e2576ca7dd43c1b56/openverse_catalog/dags/providers/provider_dag_factory.py#L208

rwidom (approved):
LGTM!

PR #626, opened 2022-07-25
Adjust load data timeout and retries

stacimc:
## Description

* Adds a `load_timeout` option to the `ProviderWorkflow` configurations to control the `execution_timeout` of the load data step. Defaults to 1hr (which is the current default, so DAGs are by default unchanged)
* Removes `retries` for the load data step, as errors in this step are non-transient and don't benefit from being retried
* Updates the timeouts for the Finnish Museum dag, which is currently timing out on both the pull and load steps.

## Testing Instructions

`just up`, make sure the DAGs all load. You can inspect the Task Instance Details for the pull and load steps of a few DAGs to make sure the timeouts look correct. To do this you will have to have a DagRun, so you may need to trigger the Dag. The Task Details view is accessible from the menu by clicking on the task in the Graph view.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_dag_factory.py
+++ b/openverse_catalog/dags/providers/provider_dag_factory.py
@@ -166,7 +166,8 @@ def create_provider_api_workflow(
     schedule_string: str = "@daily",
     dated: bool = True,
     day_shift: int = 0,
-    execution_timeout: timedelta = timedelta(hours=12),
+    pull_timeout: timedelta = timedelta(hours=12),
+    load_timeout: timedelta = timedelta(hours=1),
     doc_md: Optional[str] = "",
     media_types: Sequence[str] = ("image",),
 ):
@@ -205,8 +206,10 @@ def create_provider_api_workflow(
     day_shift:         integer giving the number of days before the
                        current execution date the `main_function` should
                        be run (if `dated=True`).
-    execution_timeout: datetime.timedelta giving the amount of time a given data
+    pull_timeout:      datetime.timedelta giving the amount of time a given data
                        pull may take.
+    load_timeout:      datetime.timedelta giving the amount of time a given load_data
+                       task may take.
     doc_md:            string which should be used for the DAG's documentation markdown
     media_types:       list describing the media type(s) that this provider handles
                        (e.g. `["audio"]`, `["image", "audio"]`, etc.)
@@ -242,7 +245,7 @@ def create_provider_api_workflow(
             python_callable=_push_output_paths_wrapper,
             op_kwargs=pull_kwargs,
             depends_on_past=False,
-            execution_timeout=execution_timeout,
+            execution_timeout=pull_timeout,
             # If the data pull fails, we want to load all data that's been retrieved
             # thus far before we attempt again
             retries=0,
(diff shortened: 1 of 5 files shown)


zackkrida:
@stacimc looks like a MacOS filesystem `openverse_catalog/dags/.DS_Store` file got committed, so this might be a good time to add them to the `.gitignore` for the repo

stacimc:
@zackkrida 🤦‍♀️ Yep, thanks. Added in #627

openverse_catalog/dags/providers/provider_dag_factory.py
```diff
@@ -279,6 +282,8 @@ def create_provider_api_workflow(
                 )
                 load_from_s3 = PythonOperator(
                     task_id="load_from_s3",
+                    execution_timeout=load_timeout,
```
rwidom, line 285: Is it just that that 1 hr is the default timeout for _any_ task in airflow right now, so this makes that more explicit and flexible? Also, should I be thinking about adding something like a `preingestion_timeout` to the inaturalist PR or a follow-up?

AetherUnbound: I believe 1h is the default for all tasks as defined by the DAG default args. We've been finding that the load step for some DAGs (Finnish and Smithsonian for example) are considerably longer, but this is a step we'd still want to ensure takes less than 1 hour for DAGs like Wikimedia. This allows us extra flexibility for those cases!

As for the preingestion tasks, I think the default timeout should be sufficient 🙂 No need to add extra complication/configuration options up front if we may never use them. If we do encounter a scenario where the 1h timeout doesn't make sense for those tasks though, we can revisit!

openverse_catalog/dags/providers/provider_workflows.py
```diff
@@ -122,11 +126,11 @@ def __post_init__(self):
...
-        execution_timeout=timedelta(hours=12),
+        pull_timeout=timedelta(hours=12),
     ),
     ProviderWorkflow(
         provider_script="rawpixel",
-        execution_timeout=timedelta(hours=12),
+        pull_timeout=timedelta(hours=12),
     ),
     ProviderWorkflow(
         provider_script="science_museum",
```
AetherUnbound, line 136: Do you mind also bumping up Smithsonian's `load_timeout` to 4 hours in this PR?

AetherUnbound (approved):
I love this! The provider workflow dataclasses make this so easy 😌 Just one request, but nothing to block a merge 🚀

openverse_catalog/dags/providers/provider_workflows.py
```diff
@@ -89,7 +92,8 @@ def __post_init__(self):
     ProviderWorkflow(
         provider_script="finnish_museums",
         start_date=datetime(2020, 9, 1),
-        execution_timeout=timedelta(days=3),
+        pull_timeout=timedelta(days=5),
```
rwidom, line 95: FWIW, I tried switching this to `pull_timeout=timedelta(seconds=5)` and running the dag and it did time out after 5 seconds, so that's good. :)

openverse_catalog/dags/providers/provider_dag_factory.py
rwidom: Awesome, thanks so much @AetherUnbound !

rwidom (approved):

openverse_catalog/dags/providers/provider_workflows.py
stacimc: Knew I forgot something! Yes, of course :)

PR #629, opened 2022-07-26
Re-raise pytest-socket errors within DelayedRequester

### LINKED ISSUE #624: Stocksnap test takes 40s to run
## Description

While running tests recently, I realized that some tests were taking an incredibly long time. After running `just test --durations=5`, I got the following output:

```
=============================================================================================== slowest 5 durations ================================================================================================
39.04s call     tests/dags/providers/provider_api_scripts/test_stocksnap.py::test_process_batch
2.00s call     tests/dags/common/test_requester.py::test_get_delays_processing
1.03s call     tests/dags/common/loader/test_sql.py::test_upsert_records_replaces_updated_on_and_last_synced_with_source
0.42s call     tests/dags/common/loader/test_sql.py::test_loaders_load_good_tsv[_load_s3_tsv]
0.25s call     tests/dags/common/sensors/test_single_run_external_dags_sensor.py::TestExternalDAGsSensor::test_fails_if_external_dag_does_not_exist
```

This seems to be a recent development, possibly related to #601

## Resolution

AetherUnbound:
## Fixes

Fixes #624 by @AetherUnbound

## Description

This PR _does_ address the root cause of #624, wherein `pytest-socket` exceptions were being raised and squelched by the `DelayedRequester` class unconditionally. This meant that, even though requests would not get through, the entire request retry mechanism would run and take up a significant amount of time.

Since the specific exception we need to catch will only be available in the dev container and not the one we use in production, we also needed to set up an import catch for that particular case. This change ensures that if that specific exception is met (or any others we might need to single out in the future), it is re-raised for the testing infrastructure to report rather than being caught.

~~**Note**: CI will fail for this DAG until we merge #628 and rebase this!~~ This was rebased and is now working as expected!

## Testing Instructions

Run `just test` on c294e3e2bfc366a8e54e6897997dbd260466a94d and ensure that the tests run successfully (albeit around ~40s or more). Then run `just test` on this branch and ensure that there is 1 Stocksnap test failure that is caused by a `pytest_socket.SocketConnectBlockedError` exception.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/requester.py
+++ b/openverse_catalog/dags/common/requester.py
@@ -5,6 +5,16 @@ import oauth2
 import requests
 
 
+# pytest_socket will not be available in production, so we must create a shim for
+# that specific exception so this code does not error out.
+try:
+    from pytest_socket import SocketConnectBlockedError
+except ImportError:
+
+    class SocketConnectBlockedError(Exception):
+        pass
+
+
 logger = logging.getLogger(__name__)
 
 
@@ -51,6 +61,10 @@ class DelayedRequester:
                     f"Status code: {response.status_code}"
                 )
             return response
+        except SocketConnectBlockedError:
+            # This exception will only be raised during testing, and it *must*
+            # be re-raised and bubbled up the stack
+            raise
         except Exception as e:
             logger.error(f"Error with the request for URL: {url}.")
             logger.info(f"{type(e).__name__}: {e}")


openverse_catalog/dags/common/requester.py
```diff
@@ -5,6 +5,16 @@
 import requests
 
 
+# pytest_socket will not be available in production, so we must create a shim for
+# that specific exception so this code does not error out.
+try:
+    from pytest_socket import SocketConnectBlockedError
```
rwidom, line 11: So, it's saying "if we already imported `pytest_socket` somewhere else, great, use that, otherwise use this nothing thing", right? ("shim" is new to me, but it means the nothing thing and/or the wrapper around it?) That's cool!

openverse_catalog/dags/common/requester.py
```diff
@@ -5,6 +5,16 @@
...
+except ImportError:
+
+    class SocketConnectBlockedError(Exception):
+        pass
```
rwidom, line 15: I followed the testing instructions and it behaved as expected -- slow pass as of c294e3e, and this one failure on stocksnap with the full branch:

```
pytest_socket.SocketConnectBlockedError: A test tried to use socket.socket.connect() with host "172.67.73.94" (allowed: "s3,postgres,172.19.0.3").

.local/lib/python3.10/site-packages/pytest_socket.py:186: SocketConnectBlockedError
```

So I believe that it works, but I'm not 100% sure _why_ it works. [StackOverflow](https://stackoverflow.com/questions/13886168/how-to-use-the-pass-statement) says a thing about "Deriving an exception class that does not add new behaviour". Is this how it works? Exceptions are classes, and when we say the "type" of the exception we literally mean the "type" as in the class name. So this is making sure that there is such a thing as a `SocketConnectBlockedError` and we always need to fully raise it as an error (not just throw it in the log, and move on), but we know for sure that it will only come up in testing, because ??? and even if it did come up in production, that's fine because it would just be like any abstract exception and raising it would offer some more useful info?

rwidom (approved):
Cool! Thanks!

openverse_catalog/dags/common/requester.py
krysal: The exception comes from the **pytest_socket** package that is only available as a [dev](https://github.com/WordPress/openverse-catalog/blob/main/requirements_dev.txt#L7) dependency, I believe that is why @AetherUnbound says it will not come up in production and also is the reason for this bit weird of setup :)

krysal (approved):
It should have been tricky to identify the cause, as this setup! Looks good to me 👍

openverse_catalog/dags/common/requester.py
AetherUnbound: Great question! Usually with python imports, it's necessary to import what you need in whatever file it's required in. Python then manages which modules have already been imported by using a [module cache](https://docs.python.org/3/reference/import.html#the-module-cache). So if we `import foo` in file `a.py` which gets loaded first, and we import it in `b.py` which gets loaded second, the second `import foo` will just reference the module that was imported as part of `a.py`.

This is a bit of a different case. The Docker images that we build for local development and for production are different. We don't need the development dependencies (like pytest, code linters, etc) when we're running Airflow in production; they're only useful when we're working on the project on our machines. As such, the images use different requirements files when being built (prod [arg](https://github.com/WordPress/openverse-catalog/blob/8754cb63c075e2045b91af7bee278566d930286b/docker/airflow/Dockerfile#L4), [file](https://github.com/WordPress/openverse-catalog/blob/8754cb63c075e2045b91af7bee278566d930286b/requirements_prod.txt)) (dev [arg](https://github.com/WordPress/openverse-catalog/blob/8754cb63c075e2045b91af7bee278566d930286b/docker-compose.override.yml#L57), [file](https://github.com/WordPress/openverse-catalog/blob/8754cb63c075e2045b91af7bee278566d930286b/requirements_dev.txt)).

All that to say: when we're running in production, the python interpreter there _won't have `pytest_socket` available at all_. While we want that exact exception for local testing, we can just create a [shim](https://en.wikipedia.org/wiki/Shim_(computing)) (or sham 😜) exception which is used in its place in production. In other words, "if we don't have `pytest_socket` installed then just reference this empty exception class instead". All this is contingent on the fact that this exception **will not and cannot** be raised in production (since we don't have the library anyway!).

openverse_catalog/dags/common/requester.py
AetherUnbound: Explained this a bit more in depth [here](https://github.com/WordPress/openverse-catalog/pull/629#discussion_r930513147), lemme know there if there's anything else confusing about this!

AetherUnbound:
> It should have been tricky to identify the cause, as this setup!

Intuition and sheer happenstance graced me with a quick realization of what it might be 😄 I'm thankful it was easy enough to diagnose once I had an idea!

openverse_catalog/dags/common/requester.py
rwidom: Awesome! Thank you so much @AetherUnbound !

PR #632, opened 2022-07-26
Partition TSVs by date

### LINKED ISSUE #449: Partition TSV files in S3 by year and month as well
## Problem

Our TSV files in S3 are currently partitioned by media type, then provider. This is useful for now, but as we get more and more TSVs it will become difficult to find a particular one.

image

## Description

We should partition the key by year & month as well, that way we only have a maximum of 31 files under a given prefix. In many cases this will lead to a single file nested in a few directories, but that seems a reasonable alternative.

## Additional context

We could partition differently depending on the schedule interval of the provider scripts. For instance, daily & weekly DAGs could be partitioned by `year/month`, while monthly DAGs could just be partitioned by `year`.

## Implementation

AetherUnbound:
## Fixes

Fixes #449 by @AetherUnbound

## Description

This PR adds some additional logic to the provider DAG factory which includes date information in the S3 prefix for the generated TSV. The prefix includes year, month, and day (the latter two fields are conditional on how frequently the DAG is run). 

The function added in `factory_utils.py` is added as a [`user_defined_macro`](https://airflow.apache.org/docs/apache-airflow/1.10.12/tutorial.html?highlight=user_defined_macros#templating-with-jinja) to the DAG. This allows it to be used as a Jinja macro within templated parameters similar to [those available by default with Airflow](https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html). Its call pattern is similar to our `XCOM_PULL_TEMPLATE` setup. The end result is that the generated S3 key used to upload the TSV created during ingestion is also partitioned by the date parts described above.

## Testing Instructions

Try running provider DAGs with various schedule intervals (e.g. `@weekly`, `@monthly`, `@daily`). You should be able to click on the `copy_to_s3`'s `Rendered templates` tab and view the rendered `s3_prefix`. Depending on how far the DAG gets in processing, you can also do the same on `load_to_s3` using the `key` parameter.

You can also visit minio at http://localhost:5011/ after some ingestions have run to observe the new file structure within the bucket.

### Screenshots

**Copy to S3 rendered params**
Screenshot_2022-07-26_16-15-34

**Load from S3 rendered params**
Screenshot_2022-07-26_16-12-04

**Minio file structure**
Screenshot_2022-07-26_16-36-41

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/factory_utils.py
+++ b/openverse_catalog/dags/providers/factory_utils.py
@@ -5,8 +5,10 @@ from types import FunctionType
 from typing import Callable, Sequence
 
 from airflow.models import TaskInstance
+from airflow.utils.dates import cron_presets
 from common.constants import MediaType
 from common.storage.media import MediaStore
+from pendulum import datetime
 
 
 logger = logging.getLogger(__name__)
@@ -138,3 +140,34 @@ def pull_media_wrapper(
     duration = end_time - start_time
     ti.xcom_push(key="duration", value=duration)
     return data
+
+
+def date_partition_for_prefix(
+    schedule_interval: str | None, logical_date: datetime
+) -> str:
+    """
+    Given a schedule interval and the logical date for a DAG run, determine an
+    appropriate partition for the run. This partition will be used as part of the S3
+    key prefix for a given TSV.
+
+    Prefix mapping (schedule interval to partition):
+        - Hourly -> `year=YYYY/month=MM/day=DD`
+        - Daily -> `year=YYYY/month=MM`
+        - None/yearly/monthly/weekly/other -> `year=YYYY`
+    """
+    # Always partition by year
+    prefix = f"year={logical_date.year}"
+    hourly_airflow = "@hourly"
+    hourly_cron = cron_presets[hourly_airflow]
+    daily_airflow = "@daily"
+    daily_cron = cron_presets[daily_airflow]
+
+    # Add month in daily/hourly cases
+    if schedule_interval in {hourly_airflow, hourly_cron, daily_airflow, daily_cron}:
+        prefix += f"/month={logical_date.month:02}"
+
+    # Add day to hourly cases
+    if schedule_interval in {hourly_airflow, hourly_cron}:
+        prefix += f"/day={logical_date.day:02}"
+
+    return prefix
--- a/openverse_catalog/dags/providers/factory_utils.py
+++ b/openverse_catalog/dags/providers/factory_utils.py
@@ -155,12 +155,13 @@ def date_partition_for_prefix(
         - Daily -> `year=YYYY/month=MM`
         - None/yearly/monthly/weekly/other -> `year=YYYY`
     """
-    # Always partition by year
-    prefix = f"year={logical_date.year}"
     hourly_airflow = "@hourly"
     hourly_cron = cron_presets[hourly_airflow]
     daily_airflow = "@daily"
     daily_cron = cron_presets[daily_airflow]
+
+    # Always partition by year
+    prefix = f"year={logical_date.year}"
 
     # Add month in daily/hourly cases
     if schedule_interval in {hourly_airflow, hourly_cron, daily_airflow, daily_cron}:
--- a/openverse_catalog/dags/providers/provider_dag_factory.py
+++ b/openverse_catalog/dags/providers/provider_dag_factory.py
@@ -79,7 +79,7 @@ AWS_CONN_ID = os.getenv("AWS_CONN_ID", "no_aws_conn_id")
 OPENVERSE_BUCKET = os.getenv("OPENVERSE_BUCKET")
 OUTPUT_DIR_PATH = os.path.realpath(os.getenv("OUTPUT_DIR", "/tmp/"))
 DATE_RANGE_ARG_TEMPLATE = "{{{{ macros.ds_add(ds, -{}) }}}}"
-DATE_PARTITION_ARG_TEMPLATE = "{media_type}/{provider_type}/{{{{ date_partition_for_prefix(dag.schedule_interval, dag_run.logical_date) }}}}"  # noqa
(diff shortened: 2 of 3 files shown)


sarayourfriend:
> The prefix includes year, month, and day (the latter two fields are conditional on how frequently the DAG is run).

What is the reasoning behind not always including the month and day?

AetherUnbound:
> What is the reasoning behind not always including the month and day?

My reasoning is based on an operator perspective and navigating down a series of S3 buckets to get to the actual TSV file one would need. If we unconditionally included month and day, monthly runs would each be 2 more levels deep with only one file within a given day. This makes actions like visually comparing file size more difficult: 
image

Having files on the same level makes them easier to compare, but having hundreds of files in a single level (e.g. daily DAGs over time) can make it even more difficult to find the appropriate file. I believe the current partitioning should provide a solid middle ground.
Screenshot 2022-07-27 at 13-42-57 openverse-catalog - S3 bucket

sarayourfriend:
Gotcha! That's helpful for understanding the logic in the function as well. Thanks.

openverse_catalog/dags/providers/factory_utils.py
```diff
@@ -5,8 +5,10 @@
 from typing import Callable, Sequence
 
 from airflow.models import TaskInstance
+from airflow.utils.dates import cron_presets
 from common.constants import MediaType
 from common.storage.media import MediaStore
+from pendulum import datetime
```
sarayourfriend, line 11: When does the catalog repository use pendulum vs the stdlib datetime?

openverse_catalog/dags/providers/provider_dag_factory.py
```diff
@@ -75,6 +79,7 @@
 OPENVERSE_BUCKET = os.getenv("OPENVERSE_BUCKET")
 OUTPUT_DIR_PATH = os.path.realpath(os.getenv("OUTPUT_DIR", "/tmp/"))
 DATE_RANGE_ARG_TEMPLATE = "{{{{ macros.ds_add(ds, -{}) }}}}"
+DATE_PARTITION_ARG_TEMPLATE = "{}/{}/{{{{ date_partition_for_prefix(dag.schedule_interval, dag_run.logical_date) }}}}"  # noqa
```
sarayourfriend, line 82: Can the first two holes have names, so it's clear in the template what they're for?

```suggestion
DATE_PARTITION_ARG_TEMPLATE = "{media_type}/{provider_type}/{{{{ date_partition_for_prefix(dag.schedule_interval, dag_run.logical_date) }}}}"  # noqa
```

openverse_catalog/dags/providers/provider_dag_factory.py
```diff
@@ -208,7 +214,9 @@ def create_provider_api_workflow(
                             generate_filenames.task_id, f"{media_type}_tsv"
                         ),
                         "s3_bucket": OPENVERSE_BUCKET,
-                        "s3_prefix": f"{media_type}/{provider_name}",
+                        "s3_prefix": DATE_PARTITION_ARG_TEMPLATE.format(
+                            media_type, provider_name
```
sarayourfriend, line 218: In case you accept the suggestion above to name the holes:

```suggestion
                            media_type=media_type,
                            provider_name=provider_type,
```

sarayourfriend (approved):
LGTM! Just one suggestion to clarify the template and a curiosity question :slightly_smiling_face:

openverse_catalog/dags/providers/factory_utils.py
AetherUnbound: Ha, that's a good question. The `pendulum.DateTime` class (which is returned from `pendulum.datetime` function) inherits from `datetime.datetime`, and Airflow uses `pendulum` internally. One could probably be substituted for the other (esp. from a functionality perspective), but the values that get handed into functions by Airflow are `pendulum.DateTime` instances[^1]. So for the purposes of typing here, it's more accurate to use `pendulum.datetime` over `datetime.datetime`.

[^1]: See the example `data_interval` template variables [here](https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html#variables)

openverse_catalog/dags/providers/provider_dag_factory.py
AetherUnbound: Thanks!

openverse_catalog/dags/providers/factory_utils.py
sarayourfriend: Why don't we use pendulum everywhere in this repository?

AetherUnbound: We certainly could! `datetime.datetime` is in the standard library so that's just the one I typically reach for. There's nothing stopping us from using `pendulum` everywhere instead I don't think.

sarayourfriend: Consistency would be nice and reduces the number of things to think about or pay attention to (this function takes a regular date time, this one takes pendulum, etc), but if it doesn't make sense to only use one or the other though, no worries.

openverse_catalog/dags/providers/factory_utils.py
```diff
@@ -138,3 +140,34 @@ def pull_media_wrapper(
...
+    appropriate partition for the run. This partition will be used as part of the S3
+    key prefix for a given TSV.
+
+    Prefix mapping (schedule interval to partition):
+        - Hourly -> `year=YYYY/month=MM/day=DD`
+        - Daily -> `year=YYYY/month=MM`
+        - None/yearly/monthly/weekly/other -> `year=YYYY`
+    """
+    # Always partition by year
+    prefix = f"year={logical_date.year}"
```
stacimc, line 159: Total nit, but I would move this line below the schedule string/cron variables, so that it's with the month/day prefixes. The way it is structured currently makes it a little easy to miss.

stacimc (approved):
Sorry, thought I already reviewed this one -- LGTM and a great organizational improvement! I tested with Wikimedia (daily), Phylopic (weekly), and Cleveland (monthly).

PR #634, opened 2022-07-27
Only drop load table if it exists

AetherUnbound:
## Fixes

This addresses an issue identified by @rwidom as part of #549: Previously, our `pull_data` tasks would either succeed or fail, it was not possible for them to be skipped (even if they didn't pull any data). Now that we're introducing pre-ingestion tasks, it's possible that upstream tasks might skip the rest of the workflow. In these cases, the `create_table` step was skipped successfully, but the `delete_table`'s trigger rules were setup such that it was unconditionally run. This would then cause issues where it would try to delete a table that did not exist (see [here](https://github.com/WordPress/openverse-catalog/pull/549#discussion_r930949202)).

## Description

This PR adds an `IF EXISTS` clause to the drop table SQL to ensure that, even if the create table step doesn't run, the unconditionally run drop table step should succeed.

As part of this, I also wanted to see if I could add a test to catch this (and make sure the suggested fix actually addressed the issue). This proved to be more difficult than I thought, since Airflow tests usually involve task-specific unit tests or general DAG parsing rather than a full DAG run. I was able to get this working with the [`DebugExecutor`](https://airflow.apache.org/docs/apache-airflow/stable/executor/debug.html), but it turns out that each run takes at lest 20 seconds (and a total of 3 tests means that's at least an extra minute on tests).

These tests don't need to be run locally every time we're doing rapid iteration, so I opted to set up a new pytest CLI parameter which could selectively skip these longer-running tests. By default, `pytest` will now skip those long tests, but `pytest --extended` will run all of the available tests. Here's the time comparison:

```bash
$ pytest
...

Results (7.63s):
    1307 passed
       3 skipped

$ pytest --extended
...

Results (84.10s):
    1310 passed
```

As such, I've also added the `--extended` flag to our CI/CD so it always gets run on GitHub Actions for PRs.

## Testing Instructions

1. `just recreate`
1. Run `git checkout 3316d6d9243b90487e4ccbdd3ffaee141154f37d` (this will checkout the commit which proves the current setup fails)
1. Run `just test` and observe failure
1. Run `git checkout bugfix/drop-table-if-exists` (this will switch back to the tip of this branch)
1. Run `just test` and notice the rapid test completion
1. Run `just test --extended` and notice the longer test completion with no tests skipped

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -79,7 +79,7 @@ jobs:
         # rather than the production ones
         run: |
           just build
-          just test
+          just test --extended
 
   push:
     name: Publish Docker images
--- a/README.md
+++ b/README.md
@@ -187,6 +187,9 @@ At this stage, you can run the tests via:
 
 ```shell
 just test
+
+# Alternatively, run all tests including longer-running ones
+just test --extended
 ```
 
 Edits to the source files or tests can be made on your local machine, then tests
--- a/openverse_catalog/dags/common/loader/sql.py
+++ b/openverse_catalog/dags/common/loader/sql.py
@@ -304,7 +304,7 @@ def drop_load_table(
 ):
     load_table = _get_load_table_name(identifier, media_type=media_type)
     postgres = PostgresHook(postgres_conn_id=postgres_conn_id)
-    postgres.run(f"DROP TABLE {load_table};")
+    postgres.run(f"DROP TABLE IF EXISTS {load_table};")
 
 
 def _get_load_table_name(
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,24 @@
+import pytest
+
+
+def pytest_addoption(parser):
+    """
+    This functions alters the pytest CLI command options. It adds an "extended" flag
+    which will run tests that take a significant amount of time that may not be useful
+    for rapid local iteration.
+
+    Adapted from:
+    https://stackoverflow.com/a/43938191/3277713 CC BY-SA 4.0
+    """
+    parser.addoption(
+        "--extended",
+        action="store_true",
+        dest="extended",
+        default=False,
+        help="Run time-consuming 'extended' tests",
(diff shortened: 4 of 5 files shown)


AetherUnbound:
Thanks Rebecca! And good questions. This step is only to clear out the loading table, and is also specifically set up to clean up the table even if things go wrong. If we get to the drop step, I can't imagine a scenario where we'd care that it wasn't present when trying to delete it, since some error in creating the table would either be identified upstream or cause an error with copying the data into the main media type table. So I think we're clear here in all cases!

obulat (approved):
Interesting that a tiny change in code lead to so much improvement in tests :)

rwidom:
> Thanks Rebecca! And good questions. This step is only to clear out the loading table, and is also specifically set up to clean up the table even if things go wrong. If we get to the drop step, I can't imagine a scenario where we'd care that it wasn't present when trying to delete it, since some error in creating the table would either be identified upstream or cause an error with copying the data into the main media type table. So I think we're clear here in all cases!

Right, and I just realized that it will still report on whatever happened upstream in the reporting step. Yay! I'm so glad this worked out!

rwidom (approved):

krysal (approved):
> Interesting that a tiny change in code lead to so much improvement in tests :)

+1 to what Olga said! A cool compromise between speed and test coverage, learning a lot here of to test DAGs :)

stacimc (approved):
Just piling on to say it looks great to me as well! I always learn something looking at your testing code, @AetherUnbound 😄

AetherUnbound:
Thanks all! 😄 I keep needing to get myself out of the headspace of "this is a one line change" when the reality is that testing it is a much more complicated matter 😅 Glad folks are okay with this approach!

PR #635, opened 2022-07-28
Use the default provider categories during ingestion

### LINKED ISSUE #292: Move default categories from API to providers scripts
## Problem

The `category` field should be pre-populated for images as it is done for audio rather than assigned during data refresh [as performed currently](https://github.com/WordPress/openverse-api/blob/e0949d2641f20506338dc6a40250c24f5a8fa517/ingestion_server/ingestion_server/categorize.py). So default categories need to be transferred here to be used in each corresponding provider API script.

## Additional context

New sample data for images, containing the `category` column would be needed in the API. Reference to WordPress/openverse-api#357.

## Implementation

obulat:
## Fixes

Fixes #292 by @dhruvkb

## Description

This PR adds a default `category` to media items in the `MediaStore`'s `clean_media_metadata` method.
With this change, this is how `category` for an item is determined:
- if a _provider script_ can assign different categories, it should set the item category in the `get_record_data` method (or when calling `add_item` with old scripts).
- `MediaStore` uses the `category` value a provider script passes.
-  If the `category` is `None`, `MediaStore` tries to add the category that is default for provider, and is listed in `provider_details.py` module.

## Testing Instructions

Run one of the data collection DAGs until you see `Wrote 100 lines to the disc` in the `Pull data` task log. Stop the DAG. Wait a couple of seconds until the DAG finishes, and then check the data in the database (I use DataGrip connected to the localhost's DB). The `category` column should be filled. For example, for "clevelandmuseum", category should be set to `digitized_artwork`, and for "sciencemuseum", it should be set to `photograph`.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/loader/provider_details.py
+++ b/openverse_catalog/dags/common/loader/provider_details.py
@@ -121,7 +121,6 @@ class ImageCategory(Enum):
 
 # Default image category by source
 DEFAULT_IMAGE_CATEGORY = {
-    "flickr": ImageCategory.PHOTOGRAPH.value,
     "stocksnap": ImageCategory.PHOTOGRAPH.value,
     # Remains to be assigned
     "animaldiversity": ImageCategory.PHOTOGRAPH.value,
--- a/openverse_catalog/dags/common/storage/image.py
+++ b/openverse_catalog/dags/common/storage/image.py
@@ -101,11 +101,9 @@ class ImageStore(MediaStore):
                              given as an argument, the argument will
                              replace the one given in the dictionary.
         raw_tags:            List of tags associated with the image.
-        category:            The image category. `None` will be replaced
-                             with the default category for provider. If you
-                             don't want to use the default, pass the string
-                             `UNKNOWN`, which will be replaced with None
-                             in the database. @see Flickr script for example usage.
+        category:            The image category, defaults to the default
+                             category for provider from
+                             common/loader/provider_details.py.
         watermarked:         A boolean, or 't' or 'f' string; whether
                              the image has a noticeable watermark.
         source:              If different from the provider.  This might
--- a/openverse_catalog/dags/common/storage/image.py
+++ b/openverse_catalog/dags/common/storage/image.py
@@ -102,7 +102,7 @@ class ImageStore(MediaStore):
                              replace the one given in the dictionary.
         raw_tags:            List of tags associated with the image.
         category:            The image category, defaults to the default
-                             category for provider from
+                             category for the provider from
                              common/loader/provider_details.py.
         watermarked:         A boolean, or 't' or 'f' string; whether
                              the image has a noticeable watermark.
--- a/openverse_catalog/dags/common/storage/image.py
+++ b/openverse_catalog/dags/common/storage/image.py
@@ -100,9 +100,14 @@ class ImageStore(MediaStore):
                              in this dictionary, and `license_url` is
                              given as an argument, the argument will
                              replace the one given in the dictionary.
-        raw_tags:            List of tags associated with the image
-        watermarked:         A boolean, or 't' or 'f' string; whether or
-                             not the image has a noticeable watermark.
+        raw_tags:            List of tags associated with the image.
+        category:            The image category. `None` will be replaced
+                             with the default category for provider. If you
+                             don't want to use the default, pass the string
+                             `UNKNOWN`, which will be replaced with None
+                             in the database. @see Flickr script for example usage.
+        watermarked:         A boolean, or 't' or 'f' string; whether
+                             the image has a noticeable watermark.
         source:              If different from the provider.  This might
                              be the case when we get information from
                              some aggregation of images.  In this case,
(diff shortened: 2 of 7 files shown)


openverse_catalog/dags/common/storage/media.py
```diff
@@ -142,6 +144,9 @@ def clean_media_metadata(self, **media_data) -> Optional[dict]:
 
         media_data.pop("license_info", None)
         media_data["provider"] = self.provider
+        if not media_data["category"]:
+            if provider_category := prov.DEFAULT_IMAGE_CATEGORY.get(self.provider):
+                media_data["category"] = provider_category
```
AetherUnbound, line 149: Could this just be?
```suggestion
        media_data["category"] = media_date["category"] or prov.DEFAULT_IMAGE_CATEGORY.get(self.provider)
```

This will still set `media_data["category"]` to `None` if there is neither a category for that record or a default provider category.

openverse_catalog/dags/common/storage/image.py
```diff
@@ -100,9 +100,14 @@ def add_item(
...
                              replace the one given in the dictionary.
-        raw_tags:            List of tags associated with the image
-        watermarked:         A boolean, or 't' or 'f' string; whether or
-                             not the image has a noticeable watermark.
+        raw_tags:            List of tags associated with the image.
+        category:            The image category. `None` will be replaced
+                             with the default category for provider. If you
+                             don't want to use the default, pass the string
+                             `UNKNOWN`, which will be replaced with None
+                             in the database. @see Flickr script for example usage.
```
AetherUnbound, line 108: I'm not seeing any changes related to `UNKNOWN`, nor any existing logic for setting that to `NULL` in the database. Did I miss where this was occurring, or was it a change that was intended to be added?

AetherUnbound (commented):
Looks straightforward! Just one question.

openverse_catalog/dags/common/storage/image.py
obulat: Oh, I first tried to implement it this way because the default category for Flickr was set to photograph. Then I realized that it's cleaner to just remove the Flickr default  category (and set some images to photograph while leaving others as None in its script) to unify the different provider behaviors. But I forgot to remove the comment!

AetherUnbound (approved):
Looks great!

openverse_catalog/dags/common/loader/provider_details.py
```diff
@@ -121,7 +121,6 @@ class ImageCategory(Enum):
 
 # Default image category by source
 DEFAULT_IMAGE_CATEGORY = {
-    "flickr": ImageCategory.PHOTOGRAPH.value,
```
zackkrida, line 124: Did you mention why this is being removed? As I understand Flickr is exclusively photos (and some videos) and not illustrations or other media types.

openverse_catalog/dags/common/storage/image.py
```diff
@@ -100,9 +100,12 @@ def add_item(
                              in this dictionary, and `license_url` is
                              given as an argument, the argument will
                              replace the one given in the dictionary.
-        raw_tags:            List of tags associated with the image
-        watermarked:         A boolean, or 't' or 'f' string; whether or
-                             not the image has a noticeable watermark.
+        raw_tags:            List of tags associated with the image.
+        category:            The image category, defaults to the default
+                             category for provider from
```
zackkrida, line 105:
```suggestion
                             category for the provider from
```

grammatical nit

openverse_catalog/dags/common/storage/media.py
```diff
@@ -6,6 +6,7 @@
 
 from common.extensions import extract_filetype
 from common.licenses import is_valid_license_info
+from common.loader import provider_details as prov
```
zackkrida, line 9: Nit: I think I personally prefer the full `provider` here instead of `prov` because it's clearer, but it doesn't matter much.

zackkrida (approved):
LGTM! Some small nits (take them or leave them) but otherwise looks great. Successfully saw the `digitized_artwork` category on my 6k Cleveland Museum local DAG run.

openverse_catalog/dags/common/loader/provider_details.py
obulat: Flickr returns one of 3 `type` values, and we set the category as `photograph` only to the `photos`, not `screenshots` or `other`:
```python
def _get_category(image_data):
    """
    Flickr has three types:
        0 for photos
        1 for screenshots
        3 for other
    Treating everything different from photos as unknown.
    """
    if "content_type" in image_data and image_data["content_type"] == "0":
        return ImageCategory.PHOTOGRAPH.value
    return None
```
So, even before this PR, Flickr script was using `photograph` as the category for type 0. I don't know if we should investigate what types 1 and 3 are, and whether we can use them for categorizing as well.

openverse_catalog/dags/common/storage/media.py
obulat: I like the full `provider` here, but it's `prov` in all of the provider scripts, so I just used this for consistency (so that I don't have to change what I copy-pasted :) ) Do you think we should open an issue to replace `provider` with `prov`, @zackkrida?

AetherUnbound: +1 to using the full name rather than an alias 😄 I think an issue for that would be excellent! It'd be a solid "good first issue" as well

PR #513, opened 2022-05-18
Add DAG to report reported media pending review

stacimc:
## Fixes

Fixes #354

## Description

Adds a `report_pending_reported_images` DAG that runs weekly and for each media type:
* checks to see if there are any reported records in the `pending_review` status, meaning they require manual review through the Django Admin
* reports the number of _distinct_ records needing review to Slack, as well as a breakdown of the reasons for reporting (dmca, mature, other) and links to the admin page for that media

Screen Shot 2022-05-20 at 3 43 07 PM

## Additional Notes
* I verified that if a single image has multiple reports in `pending_review` associated with it, when you update _any_ of these records _all_ the other records associated to that image are updated as well. (This is why I think it's best to report _distinct_ images requiring review)
* When new media types are added, this will require you to add the name of the reports table in the configuration.

## Testing Instructions

This requires a connection to the API DB (rather than the Catalog), so you'll need to create the connection env variables. You can use the values set in `env.template` in this PR. Make sure you also have the API running locally. To test the slack messages you can create the Airflow variable `slack_message_override` and set it to `true`.

Run the `report_pending_reported_media` DAG locally. If you're running sample data, there should be no reports and you should see the `No records to report` message.

To create sample data to test the report count, you can report some images and audio (locally!) through the API. I went to `localhost:8000/v1/images` and grabbed the `id` for a couple of the listed images. For each I then navigated to `localhost:8000/v1/<id>/report` and POSTed a report. You can verify that these went through by checking out `http://localhost:8000/admin/api/imagereport/?` Likewise for Audio. Report a mix of different reasons (some 'mature', some 'other', etc).

Run the DAG again and make sure the correct numbers are reported. I also played around with manually reviewing some of the media reports (updating their status to something other than `pending_review` in the Admin) and verifying they were no longer included in the count; and also reporting the same records multiple times and verifying that it is only counted once in the output.

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [ ] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [ ] My commit messages follow [best practices][best_practices].
- [ ] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [ ] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

[...]

PR #636, opened 2022-07-29
Data refresh record difference reporting

### LINKED ISSUE #466: Report media counts before and after data refresh
## Description

It would be awesome to have a sense of how many new records were added during a given media's data refresh. I think this could be accomplished as pre and post tasks in the data refresh monitoring DAGs, and reported in Slack as the final step. We could also perform this step in the ingestion server, although it would be easier to manage the notification logic within Airflow.

## Implementation

AetherUnbound:
## Fixes

Fixes #466 by @AetherUnbound

## Description

This PR adds a few new steps to the data refresh DAGs which gather the record counts for a provided media table from the API before and after the data refresh, and then report the difference to Slack. 

The easiest way of adding this to the DAG was to include a few `PostgresOperator` steps which would query the record count before and after the refresh. Unfortunately, the `PostgresOperator` doesn't push the return value to XComs by default. In order to facilitate that behavior here (and in future DAGs), I added a `PostgresResultOperator`. This operator is identical to the `PostgresOperator`, save that it allows a `handler` to be passed in, which in-turn is passed down to the `PostgresHook` when `PostgresHook::run` is called. It also returns the results of the provided handler function, so it can be referenced as an XCom downstream. I don't love that we have essentially vendored the `PostgresOperator` with a slightly tweaked interface, but I don't presume there will be many changes that we need to incorporate going forward (seeing as there were [only half a dozen changes to the file in as many months recently](https://github.com/apache/airflow/commits/main/airflow/providers/postgres/operators/postgres.py)). Nevertheless, I added some tests for this operator to ensure it's functioning as expected.

It's worth noting that we're using a _row count estimate_ here rather than the exact row count. Discussion around this can be found on our Make WordPress Slack: https://wordpress.slack.com/archives/C02012JB00N/p1659039686501669. The TL;DR is that an exact count is extremely slow and the estimation method we're using is pretty darn accurate.

:warning: **NOTE**: We will need to add the `postgres_openledger_api` connection to production Airflow before this is run. It was introduced in #513 but it doesn't appear to be listed under the connections in Airflow yet.

### Screenshots

**DAG Definition** (with new steps)
Screenshot_2022-07-28_17-56-30

**Sample Slack message**
Screenshot_2022-07-28_18-18-20

## Testing Instructions

Testing for this is a bit complex, so let me know if any of these steps fail or require a bit more explanation:

0. Prerequisite: set up your local Airflow instance so it can communicate with a local API stack. @stacimc's instructions in #397 (under "Testing Instructions") are useful for these steps, with the caveat that `AIRFLOW_CONN_DATA_REFRESH` can now just be `http://172.17.0.1:8001` or `http://host.docker.internal:8001` for linux & Mac respectively.
1. In the API project directory, run the following (this will set up an actual difference in records to be reported)
    1. `just recreate`
    2. `just init`
    3. `just exec "upstream_db psql -U deploy openledger -c \"DELETE FROM image_view WHERE provider = 'flickr'\""`
2. Trigger `image_data_refresh` DAG (either by enabling it in the UI or triggering a run)
3. When the DAG gets to the `wait_for_ingest_upstream` step, you may notice it go from `up_for_reschedule` to `up_for_retry`. This appears to be a result of https://github.com/apache/airflow/issues/23411, which we'll try to address in another PR. Wait a few seconds, then clear the `wait_for_ingest_upstream` step. This will allow the DAG to proceed as normal
4. Check the `report_record_counts` step and see that it shows a similar message to the screenshot!

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/operators/postgres_result.py
+++ b/openverse_catalog/dags/common/operators/postgres_result.py
@@ -1,0 +1,81 @@
+from typing import TYPE_CHECKING, Callable, Iterable, List, Mapping, Optional, Union
+
+from airflow.providers.postgres.hooks.postgres import PostgresHook
+from airflow.providers.postgres.operators.postgres import PostgresOperator
+from psycopg2.sql import SQL, Identifier
+
+
+if TYPE_CHECKING:
+    from airflow.utils.context import Context
+
+
+class PostgresResultOperator(PostgresOperator):
+    """
+    Override for the PostgresOperator which is functionally identical except that a
+    handler can be specified for the PostgresHook. The value(s) accumulated from the
+    run function will then be pushed as an XCom.
+    """
+
+    def __init__(
+        self,
+        *,
+        sql: Union[str, List[str]],
+        handler: Callable,
+        postgres_conn_id: str = "postgres_default",
+        autocommit: bool = False,
+        parameters: Optional[Union[Mapping, Iterable]] = None,
+        database: Optional[str] = None,
+        runtime_parameters: Optional[Mapping] = None,
+        **kwargs,
+    ) -> None:
+        super().__init__(
+            sql=sql,
+            postgres_conn_id=postgres_conn_id,
+            autocommit=autocommit,
+            parameters=parameters,
+            database=database,
+            runtime_parameters=runtime_parameters,
+            **kwargs,
+        )
+        self.handler = handler
+
+    def execute(self, context: "Context"):
+        """
+        This almost exactly mirrors PostgresOperator::execute, except that allows
+        passing a handler into the hook.
+        """
+        self.hook = PostgresHook(
+            postgres_conn_id=self.postgres_conn_id, schema=self.database
+        )
+        if self.runtime_parameters:
+            final_sql = []
+            sql_param = {}
+            for param in self.runtime_parameters:
+                set_param_sql = f"SET {{}} TO %({param})s;"
+                dynamic_sql = SQL(set_param_sql).format(Identifier(f"{param}"))
+                final_sql.append(dynamic_sql)
+            for param, val in self.runtime_parameters.items():
+                sql_param.update({f"{param}": f"{val}"})
+            if self.parameters:
+                sql_param.update(self.parameters)
+            if isinstance(self.sql, str):
+                final_sql.append(SQL(self.sql))
+            else:
+                final_sql.extend(list(map(SQL, self.sql)))
+            results = self.hook.run(
+                final_sql,
+                self.autocommit,
+                parameters=sql_param,
+                handler=self.handler,
+            )
+        else:
+            results = self.hook.run(
+                self.sql,
+                self.autocommit,
+                parameters=self.parameters,
+                handler=self.handler,
+            )
+        for output in self.hook.conn.notices:
+            self.log.info(output)
+
+        return results
--- a/openverse_catalog/dags/common/operators/postgres_result.py
+++ b/openverse_catalog/dags/common/operators/postgres_result.py
@@ -41,7 +41,7 @@ class PostgresResultOperator(PostgresOperator):
 
     def execute(self, context: "Context"):
         """
-        This almost exactly mirrors PostgresOperator::execute, except that allows
+        This almost exactly mirrors PostgresOperator::execute, except that it allows
         passing a handler into the hook.
         """
         self.hook = PostgresHook(
--- a/openverse_catalog/dags/data_refresh/dag_factory.py
+++ b/openverse_catalog/dags/data_refresh/dag_factory.py
@@ -40,7 +40,11 @@ from airflow.operators.python import BranchPythonOperator, PythonOperator
 from airflow.settings import SASession
 from airflow.utils.session import provide_session
 from airflow.utils.state import State
-from common.constants import DAG_DEFAULT_ARGS, XCOM_PULL_TEMPLATE
+from common.constants import (
+    DAG_DEFAULT_ARGS,
+    OPENLEDGER_API_CONN_ID,
+    XCOM_PULL_TEMPLATE,
+)
 from common.operators.postgres_result import PostgresResultOperator
 from data_refresh.data_refresh_task_factory import create_data_refresh_task_group
 from data_refresh.data_refresh_types import DATA_REFRESH_CONFIGS, DataRefresh
@@ -66,7 +70,6 @@ REFRESH_POPULARITY_METRICS_TASK_ID = (
     f"{REFRESH_POPULARITY_METRICS_GROUP_ID}"
     f".{UPDATE_MEDIA_POPULARITY_METRICS_TASK_ID}"
 )
-API_DB_CONN_ID = "postgres_openledger_api"
 
 
 def _single_value(cursor):
@@ -188,7 +191,7 @@ def create_data_refresh_dag(data_refresh: DataRefresh, external_dag_ids: Sequenc
         # Get the current number of records in the target API table
         before_record_count = PostgresResultOperator(
             task_id="get_before_record_count",
-            postgres_conn_id=API_DB_CONN_ID,
+            postgres_conn_id=OPENLEDGER_API_CONN_ID,
             sql=count_sql,
             handler=_single_value,
         )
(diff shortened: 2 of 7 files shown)


openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -64,6 +66,15 @@
     f"{REFRESH_POPULARITY_METRICS_GROUP_ID}"
     f".{UPDATE_MEDIA_POPULARITY_METRICS_TASK_ID}"
 )
+API_DB_CONN_ID = "postgres_openledger_api"
```
stacimc, line 69: Should we get this from the `OPENLEDGER_API_CONN_ID` environment variable, and maybe add it to https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/constants.py?

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -177,9 +210,34 @@ def create_data_refresh_dag(data_refresh: DataRefresh, external_dag_ids: Sequenc
...
+            task_id="report_record_counts",
+            python_callable=report_record_difference,
+            op_kwargs={
+                "before": XCOM_PULL_TEMPLATE.format(
+                    before_record_count.task_id, "return_value"
+                ),
+                "after": XCOM_PULL_TEMPLATE.format(
+                    after_record_count.task_id, "return_value"
+                ),
+                "media_type": data_refresh.media_type,
```
stacimc, line 232: #640 is going to make this look so clean! Excellent idea 😄

openverse_catalog/dags/data_refresh/record_reporting.py
```diff
@@ -0,0 +1,28 @@
...
+    # ',' - number is comma separated
+    # '.' - number is a float
+    message = f"""
+_Note: All values are row estimates and are not (but nearly) exact_
+*Record count difference for `{media_type}`*: {before:,} → {after:,}
+*Change*: {count_diff:+,} ({percent_diff:+}% Δ)
+"""
+    log.info(message)
+    if slack.should_send_message():
+        slack.send_message(text=message, username="Data refresh record difference")
```
stacimc, line 25: The slack utility already takes care of checking `should_send_message` [here](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/slack.py#L225), so we shouldn't need to check.

I'm also updating `send_message` in #643 to automatically[ log the message](https://github.com/WordPress/openverse-catalog/pull/643/files#diff-4284b9bf552cdfb3dbbfa19c55fcb442f98d3933939dc34c25790834988c38edR225) (even if `should_send_message` fails), since I think we want that by default -- so we should be okay to remove the `log.info` here.

stacimc (commented):
I am *so* excited about this addition! We've really been missing quick insight into this information, and it'll be especially rewarding to have as the data refreshes get turned back on after a hiatus 😄 I've run out of time to manually test but will pick back up with reviewing tomorrow!

> I don't love that we have essentially vendored the PostgresOperator with a slightly tweaked interface, but I don't presume there will be many changes that we need to incorporate going forward.

Given this ^, can you expand on the benefits of using the new Operator instead of using a `PostgresHook` in a `PythonOperator`? Do you think this could be used in some of the other places we're currently interacting with the DB? 😮

openverse_catalog/dags/common/operators/postgres_result.py
```diff
@@ -0,0 +1,81 @@
+from typing import TYPE_CHECKING, Callable, Iterable, List, Mapping, Optional, Union
+
+from airflow.providers.postgres.hooks.postgres import PostgresHook
+from airflow.providers.postgres.operators.postgres import PostgresOperator
+from psycopg2.sql import SQL, Identifier
+
+
+if TYPE_CHECKING:
```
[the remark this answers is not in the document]
AetherUnbound: Copied straight from the `PostgresOperator`! 😄 
https://github.dev/apache/airflow/blob/42ea000401c50d8b47e791b746859854420ac0f6/airflow/providers/postgres/operators/postgres.py#L27

openverse_catalog/dags/data_refresh/dag_factory.py
AetherUnbound: Fully support adding this to constants!! That's a great idea.

openverse_catalog/dags/data_refresh/record_reporting.py
AetherUnbound: Oh good! That's perfect, I'll remove all this then, thank you 😄

AetherUnbound:
> Given this ^, can you expand on the benefits of using the new Operator instead of using a `PostgresHook` in a `PythonOperator`? Do you think this could be used in some of the other places we're currently interacting with the DB? 😮 

Sure! I'm not sure if it could be used in place _now_, but I believe that the "run this query and post the results to XComs" paradigm is something we will definitely use in the future. Since this operator also takes whatever callable we want to hand into the postgres hook, it's robust to more than just `SELECT COUNT` and the ilk. It also prevents us from having to have multiple functions down the line that are of the form:

```python

def get_pg_results(postgres_conn_id: str, sql: str):
    hook = PostgresHook(postgres_conn_id)
    results = hook.run(sql, handler=lambda c: ...)
    return results

...

get_results = PythonOperator(
    task_id=...,
    python_callable=get_pg_results,
    op_kwargs={
        "postgres_conn_id": PG_CONN_ID,
        "sql": ...,
    }
)
```

And instead simply use:

```python

get_results = PostgresResults(
    task_id=...,
    postgres_conn_id=PG_CONN_ID,
    sql=...,
    handler=lambda c: ...
)
```

Another advantage of the latter is that we only have to write the tests against PostgresResults to ensure it behaves as expected, rather than writing tests against every `get_pg_results`-like function. If we needed to pass in a more complicated handler, we could have that be a defined function (rather than an in-line lambda) and test against that instead!

tests/dags/common/operators/test_postgres_result.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+@pytest.fixture(autouse=True)
+def clean_db():
+    with create_session() as session:
+        session.query(DagRun).filter(DagRun.dag_id == TEST_DAG_ID).delete()
+        session.query(TaskInstance).filter(TaskInstance.dag_id == TEST_DAG_ID).delete()
+        yield
+        session.query(DagRun).filter(DagRun.dag_id == TEST_DAG_ID).delete()
+        session.query(TaskInstance).filter(TaskInstance.dag_id == TEST_DAG_ID).delete()
```
stacimc, line 33: Why does this `yield` and re-run the clean up?

openverse_catalog/dags/data_refresh/dag_factory.py
```diff
@@ -66,6 +72,14 @@
 )
 
 
+def _single_value(cursor):
```
stacimc, line 75: Not necessarily now, but this is so generic I'd imagine we could move it out to a util file somewhere. Maybe when we have more of these `PostgresResultOperator`s 🎉

stacimc (approved):
I was able to test this manually after a `just recreate` in my local Catalog as well, and LGTM! I'm sold on the reusability of the `PostgresResultOperator` -- in fact if anything I suspect this will inspire future work 😄

I appreciate linking the conversation about the row estimation technique as well as documenting it in the code itself, TIL. It's great we're able to get such an accurate estimate! That definitely seems sufficient for our purposes provided it's documented as an estimate in reports, like you say.

Super excited for this one! ✨

obulat (approved):
Really nice additions with PostgresResult and single_value!

tests/dags/common/operators/test_postgres_result.py
AetherUnbound: This might be overkill 😅 I wanted to catch the case where maybe a previous cleanup didn't run successfully or something, so it cleans up both before _and_ after the test. Do you think that's too much? Should it just clean up after the test is run?

openverse_catalog/dags/data_refresh/dag_factory.py
AetherUnbound: Totally agree! I think there might be a few of them, like this single result, row count, `fetchall`, etc.

AetherUnbound:
`postgres_openledger_api` connection has been added! Merging 🚀

PR #643, opened 2022-08-01
Omit DAGs that are known to fail from alerts

### LINKED ISSUE #438: Flag DAGs that are expected to fail to omit them from alerts
## Problem

We have a couple provider DAGs running that we know have outstanding issues that will cause errors, but still allow for partial data to be processed -- for example, Wikimedia Commons is currently throwing a TypeError in its `pull_data` task (tracked in #436). We know the DAG will fail and have an issue set up to fix it, but in the meantime we'd like to keep the DAG turned on so we can continue getting at least some data.

The problem is that we still get Slack alerts. Folks who are unaware of the open issue may spend time investigating the error. It also just clutters up the alerts channel, particularly for daily DAGs.

## Description

Per a suggestion by @AetherUnbound 💯:

> it would be cool to use Variables to, on the fly, set DAGs that we know will fail. The slack alert could check for this as part of it's alerting logic. That'd allow us to set a known failing DAG in the UI but still let whatever records we catch get processed
 

## Implementation

stacimc:
## Fixes

Fixes #438

## Description

Sometimes DAGs with known errors are left active in order to continue consuming partial data. In this case an issue should be opened to address the error, but we may want to temporarily turn off Slack alerts for the DAG to prevent (1) cluttering the channel and (2) causing time to be wasted by team members investigating already tracked issues.

This PR introduces a new Airflow variable, `silenced_slack_alerts`, used to configure this. Example configuration:
Screen Shot 2022-08-02 at 5 41 05 PM

Each key is the `dag_id` of a DAG who should have Slack alerts turned off. The value is the URL of a GitHub issue tracking the known failure.

This PR:
* Adds a check to `slack#send_alert` to skip sending the notification if the DAG is in this dict
* Adds a new `check_silenced_dags` DAG which runs `@weekly` and verifies that for each silenced dag, the associated github issue is still open. If the issue has been closed, it sends a slack alert reminding developers to turn alerts back on.

Screen Shot 2022-08-01 at 2 37 35 PM

I can split the new DAG out into a separate PR if folks would prefer.

## Testing Instructions

* Create an Airflow variable named `silenced_slack_alerts` in the Admin > Variables UI. Silence alerts for a few DAGs, making sure to link at least one to an open issue, and at least one to a closed issue. For my tests I used this configuration:
```
{
  # Turns off alerts for Finnish Museums, linking to a closed issue
  "finnish_museums_workflow": "https://github.com/WordPress/openverse-catalog/issues/229",
  # Turns off alerts for Cleveland Museum, linking to an open issue
  "cleveland_museum_workflow": "https://github.com/WordPress/openverse-catalog/issues/438"
}
```
* For the DAGs you've configured above, force an error to be thrown. I did this by editing the code to manually raise an Exception somewhere in the `pull_data` task.
* Make sure you have `slack_message_override` set to `true` in your Airflow variable, to force slack messages to send in your local environment.
* For each of the configured DAGs, run them. You should see the error, but no Slack message should be sent. In the logs of the failed task, you should see something like `Skipping Slack alert for finnish_museums_workflow`. This should be true _regardless of whether the associated issue is open or closed._

Now test the `check_silenced_dags` DAG:
* Run the DAG. You should see a Slack alert like the screenshot above, listing all of the DAGs linked to closed github issues.
* Edit the `silenced_slack_alerts` dict to remove any DAGs associated to closed github issues, and run the DAG again. You should see no Slack alert.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


tests/dags/maintenance/test_check_silenced_alerts.py
```diff
@@ -0,0 +1,136 @@
...
+            ("WordPress", "openverse-catalog", "100"),
+        ),
+        pytest.param(
+            "openverse/issues/10",
+            None,
+            marks=pytest.mark.raises(
+                exception=AirflowException,
+                match="Issue url openverse/issues/10 could not be parsed.",
+            ),
+        ),
```
sarayourfriend, line 131: TIL! I didn't know you could parameterize raised exceptions. I think I tried this in another project without `pytest.param` and came up with a really wacky solution and then ended up just doing it inline in the test anyway :sweat_smile:

tests/dags/maintenance/test_check_silenced_alerts.py
```diff
@@ -0,0 +1,136 @@
...
+def test_check_configuration(silenced_dags, dags_to_reenable, should_send_alert):
+    with mock.patch(
+        "maintenance.check_silenced_alerts.check_silenced_alerts.Variable",
+        return_value=silenced_dags,
+    ), mock.patch(
+        "maintenance.check_silenced_alerts.check_silenced_alerts.get_dags_with_closed_issues",
+        return_value=dags_to_reenable,
+    ) as get_dags_with_closed_issues_mock, mock.patch(
+        "maintenance.check_silenced_alerts.check_silenced_alerts.send_alert"
+    ) as send_alert_mock:
```
sarayourfriend, line 50: nit: If there's any way to reasonably split this into separate `mock.patch` calls into separate expressions without introducing the nesting that's avoided here, that would be appreciated.

One option, if you're okay with not using a context manager and having the mock last the lifetime of the test (which should be fine with the current implementation) is pytest's [`monkeypatch`](https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html) fixture. It doesn't require a context manager so can be split into multiple lines without having to put them into a single expression.

sarayourfriend (commented):
This LGTM. I still need to test this locally, but one question that comes to mind immediately is whether DAGs that need to silence errors could export a predicate that is passed the error and decides whether to alert or not. I'm not sure how this would work with the issue checking DAG (which is cool, and a nice accountability tool) but it would allow silencing _some_ errors from Slack without silencing _all_ errors. Silencing all errors raised in a DAG seems like it might work against the goal of stabilizing DAGs and making DAG alerts more meaningful.

tests/dags/maintenance/test_check_silenced_alerts.py
AetherUnbound: It's definitely a useful thing to have! Unfortunately it requires [an extra plugin](https://pypi.org/project/pytest-raises/) in order to do, but once I found that I personally couldn't live without it!

openverse_catalog/dags/maintenance/check_silenced_dags/check_silenced_dags.py
```diff
@@ -0,0 +1,57 @@
...
+from airflow.exceptions import AirflowException
+from airflow.models import Variable
+from common.github import GitHubAPI
+from common.slack import send_alert
+
+
+logger = logging.getLogger(__name__)
+
+
+def check_configuration(github_pat: str, airflow_variable: str):
```
AetherUnbound, line 13: Nit: Would you be comfortable reversing the order of these functions? I tend to read from top to bottom both when doing a review and accessing the code, and so having the order of composition match the order in the file can be really helpful for that. I know this repo isn't consistent about this, but it would be nice to do that ordering for new modules!

openverse_catalog/dags/maintenance/check_silenced_dags/check_silenced_dags.py
```diff
@@ -0,0 +1,57 @@
...
+    silenced_dags = Variable.get(airflow_variable, {}, deserialize_json=True)
+    dags_to_reenable = get_dags_with_closed_issues(github_pat, silenced_dags)
+
+    if not dags_to_reenable:
+        logger.info(
+            "All DAGs configured to silence messages have work still in progress."
+            " No configuration updates needed."
+        )
+        return
```
AetherUnbound, line 22: I think raising an `AirflowSkipException` here might be best, that way we have a clear view from the grid or calendar views which runs produced slack messages and which did not!

openverse_catalog/dags/maintenance/check_silenced_dags/check_silenced_dags_dag.py
```diff
@@ -0,0 +1,60 @@
...
+from airflow.operators.python import PythonOperator
+from common.constants import DAG_DEFAULT_ARGS
+from maintenance.check_silenced_dags import check_silenced_dags
+
+
+logger = logging.getLogger(__name__)
+
+
+DAG_ID = "check_silenced_dags"
+MAX_ACTIVE_TASKS = 1
```
AetherUnbound, line 32: Since this covers both max active settings, perhaps this should be `MAX_ACTIVE`?

openverse_catalog/dags/maintenance/check_silenced_dags/check_silenced_dags_dag.py
```diff
@@ -0,0 +1,60 @@
...
+GITHUB_PAT = Variable.get("GITHUB_API_KEY", default_var="not_set")
+
+
+dag = DAG(
```
AetherUnbound, line 36: I know I keep saying this so apologies if this comes across as pushy, but we could definitely use the TaskFlow API for this DAG! 😄

tests/dags/maintenance/test_check_silenced_alerts.py
AetherUnbound: It's also worth noting that (thanks to python 3.10) this block can be wrapped in parenthesis and `black` will format it a bit better:
```python
    with (
        mock.patch(
            "maintenance.check_silenced_dags.check_silenced_dags.Variable",
            return_value=silenced_dags,
        ),
        mock.patch(
            "maintenance.check_silenced_dags.check_silenced_dags.get_dags_with_closed_issues",
            return_value=dags_to_reenable,
        ) as get_dags_with_closed_issues_mock,
        mock.patch(
            "maintenance.check_silenced_dags.check_silenced_dags.send_alert"
        ) as send_alert_mock,
    ):
```

openverse_catalog/dags/maintenance/check_silenced_dags/check_silenced_dags.py
stacimc: Oh I love this! Making a note of that for the future as well 😄

AetherUnbound (approved):
This is fantastic!! I love the extra DAG as well, that's a great check for us to have for when we use this feature. The bonus of using GitHub issue links as a check is so smart.

One note though, I had to change my variable to `cleveland_museum_workflow` (from `cleveland_museums_workflow`), otherwise the worfklow actually did report the error. Other than that, 
I was able to test this locally and everything worked as expected! 🚀 🤖

tests/dags/maintenance/test_check_silenced_dags.py
```diff
@@ -14,8 +14,13 @@
 @pytest.mark.parametrize(
     "silenced_dags, dags_to_reenable, should_send_alert",
     (
-        # No Dags to reenable, don't alert
-        ({}, [], False),
+        # No Dags to reenable, task should skip
```
AetherUnbound, line 17: Do you mind adding one more case here where there are DAGs that are silenced, should be silenced, but don't alert and raise a skip?

stacimc: I don't think that would be meaningful in this test because `get_dags_to_reenable` is mocked. If there are silenced DAGs but they should be silenced, `get_dags_to_reenable` will return an empty list, which is covered by this first test case. `get_dags_to_reenable` is tested separately to make sure it does not return silenced DAGs that _should_ be silenced.

AetherUnbound: Got it, sounds good!

stacimc:
> one question that comes to mind immediately is whether DAGs that need to silence errors could export a predicate that is passed the error and decides whether to alert or not

Yeah, I definitely agree that we'll need to add this feature long term. We can expand the configuration variable and make it possible to support the issue checking DAG as well. I'd prefer to add it as a fast follow; otherwise I'll get to this as soon as I can.

sarayourfriend:
That sounds good. I'll finish reviewing this PR this afternoon.

openverse_catalog/dags/maintenance/check_silenced_dags/check_silenced_dags.py
```diff
@@ -0,0 +1,56 @@
...
+
+    message = (
+        "The following DAGs have Slack messages silenced, but the associated issue is"
+        f" closed. Please remove them from the `{airflow_variable}` Airflow variable"
+        " or assign a new issue."
+    )
+    for (dag, issue) in dags_to_reenable:
+        message += f"\n  - <{issue}|{dag}>"
+    send_alert(message, username="Silenced DAG Check", unfurl_links=False)
+    return message
```
sarayourfriend, line 56: It'd be nice if this created a GitHub issue or something that could actually be assigned and tracked or have the maintainers pinged. Or maybe even just left a comment on the issue itself like "Please un-silence the DAG errors". Or maybe both, a new issue and a ping on the old, with the new issue just tracking the work of actually updating the prod configuration.

Just worried a slack ping could easily get lost (especially if lots of people are on vacation or distracted by something else, for example) in a way that a GitHub issue won't, as it acts more like a formal "todo" item.

sarayourfriend (approved):
LGTM! Just one suggestion for how to improve the accountability around this process. This is great though. I love seeing our different systems come together like this.

obulat (approved):
This is a great improvement for our Slack channels :)
It would be nice to have a more visible record of silenced DAGs, somewhere in a GitHub issue or a file in the repository.

openverse_catalog/dags/maintenance/check_silenced_dags/check_silenced_dags.py
stacimc: Oh I _love_ this idea! I'm going to create a follow-up issue and link back, this would be fantastic.

PR #645, opened 2022-08-02
Tighten exception handling, always flush buffer

### LINKED ISSUE #366: Write current buffer to disk on failure

AetherUnbound:
- Use more specific exceptions when catching errors
- Use and catch a specific retries-exceeded exception
- Return total values committed by all stores
- Clean up comments
- Add unittest for failure mode

## Fixes

Fixes #366 by @AetherUnbound

## Description

This PR ensures that the results within the media store always get written to disk, including when Airflow issues a `SIGTERM` to the worker tasks. There are a lot of small changes in this PR, but they all fall into two categories:

1. Making exception handling _where we're expected to continue execution_ more specific
2. Wrap the `ingest_records` function of the provider data ingester steps in a `try`/`finally` to ensure that records in the stores' buffers will be written to disk.

### Tighter exception handling

One thing I notice when trying to test this out was that when I marked an ingestion task as failed in Airflow, there were some cases where the `SIGTERM` signal would just be ignored and execution would continue. In Airflow it would look like the task was failed (or succeeded), but the job would actually _continue running_ because it just disregarded the exception that Airflow raised as a way of terminating the job. Fortunately the job wouldn't run too much longer because Airflow's default task kill timeout is 60 seconds:

https://github.com/apache/airflow/blob/007b1920ddcee1d78f871d039a6ed8f4d0d4089d/airflow/config_templates/default_airflow.cfg#L131-L133

Still though, this could mean the TSV file was still open/being edited by the time it was uploaded to S3.

The root cause of this ended up being numerous cases where we were using `except Exception as e` and then continuing with the processing. A prime example of this is the requester functions - the _intention_ was to handle any exception case that might come up as part of a request, but the `except Exception` is so permissive that it catches everything (including things like `KeybordIterrupt` and `AirflowException`).

I tried to rein in which exceptions were being caught to a specific subset relevant for each occurrence of `except Exception`. I also had to alter some tests that were relying on the behavior of these catch-alls. One case of note was the `requester.py` retries exception case. The provider data ingester was relying on that exception being caught, but `Exception` is too generic since we want to let others bubble up appropriately. So I opted to add a custom `RetriesExceeded` exception there.

### Always flush the buffer

Once I had identified and resolved the above, the buffer flushing issue was way easier than I thought it'd be! Thanks to how nice Airflow is about stopping (it just raises an `AirflowException` as mentioned earlier), we can just wrap the `ingest_records` function of the provider data ingester in a try/finally which ensures that the stores are always committed before execution completes. I also added a test to ensure this behaves as expected.

I chose at this time _not_ to alter the behavior of the legacy provider scripts. We're hoping to tackle the rest of them as part of the v1.3.2 milestone, so I didn't want to put too much effort into supporting them here. For the future though, we could also apply the `try`/`finally` logic to the `pull_media_wrapper` function to make sure we get duration data even on failed runs! Now that we know Airflow will ask us nicely to stop :grin:

## Testing Instructions

`just recreate && just test`

You can also try making the following edits to the Cleveland Museum provider, run the DAG, and make sure that all 147 records get written to the disk:

```bash
$ git diff openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py 
diff --git a/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py b/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
index 32f640f2..9c571535 100644
--- a/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
@@ -16,6 +16,7 @@ class ClevelandDataIngester(ProviderDataIngester):
     endpoint = "http://openaccess-api.clevelandart.org/api/artworks/"
     batch_limit = 1000
     delay = 5
+    foo = 0
 
     def get_next_query_params(self, prev_query_params, **kwargs):
         if not prev_query_params:
@@ -38,6 +39,10 @@ class ClevelandDataIngester(ProviderDataIngester):
         return None
 
     def get_record_data(self, data):
+        self.foo += 1
+        if self.foo > 147:
+            raise KeyboardInterrupt("Whoops!")
+
         license_ = data.get("share_license_status", "").lower()
         if license_ != "cc0":
             logger.error("Wrong license image")
```

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/requester.py
+++ b/openverse_catalog/dags/common/requester.py
@@ -3,6 +3,8 @@ import time
 
 import oauth2
 import requests
+from airflow.exceptions import AirflowException
+from requests.exceptions import JSONDecodeError
 
 
 # pytest_socket will not be available in production, so we must create a shim for
@@ -65,6 +67,12 @@ class DelayedRequester:
             # This exception will only be raised during testing, and it *must*
             # be re-raised and bubbled up the stack
             raise
+        except (AirflowException, KeyboardInterrupt):
+            # These exceptions are raised by task managers and should be respected &
+            # re-raised. Airflow runs its tasks in a separate thread, so if this
+            # exception is received, typically it means that the task has been
+            # sent a SIGTERM, which means that the task should be stopped.
+            raise
         except Exception as e:
             logger.error(f"Error with the request for URL: {url}.")
             logger.info(f"{type(e).__name__}: {e}")
@@ -89,7 +97,7 @@ class DelayedRequester:
         if response is not None and response.status_code == 200:
             try:
                 response_json = response.json()
-            except Exception as e:
+            except JSONDecodeError as e:
                 logger.warning(f"Could not get response_json.\n{e}")
                 response_json = None
 
--- a/openverse_catalog/dags/common/requester.py
+++ b/openverse_catalog/dags/common/requester.py
@@ -20,6 +20,14 @@ except ImportError:
 logger = logging.getLogger(__name__)
 
 
+class RetriesExceeded(Exception):
+    """
+    Custom exception for when the number of allowed retries has been exceeded.
+    """
+
+    pass
+
+
 class DelayedRequester:
     """
     Provides a method `get` that is a wrapper around `get` from the
@@ -91,7 +99,7 @@ class DelayedRequester:
 
         if retries < 0:
             logger.error("No retries remaining.  Failure.")
-            raise Exception("Retries exceeded")
+            raise RetriesExceeded("Retries exceeded")
 
         response = self.get(endpoint, params=query_params, **kwargs)
         if response is not None and response.status_code == 200:
--- a/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
@@ -121,8 +121,8 @@ class ProviderDataIngester(ABC):
                     should_continue = False
         finally:
             logger.info("======Something went wrong!! But that's OK I'm exiting======")
-
-        self.commit_records()
+            total = self.commit_records()
+            logger.info(f"Committed {total} records")
 
     @abstractmethod
     def get_next_query_params(
@@ -259,6 +259,8 @@ class ProviderDataIngester(ABC):
         """
         pass
 
-    def commit_records(self):
+    def commit_records(self) -> int:
+        total = 0
         for store in self.media_stores.values():
-            store.commit()
+            total += store.commit()
+        return total
--- a/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
@@ -3,7 +3,7 @@ from abc import ABC, abstractmethod
 from typing import Dict, List, Optional, Tuple
 
 from airflow.models import Variable
-from common.requester import DelayedRequester
+from common.requester import DelayedRequester, RetriesExceeded
 from common.storage.media import MediaStore
 from common.storage.util import get_media_store_class
 from requests.exceptions import JSONDecodeError, RequestException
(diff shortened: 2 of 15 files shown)


sarayourfriend (approved):
LGTM. The specific exceptions are a welcome change.

openverse_catalog/dags/common/requester.py
```diff
@@ -65,6 +75,12 @@ def get(self, url, params=None, **kwargs):
             # This exception will only be raised during testing, and it *must*
             # be re-raised and bubbled up the stack
             raise
+        except (AirflowException, KeyboardInterrupt):
+            # These exceptions are raised by task managers and should be respected &
+            # re-raised. Airflow runs its tasks in a separate thread, so if this
+            # exception is received, typically it means that the task has been
+            # sent a SIGTERM, which means that the task should be stopped.
+            raise
```
obulat, line 83: TIL that you can re-raise the last error with simple `raise` statement with no arguments.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -100,23 +101,25 @@ def ingest_records(self, **kwargs) -> None:
...
+                    logger.info("Batch complete.")
+                    should_continue = False
 
-        self.commit_records()
+                if self.limit and record_count >= self.limit:
+                    logger.info(f"Ingestion limit of {self.limit} has been reached.")
+                    should_continue = False
+        finally:
+            total = self.commit_records()
+            logger.info(f"Committed {total} records")
```
obulat, line 122: Interesting that you don't need an `except` clause, and can handle the saved exception after the `finally` clause.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -163,7 +166,13 @@ def get_batch(self, query_params: Dict) -> Tuple[Optional[List], bool]:
...
 
-        except Exception as e:
+        except (
+            RequestException,
+            RetriesExceeded,
+            JSONDecodeError,
+            ValueError,
+            TypeError,
+        ) as e:
             logger.error(f"Error due to {e}")
```
obulat, line 176: Would something like "Error getting next query parameters due to {e}" be clearer?

obulat (approved):
Really nice cleanup for killing the unkillable :)

I agree that we don't need to add too much error-handling to the provider scripts because re-writing them using ProviderDataIngester is scheduled for the next milestone.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
sarayourfriend: This kind of thing I think is usually clearer (at least in the intention communicated by the code) when a context manager is used. I considered suggesting it but it would require factoring code out into another class to act as the manager so it didn't seem worth it. `try/finally` [also works in JavaScript](https://github.com/WordPress/openverse-frontend/blob/b863e3e4b88eb667a87a2c620abc688ab29838eb/src/plugins/api-token.server.ts#L146-L162) and other languages (at least the ones I've used) that have `try/catch/finally` pattern.

openverse_catalog/dags/common/requester.py
AetherUnbound: Very handy for cases like this 😄 There's also [`raise ... from ...`](https://www.pythontutorial.net/python-oop/python-raise-from/), which can be really useful for making stack traces more specific!

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
AetherUnbound: Great idea!

PR #650, opened 2022-08-06
Add configuration options to skip ingestion errors

### LINKED ISSUE #367: Allow failed ingestion tasks to retry where they left off
## Problem

See #357 

If an ingestion task fails, a retry can be attempted. This retry has no context for the previous run, and so it will begin reprocessing the data from the API at the very beginning, rather than where it left off.

## Description

We should investigate the feasibility of allowing a task retry to pick up from the last set of query parameters.

We'll need to be careful how we implement this, as the ramifications might cause problems for us. If we attempt to retry within the same DAG, we'll create a situation where different task attempts actually have different outcomes. This goes a bit against the Airflow paradigm, particularly since Airflow displays the most recent log first and said log may not have all of the run information (if previous runs captured _some_ amount of data but later failed). Additionally, the downstream loader steps will also require multiple attempts, each of which may result in separate successful data loads.

I think one alternative to this may be:
1. Scheduled runs will ignore any preserved parameter information and start from the beginning
2. Manual DAG runs will attempt to use previous parameter information

This would let us start _another_ run after a failure manually, which would pick up the query parameters and continue from where the previous run left off. It would also encapsulate that in two (or more) distinct DAG runs, so we don't have to expect an Airflow user to look through multiple task retry logs to get the full picture.

## Implementation

stacimc:
## Fixes

Fixes #367 by @AetherUnbound 

## Background
_Problem_
If a provider workflow encounters errors during the `pull_data` step, the task will fail and any partial data that was collected will be loaded. To resume ingestion, the DAG must be manually re-reun, and because the ingestion task has no context for the previous run, it will re-process all of the data from the API from the very beginning. This can be very time consuming!

_A note about alternative solutions_
I attempted several strategies for allowing the `pull_data` task to _retry_, picking up from the last set of query parameters. This goes against the philosophy of Airflow, which expects tasks to be idempotent. Ultimately, even if we were willing to slightly misuse Airflow in this way, I think it would be very difficult if not impossible. For instance, Airflow actually explicitly [clears XComs](https://airflow.apache.org/docs/apache-airflow/stable/concepts/xcoms.html) on every task retry.

## Description

This PR attempts to tackle the problem with the addition of two new DagRun configuration options, which will be supported out of the box by all provider workflows extending the `ProviderDataIngester`:
* `skip_ingestion_errors`: When set to true, any errors encountered during the fetching/processing of batch data will be _caught_, and ingestion allowed to continue. At the end of the ingestion, if there were any caught errors the task will still fail and report a summary of all encountered errors.
* `initial_query_params`: When set, this will be passed in as the first set of query_params, and ingestion will resume from there.
* `query_params_list`: A list of query_params. When set, ingestion will run for just this list of query_params. This can be used to re-run ingestion for just the problematic batches. 

This PR also adds a new `IngestionError` custom exception which extends error messages to also report the `query_params` that were being used when the error was reached. These values can then be pasted into the DagRun configuration `initial_query_params` setting to re-run starting at the failed batch.

## Notes/To Do
Because errors are caught within `ingest_records`, this means batches are skipped as opposed to individual records. We could fine tune this by only (or additionally) wrapping `get_record_data` in the try/except. What do you think?

### Setup
To test this you will need to make a provider script throw periodic errors. I used Cleveland Museum and edited [get_batch_data](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py#L36) to add the following:
```
# Throw errors on batches 3, 6, and 8 (since skip starts at 0)
if response_json['info']['parameters']['skip'] in ['200', '500', '700']:
            raise Exception("Test error")
```

I also updated the `batch_limit` of Cleveland Museum to `100` to make it run slightly faster (if you do not do this, you'll need to use values `['2000', '5000', '7000']` in the code above).

I also capped the amount of records processed by adding an `ingestion_limit` Airflow variable in the UI with a value of 1000.

### Tests

Let's simulate how we might use these configuration variables in the real world!

#### Simulate a scheduled DagRun (no conf options supplied) encountering an error
- [ ] Run the Cleveland Museum workflow. If configured like mine, it will throw an error on page 2. Verify that the new Error message looks something like this:
```
[2022-08-06, 00:20:43 UTC] {taskinstance.py:1889} ERROR - Task failed with exception
providers.provider_api_scripts.provider_data_ingester.IngestionError: Test error
    query_params: {"cc": "1", "has_image": "1", "limit": 100, "skip": 200}

The above exception was the direct cause of the following exception:

<...>
```

#### Re-start the DAG from the point at which it failed, and silence errors
You can trigger a manual run with a config by right clicking the play button and selecting the option:

<p align="center">
Screen Shot 2022-08-05 at 5 22 03 PM
</p>

Here's an example of what my configuration looks like. For the `initial_query_params`, I've just copied directly from the error message in the previous step.
<p align="center">
Screen Shot 2022-08-12 at 12 54 17 PM
</p>

- [ ] When `skip_ingestion_errors` is enabled, the DAG should run until the `ingestion_limit` is reached, and then the `pull_data` task should fail with a single `AggregateIngestionError`. Check the logs to see that all of the individual errors with their tracebacks were logged, along with a list of the query_params that caused the issues. This is logged separately for ease of copy/pasting into the dagrun conf.

- [ ] Check the logs to verify that the `initial_query_params` used were the expected ones. You should see something like:
```
Using initial_query_params from dag_run conf: {"cc": "1", "has_image": "1", "limit": 100, "skip": 200}
```

- [ ] Optionally play around with only setting one of `skip_ingestion_errors` or `initial_query_params` at a time

#### Simulate 'fixing the bug' and re-run the DAG for only the failed query_parameters
Our bug is easy to fix: just remove the lines you inserted into the provider script to raise the Exception 😄 Now we want to re-run the DAG but only for the affected batches. Here's my conf (you can copy/paste the list of query_params that resulted in errors from the logs of the failed run for convenience):
<p align="center">
Screen Shot 2022-08-10 at 4 31 58 PM
</p>

- [ ] Re-run the DAG with the `query_params_list` and check the logs to make sure that `get_batch` was only run for those params!

- [ ] Finally, run a provider_workflow that uses the ProviderDataIngester and which has _not_ been configured to throw errors (eg Science Museum). Make sure it still works/passes.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
@@ -21,17 +21,15 @@ class IngestionError(Exception):
     were being used when the error was encountered.
     """
 
-    def __init__(self, error, traceback, query_params, next_query_params):
+    def __init__(self, error, traceback, query_params):
         self.error = error
         self.traceback = traceback
         self.query_params = query_params
-        self.next_query_params = next_query_params
 
     def __str__(self):
         # Append query_param info to error message
         return f"""{self.error}
-    query_params: {json.dumps(self.query_params)}
-    next_query_params: {json.dumps(self.next_query_params)}"""
+    query_params: {json.dumps(self.query_params)}"""
 
     def print_with_traceback(self):
         # Append traceback
@@ -142,22 +140,18 @@ class ProviderDataIngester(ABC):
         """
         should_continue = True
         record_count = 0
-
-        # Get initial query_params
-        if query_params := self.initial_query_params:
-            logger.info(f"Using initial_query_params from dag_run conf: {query_params}")
-        else:
-            query_params = self.get_query_params(None, **kwargs)
+        query_params = None
 
         logger.info(f"Begin ingestion for {self.__class__.__name__}")
 
         while should_continue:
-            try:
+            query_params = self.get_query_params(query_params, **kwargs)
+            if query_params is None:
                 # Break out of ingestion if no query_params are supplied. This can
                 # happen when the final `manual_query_params` is processed.
-                if query_params is None:
-                    break
+                break
 
+            try:
                 batch, should_continue = self.get_batch(query_params)
 
                 if batch and len(batch) > 0:
@@ -168,27 +162,20 @@ class ProviderDataIngester(ABC):
                     should_continue = False
 
             except Exception as error:
-                next_query_params = self.get_query_params(query_params, **kwargs)
-
                 ingestion_error = IngestionError(
-                    error, traceback.format_exc(), query_params, next_query_params
+                    error, traceback.format_exc(), query_params
                 )
 
                 if self.skip_ingestion_errors:
                     # Add this to the errors list but continue processing
                     self.ingestion_errors.append(ingestion_error)
-                    logger.error(f"Skipping ingestion error: {error}")
-
-                    query_params = next_query_params
+                    logger.error(f"Skipping batch due to ingestion error: {error}")
                     continue
 
                 # Commit whatever records we were able to process, and rethrow the
                 # exception so the taskrun fails.
                 self.commit_records()
                 raise error from ingestion_error
-
-            # Update query params before iterating
-            query_params = self.get_query_params(query_params, **kwargs)
 
             if self.limit and record_count >= self.limit:
                 logger.info(f"Ingestion limit of {self.limit} has been reached.")
@@ -210,14 +197,25 @@ class ProviderDataIngester(ABC):
         self, prev_query_params: Optional[Dict], **kwargs
     ) -> Optional[Dict]:
         """
-        Returns the next set of query_params for the next request. If a
-        `query_params_list` has been provided in the dag_run conf, it will fetch
-        the next set of query_params from that list. If not, it just passes through
-        to the class implementation of `get_next_query_params`.
+        Returns the next set of query_params for the next request, handling
+        optional overrides via the dag_run conf.
         """
+        # If we are getting query_params for the first batch and initial_query_params
+        # have been set, return them.
+        if prev_query_params is None and self.initial_query_params:
+            logger.info(
+                "Using initial_query_params from dag_runconf:"
+                f" {self.initial_query_params}"
+            )
+            return self.initial_query_params
+
+        # If a list of query_params was provided, return the next value.
         if self.override_query_params:
             next_params = next(self.override_query_params, None)
-            logger.info(f"Using query params from dag_run conf: {next_params}")
+            logger.info(
+                "Using query params from `query_params_list` set in dag_run conf:"
+                f" {next_params}"
+            )
             return next_params
 
         # Default behavior when no conf options are provided; build the next
--- a/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
@@ -160,6 +160,16 @@ class ProviderDataIngester(ABC):
                     logger.info("Batch complete.")
                     should_continue = False
 
+            except AirflowException as error:
+                # AirflowExceptions should not be caught, as execution should not
+                # continue when the task is being stopped by Airflow.
+
+                # If errors have already been caught during processing, raise them
+                # as well.
+                if error_summary := self.get_ingestion_errors():
+                    raise error_summary from error
+                raise
+
             except Exception as error:
                 ingestion_error = IngestionError(
                     error, traceback.format_exc(), query_params
@@ -184,6 +194,16 @@ class ProviderDataIngester(ABC):
         self.commit_records()
 
         # If errors were caught during processing, raise them now
+        if error_summary := self.get_ingestion_errors():
+            raise error_summary
+
+    def get_ingestion_errors(self) -> AirflowException | None:
+        """ "
+        If any errors were skipped during ingestion, returns an AirflowException
+        with a summary of all errors.
+
+        It there are no errors to report, returns None.
+        """
         if self.ingestion_errors:
             # Log the affected query_params
             bad_query_params = ", \n".join(
@@ -196,9 +216,10 @@ class ProviderDataIngester(ABC):
             errors_str = "\n".join(
                 e.repr_with_traceback() for e in self.ingestion_errors
             )
-            raise AirflowException(
+            return AirflowException(
                 f"The following errors were encountered during ingestion:\n{errors_str}"
             )
+        return None
 
     def get_query_params(
         self, prev_query_params: Optional[Dict], **kwargs
--- a/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
@@ -103,9 +103,22 @@ class ProviderDataIngester(ABC):
         self.media_stores = self.init_media_stores()
         self.date = date
 
-        # Keep track of ingestion errors
-        self.conf = conf or {}
-        self.ingestion_errors = []
+        # dag_run configuration options
+        conf = conf or {}
+
+        # Used to skip over errors and continue ingestion. When enabled, errors
+        # are not reported until ingestion has completed.
+        self.skip_ingestion_errors = conf.get("skip_ingestion_errors", False)
+        self.ingestion_errors: List[IngestionError] = []  # Keep track of errors
+
+        # An optional set of initial query params from which to begin ingestion.
+        self.initial_query_params = conf.get("initial_query_params")
+
+        # An optional list of `query_params`. When provided, ingestion will be run for
+        # just these sets of params.
+        self.override_query_params = None
+        if query_params_list := conf.get("query_params_list"):
+            self.override_query_params = (qp for qp in query_params_list)
 
     def init_media_stores(self) -> dict[str, MediaStore]:
         """
@@ -131,17 +144,20 @@ class ProviderDataIngester(ABC):
         record_count = 0
 
         # Get initial query_params
-        query_params = self.conf.get("initial_query_params")
-        if not query_params:
-            query_params = self.get_next_query_params(None, **kwargs)
+        if query_params := self.initial_query_params:
+            logger.info(f"Using initial_query_params from dag_run conf: {query_params}")
+        else:
+            query_params = self.get_query_params(None, **kwargs)
 
-        logger.info(
-            f"Begin ingestion for {self.__class__.__name__}"
-            f" using initial query params: {query_params}"
-        )
+        logger.info(f"Begin ingestion for {self.__class__.__name__}")
 
         while should_continue:
             try:
+                # Break out of ingestion if no query_params are supplied. This can
+                # happen when the final `manual_query_params` is processed.
+                if query_params is None:
+                    break
+
                 batch, should_continue = self.get_batch(query_params)
 
                 if batch and len(batch) > 0:
@@ -152,13 +168,13 @@ class ProviderDataIngester(ABC):
                     should_continue = False
 
             except Exception as error:
-                next_query_params = self.get_next_query_params(query_params, **kwargs)
+                next_query_params = self.get_query_params(query_params, **kwargs)
 
                 ingestion_error = IngestionError(
                     error, traceback.format_exc(), query_params, next_query_params
                 )
 
-                if self.conf.get("skip_ingestion_errors", False):
+                if self.skip_ingestion_errors:
                     # Add this to the errors list but continue processing
                     self.ingestion_errors.append(ingestion_error)
                     logger.error(f"Skipping ingestion error: {error}")
@@ -171,12 +187,12 @@ class ProviderDataIngester(ABC):
                 self.commit_records()
                 raise error from ingestion_error
 
+            # Update query params before iterating
+            query_params = self.get_query_params(query_params, **kwargs)
+
             if self.limit and record_count >= self.limit:
                 logger.info(f"Ingestion limit of {self.limit} has been reached.")
                 should_continue = False
-
-            # Update query_params before iterating
-            query_params = self.get_next_query_params(query_params, **kwargs)
 
         # Commit whatever records we were able to process
         self.commit_records()
@@ -189,6 +205,24 @@ class ProviderDataIngester(ABC):
             raise AirflowException(
                 f"The following errors were encountered during ingestion:\n{errors_str}"
             )
+
+    def get_query_params(
+        self, prev_query_params: Optional[Dict], **kwargs
+    ) -> Optional[Dict]:
+        """
+        Returns the next set of query_params for the next request. If a
+        `query_params_list` has been provided in the dag_run conf, it will fetch
+        the next set of query_params from that list. If not, it just passes through
+        to the class implementation of `get_next_query_params`.
+        """
+        if self.override_query_params:
+            next_params = next(self.override_query_params, None)
+            logger.info(f"Using query params from dag_run conf: {next_params}")
+            return next_params
+
+        # Default behavior when no conf options are provided; build the next
+        # set of query params, given the previous.
+        return self.get_next_query_params(prev_query_params, **kwargs)
 
     @abstractmethod
     def get_next_query_params(
--- a/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import traceback
 from abc import ABC, abstractmethod
 from typing import Dict, List, Optional, Tuple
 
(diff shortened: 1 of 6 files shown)


stacimc:
@AetherUnbound @obulat I'm putting this into review for early feedback, but especially curious for thoughts on the overall approach and my question about catching errors in processing individual records versus batches.

obulat (commented):
This is a really interesting solution to a very difficult problem, @stacimc ! I really like it.

>Because errors are caught within ingest_records, this means batches are skipped as opposed to individual records. We could fine tune this by only (or additionally) wrapping get_record_data in the try/except. What do you think?

I think using batches is the best option because a batch uses the same query parameters, so we probably won't win too much if we try re-trying with individual records: we would still make a lot of requests anyways.

We could also add a different mode to the ProviderDataIngester for re-trying the batches that raised errors. It could accept a list of `next_query_parameters` and would not attempt to build the new queries. It would only go through the URLs for the failed batches. What do you think, @stacimc ?

stacimc:
@obulat That's a great idea! Definitely feels useful to be able to re-run just the failed batches after a bug fix for example. I've added a `query_params_list` configuration option that should do just that.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -101,23 +144,89 @@ def ingest_records(self, **kwargs) -> None:
...
+                f"{bad_query_params}"
+            )
+            errors_str = "\n".join(
+                e.print_with_traceback() for e in self.ingestion_errors
+            )
+            raise AirflowException(
+                f"The following errors were encountered during ingestion:\n{errors_str}"
+            )
+
+    def get_query_params(
```
stacimc, line 201: I was attempting to avoid renaming the current `get_next_query_params` method, but I'm not happy with this naming. Any suggestions?

stacimc:
This is going to take a little bit of thought to rebase with #645. I'll work on that today.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -12,6 +15,28 @@
 logger = logging.getLogger(__name__)
 
 
+class IngestionError(Exception):
```
AetherUnbound, line 18: We have a couple of places where custom exceptions are defined:
```bash
$ rg -g "*.py" "^class .*Exception"
openverse_catalog/dags/common/requester.py
23:class RetriesExceeded(Exception):

openverse_catalog/dags/common/licenses/licenses.py
21:class InvalidLicenseURLException(Exception):

openverse_catalog/dags/retired/common/pg_cleaner.py
79:class CleaningException(Exception):

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
18:class IngestionError(Exception):
```

Do you think there's any value in centralizing these into a `common/exceptions.py` file?

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -12,6 +15,28 @@
...
+
+    def __str__(self):
+        # Append query_param info to error message
+        return f"""{self.error}
+    query_params: {self.query_params}"""
+
+    def print_with_traceback(self):
+        # Append traceback
+        return f"""{str(self)}
+    {self.traceback}"""
```
AetherUnbound, line 37: This will include the traceback, but it looks like it will also include the indentation on the first line. This might be a bit better to prevent having to `dedent` or include erroneous whitespace:
```suggestion
        return f"{str(self)}\n{self.traceback}"
```

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -12,6 +15,28 @@
...
+        self.error = error
+        self.traceback = traceback
+        self.query_params = json.dumps(query_params)
+
+    def __str__(self):
+        # Append query_param info to error message
+        return f"""{self.error}
+    query_params: {self.query_params}"""
+
+    def print_with_traceback(self):
```
AetherUnbound, line 34: This doesn't quite print, which made it a bit confusing amidst the `logger.info` and `raise AirflowException` calls below. I can't really think of a better function name, maybe `repr_with_traceback`? 😅

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -75,6 +101,23 @@ def __init__(self, date: str = None):
...
+        self.ingestion_errors: List[IngestionError] = []  # Keep track of errors
+
+        # An optional set of initial query params from which to begin ingestion.
+        self.initial_query_params = conf.get("initial_query_params")
+
+        # An optional list of `query_params`. When provided, ingestion will be run for
+        # just these sets of params.
+        self.override_query_params = None
+        if query_params_list := conf.get("query_params_list"):
+            self.override_query_params = (qp for qp in query_params_list)
```
AetherUnbound, line 119: Nit - potentially:
```suggestion
            self.override_query_params = (*query_params_list)
```

or
```suggestion
            self.override_query_params = tuple(query_params_list)
```

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -101,10 +144,14 @@ def ingest_records(self, **kwargs) -> None:
...
         logger.info(f"Begin ingestion for {self.__class__.__name__}")
 
-        try:
-            while should_continue:
-                query_params = self.get_next_query_params(query_params, **kwargs)
+        while should_continue:
+            query_params = self.get_query_params(query_params, **kwargs)
+            if query_params is None:
+                # Break out of ingestion if no query_params are supplied. This can
+                # happen when the final `manual_query_params` is processed.
```
AetherUnbound, line 151: Should this be?
```suggestion
                # happen when the final `override_query_params` is processed.
```

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -114,12 +161,74 @@ def ingest_records(self, **kwargs) -> None:
                     logger.info("Batch complete.")
                     should_continue = False
 
-                if self.limit and record_count >= self.limit:
-                    logger.info(f"Ingestion limit of {self.limit} has been reached.")
-                    should_continue = False
-        finally:
-            total = self.commit_records()
-            logger.info(f"Committed {total} records")
+            except Exception as error:
```
AetherUnbound, line 164: Based on the changes in #645, I think we'll need to be careful about bare `Exception` catches. It looks like if `skip_ingestion_errors` is `True`, then execution will continue. This breaks the case where the task is stopped by Airflow. I _think_ we can catch that specific case by handling `AirflowException` differently and immediately re-raising that though!

tests/dags/providers/provider_api_scripts/test_provider_data_ingester.py
```diff
@@ -20,223 +20,315 @@
     os.path.abspath(os.path.dirname(__file__)), "resources/provider_data_ingester"
 )
 
+ingester = MockProviderDataIngester()
```
AetherUnbound, line 23: Yay! Very happy to see these become functional tests 😄

tests/dags/providers/provider_api_scripts/test_provider_data_ingester.py
```diff
@@ -20,223 +20,315 @@
...
+        ]
+
+        with pytest.raises(Exception) as error:
+            ingester.ingest_records()
+
+        # By default, `skip_ingestion_errors` is False and get_batch_data
+        # is no longer called after encountering an error
+        assert get_batch_mock.call_count == 1
+
+        assert str(error.value) == "Mock exception message"
```
AetherUnbound, line 296: FWIW we can use `pytest.raises`'s `match` keyword to run this last check for us:
```suggestion
        with pytest.raises(Exception, match="Mock exception message") as error:
            ingester.ingest_records()

        # By default, `skip_ingestion_errors` is False and get_batch_data
        # is no longer called after encountering an error
        assert get_batch_mock.call_count == 1
```

Then if the exception doesn't match pytest will reraise whatever occurs in `ingest_records`.

AetherUnbound (commented):
This is fantastic! Way more thorough than I was initially thinking with this issue, which is excellent. This gives us a lot of fine-grain control for reruns which is exactly what we need 🤩 

I have a few comments, my biggest concern is conformity to the changes in #645 with the main loop.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
stacimc: This line is creating a generator from the `query_params_list`, so that we can repeatedly call `next(self.override_query_params)` to get the next manually configured set of query_params. I can add a comment to that effect.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
stacimc: Yep! Variable name changed, thank you.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
stacimc: Thank you, this was bugging me but I couldn't come up with anything better! `repr_with_traceback` is a bit better, sold.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
stacimc: Hm that's a good point. I think each of those custom exceptions are only used within their respective module, with the exception of `RetriesExceeded` being used once here. I don't imagine the `IngestionError` would be used outside of this module, either, although maybe a use case will come up. I think I'm slightly biased to keeping them where they are, at least until that situation changes.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
stacimc: Extremely good point. I think we'll also want to make sure we raise any ingestion errors that have previously been skipped before we encounter this.

What do you think about making `skip_ingestion_errors` a list of Exception types to skip, rather than a simple boolean?

tests/dags/providers/provider_api_scripts/test_provider_data_ingester.py
stacimc: For some reason I always forget this! Thanks 😄

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
AetherUnbound: That works for me! You're right, each of these are pretty context specific. If it becomes super hairy down the line maybe we can centralize them then.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
AetherUnbound: D'oh! I always forget about generator syntax, thank you!

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
AetherUnbound: I like it! Only trouble would be matching on the exceptions since we have to go from a string to an exception type 😅 I guess we could just do `if exception.__type__.__name__ in skip_ingestion_errors`?

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -161,6 +160,16 @@ def ingest_records(self, **kwargs) -> None:
...
                     should_continue = False
 
+            except AirflowException as error:
+                # AirflowExceptions should not be caught, as execution should not
+                # continue when the task is being stopped by Airflow.
+
+                # If errors have already been caught during processing, raise them
+                # as well.
+                if error_summary := self.get_ingestion_errors():
+                    raise error_summary from error
```
AetherUnbound, line 170: LOVE this!

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
AetherUnbound: Maybe `determine_query_params`?

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -114,12 +160,95 @@ def ingest_records(self, **kwargs) -> None:
...
+                [f"{e.query_params}" for e in self.ingestion_errors]
+            )
+            logger.info(
+                "The following query_params resulted in errors: \n"
+                f"{bad_query_params}"
+            )
+            errors_str = "\n".join(
+                e.repr_with_traceback() for e in self.ingestion_errors
+            )
+            return AirflowException(
```
AetherUnbound, line 219: Thinking on this again, should this maybe be a new exception type, like `AggregateIngestionError`? `AirflowException` doesn't seem to be the right one to use 🤔

AetherUnbound: Since this exception will also get sent to Slack, maybe it would be ideal to log the more verbose errors and then raise a single exception at the end that says something along the lines of `{len(self.ingestion_errors)} ingestion errors occurred while running`?

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -114,12 +160,95 @@ def ingest_records(self, **kwargs) -> None:
...
+        """
+        Returns the next set of query_params for the next request, handling
+        optional overrides via the dag_run conf.
+        """
+        # If we are getting query_params for the first batch and initial_query_params
+        # have been set, return them.
+        if prev_query_params is None and self.initial_query_params:
+            logger.info(
+                "Using initial_query_params from dag_run conf:"
+                f" {self.initial_query_params}"
```
AetherUnbound, line 236: Not sure how important or useful this will be, but it seems like the string that gets spit out here uses single quotes rather than double. When I try to paste that directly into the dagrun conf box, it gets upset about the difference in quotes:

image

```
[2022-08-15, 19:46:33 UTC] {provider_data_ingester.py:144} INFO - Begin ingestion for ClevelandDataIngester
[2022-08-15, 19:46:33 UTC] {provider_data_ingester.py:234} INFO - Using initial_query_params from dag_run conf: {'cc': '1', 'has_image': '1', 'limit': 100, 'skip': 200}
```

AetherUnbound (commented):
This is so close! I just have some thoughts about the final exception and what gets logged.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
stacimc: Good call -- I had updated the format of the query_params when they're logged in the error message since I figured folks are likely to copy/paste from there, but this might be copied as well.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
stacimc: I like `AggregateIngestionError` (I had considered making a custom error before and couldn't come up with a name 😅 ). Totally agreed on the error logging -- I was under the impression the Slack messages would truncate but it doesn't look like that's the case, thanks for testing!

Updated to only _log_ the verbose errors, while throwing an `AggregateIngestionError`.

AetherUnbound (approved):
I tweaked the aggregate exception message, hope that's okay! This looks _fantastic_ and works great. I'm very excited for us to be able to employ this going forward 🚀 ⚡ 🔧

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -229,8 +229,9 @@ def get_ingestion_errors(self) -> AggregateIngestionError | None:
                 f"The following errors were encountered during ingestion:\n{errors_str}"
             )
             return AggregateIngestionError(
-                f"{len(self.ingestion_errors)} errors were skipped during ingestion "
-                " using the `skip_ingestion_errors` flag. See the log for more details."
+                f"{len(self.ingestion_errors)} query batches were skipped due to "
+                "errors during ingestion using the `skip_ingestion_errors` flag. "
```
stacimc, line 233: > I tweaked the aggregate exception message, hope that's okay!

Very much so, this adds valuable context!

obulat (approved):
I love how this well this worked together with #645, and the evolution of the error messages in the logs was really entertaining :)
Great improvement, hopefully we'll have more efficient ingestion process with fewer request to the providers.

PR #654, opened 2022-08-11
Allow DAGs to silence only errors matching predicate

stacimc:
## Fixes

Fast follow from  #643 to implement a suggestion by @sarayourfriend 

⚠️ Because this introduces a breaking change to the configuration variable, we must make sure the variable is not configured in production before merging. If it is configured, we need to make sure no affected DAGs (`check_silenced_dags`, any silenced DAGs) are running while we deploy these changes and update the variable interface. [Discussion](https://github.com/WordPress/openverse-catalog/pull/654#issuecomment-1212176723).  ⚠️ 

## Description

#643 introduced an Airflow variable (`silenced_slack_alerts`) that allows configuring a DAG to skip Slack reporting of errors. The implementation in that PR only allows for skipping Slack reporting for _all_ error messages, but in reality we're unlikely to want this behavior. It's much safer to be able to silence specific known errors.

This PR changes the shape of the `silenced_slack_alerts`:
Screen Shot 2022-08-10 at 7 10 51 PM

For each `dag_id` you must now provide both an `issue` and a list of `errors`. A slack alert will not be sent if it matches any of the predicates in the `errors` list for the given DAG.

Because our [on_failure_callback](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/slack.py#L314) includes information like the ExceptionType and task id in the message, we _can_ use this to selectively silence messages matching those parameters as well.

## Testing Instructions

* Create an Airflow variable named `silenced_slack_alerts` in the Admin > Variables UI. For my tests I used this configuration:
```
{
  "audio_data_refresh": {
      "issue": "https://github.com/WordPress/openverse-catalog/issues/438",
      "errors": ["A test exception"]
  }
}
```
* For the DAGs you've configured above, force an error to be thrown by editing the code to manually raise one. Make sure that it includes the configured predicate.
* Run the DAG and verify that you can see a message in the logs saying `Skipping Slack alert for <dag_id>`.
* Play around with your test error and the `errors` configuration to test that errors that do not match any of the predicates are still reported to Slack.

Now test `check_silenced_dags` DAG:
* Edit your configuration to link to a GH issue that is closed. Run the DAG and verify you get a Slack alert indicating that the alerts can be turned back on.
* Edit your configuration to link to a GH issue that is open. Run the DAG and verify you do not see a Slack alert.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/slack.py
+++ b/openverse_catalog/dags/common/slack.py
@@ -278,7 +278,7 @@ def send_alert(
         "silenced_slack_alerts", default_var={}, deserialize_json=True
     )
     if dag_id in known_failures and any(
-        error in text for error in known_failures[dag_id]["errors"]
+        error.lower() in text.lower() for error in known_failures[dag_id]["errors"]
     ):
         log.info(f"Skipping Slack alert for {dag_id}: {text}")
         return
--- a/openverse_catalog/dags/common/slack.py
+++ b/openverse_catalog/dags/common/slack.py
@@ -277,7 +277,9 @@ def send_alert(
     known_failures = Variable.get(
         "silenced_slack_alerts", default_var={}, deserialize_json=True
     )
-    if dag_id in known_failures:
+    if dag_id in known_failures and any(
+        error in text for error in known_failures[dag_id]["errors"]
+    ):
         log.info(f"Skipping Slack alert for {dag_id}: {text}")
         return
 
--- a/openverse_catalog/dags/maintenance/check_silenced_dags/check_silenced_dags.py
+++ b/openverse_catalog/dags/maintenance/check_silenced_dags/check_silenced_dags.py
@@ -24,7 +24,8 @@ def get_dags_with_closed_issues(github_pat, silenced_dags):
     gh = GitHubAPI(github_pat)
 
     dags_to_reenable = []
-    for dag_id, issue_url in silenced_dags.items():
+    for dag_id, conf in silenced_dags.items():
+        issue_url = conf["issue"]
         owner, repo, issue_number = get_issue_info(issue_url)
         github_issue = gh.get_issue(repo, issue_number, owner)
 
--- a/tests/dags/common/test_slack.py
+++ b/tests/dags/common/test_slack.py
@@ -364,6 +364,14 @@ def test_send_alert():
             "Error triggering data refresh",
             False,
         ),
+        # error matches are case-insensitive
+        (
+            [
+                "kEYeRROR",
+            ],
+            "KeyError: 'image'",
+            False,
+        ),
         # none of the silenced errors matches alert string
         (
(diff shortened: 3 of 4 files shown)


sarayourfriend (commented):
General question for airflow deployment stuff: when merging this PR change, what is the process for handling the change in the variable interface? These changes are not backwards compatible with the existing variable interface, so this dag would crash on a type error (I think). Is the process to first turn the DAG off in production, then in any order, merge this PR and update the variable to the new interface, then re-enable the DAG after the new DAG code has been pulled?

stacimc:
> General question for airflow deployment stuff: when merging this PR change, what is the process for handling the change in the variable interface? These changes are not backwards compatible with the existing variable interface, so this dag would crash on a type error (I think). Is the process to first turn the DAG off in production, then in any order, merge this PR and update the variable to the new interface, then re-enable the DAG after the new DAG code has been pulled?

In general, yes. In this case this would mean turning off `check_silenced_dags` and making sure no actually configured silenced DAGs are running (or temporarily reenabling alerting for them).

Because we don't have the variable set up yet, we should be fine to simply merge this as a fast-follow -- but I'll update the PR description to make sure we double-check that before merging.

openverse_catalog/dags/common/slack.py
```diff
@@ -277,7 +277,9 @@ def send_alert(
     known_failures = Variable.get(
         "silenced_slack_alerts", default_var={}, deserialize_json=True
     )
-    if dag_id in known_failures:
+    if dag_id in known_failures and any(
+        error in text for error in known_failures[dag_id]["errors"]
```
AetherUnbound, line 281: I think it might be good to do a case-insensitive comparison here, especially for longer error messages where case may differ (we'll also want new tests to this affect should this change be accepted):
```suggestion
        error.lower() in text.lower() for error in known_failures[dag_id]["errors"]
```

AetherUnbound (commented):
This looks great! I was worried we'd have to match on exception text, which would be an issue for things like `KeyError`s where the error text is only the missing key and nothing else 😶 Hard to know that ahead of time! But since we're doing a simple substring match _and_ we include the exception type in the slack message, we can match that too 🥳 I used `KeyError` as my `errors` value and raised a `KeyError` within the DAG, it worked exactly as expected 🚀 

I have one suggestion regarding string comparisons.

openverse_catalog/dags/common/slack.py
stacimc: Good point! Done ✔️

AetherUnbound (approved):
🎊 🤖

sarayourfriend (approved):
Awesome!

PR #651, opened 2022-08-08
Only delete dag runs/task instances during testing that match pattern

AetherUnbound:
## Fixes

Fixes a testing issue identified by @stacimc

## Description

After the merging of #636, it appeared that our tests were failing on `main`. Further investigation showed that tests would run fine synchronously (e.g. `just test -n 1`) but would fail then run in parallel. I was able to track this down to the `test_single_run_external_dags_sensor.py` tests, which were deleting all `DagRun`s and `TaskInstances` as part of setup/teardown. This PR changes those deletions to match a specific DAG prefix, which is used when creating DAGs throughout the test.

## Testing Instructions

`just test` should be sufficient.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #660, opened 2022-08-12
Update audioset_view to use most recently updated f_id/provider pair

### LINKED ISSUE #658: Handle duplicate audioset records
## Description

The audio data refresh fails during the `promote` step due to duplicate keys, historically from the Freesound provider:
Screen Shot 2022-08-11 at 11 33 59 AM

One initial theory is that this may happen when an audio provider script fails, and only partially ingests data. A real example from Freesound:
* A user created an audio_set
* Provider script runs and ingests the data into the Catalog db
* The user changes their username
* Provider script runs and updates some but not all records in the audio set before failing. 

The audio_set view now contains multiple records with the same audio_set `foreign_identifier` and `provider`, because there are records with those values associated with distinct creators ([audioset view query](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/popularity/sql.py#L311))

We should investigate this theory and consider if there are other edge cases that may cause it, and then add handling (possibly in the loader tasks) to prevent duplicate audio_sets.

## Additional context

Sample data to reproduce the issue is associated with the audio data refresh runs on May 01 2022 and Aug 11 2022, related to the Freesound provider.

## Resolution

AetherUnbound:
## Fixes

Fixes #658 by @stacimc

## Description

_Please see #658 for extended context._

This PR alters the `audioset_view` to select records which are distinct on `foreign_identifier` & `provider`, using the most recently updated record as the source of truth. This prevents cases where some records which share `audio_set` information actually have different values within the JSONB body. Since this circumvents the previous `SELECT DISTINCT` clause, we were seeing errors in the data refresh due to violated unique key constraints.

The original issue arises from the fact that `audioset_view` is a `VIEW` (not a table or materialized view) that pulls information from the `audio.audio_set` JSONB column. Since `audio.audio_set` is not indexed, there are no constraints on the data within them. With Freesound, we got into a situation where the audio set's foreign identifier & provider were the same, but other fields within the JSONB body did not match. This would cause duplicate records with the same `foreign_identifier` and `provider` to appear. While this isn't a problem in the catalog, this view gets materialized into a table on the API that _does_ have unique constraints on it, and that's what was breaking the data refresh. This PR ensures we're only populating the `audioset_view` on the catalog with one record for each `foreign_identifier` & `provider`. The record chosen is determined by which was most recently updated.

## Testing Instructions

This has been tested successfully with a production data refresh! We also confirmed in the catalog that the new view included all of the old view's records _except_ the 21 records that violated the unique-on-fid-provider clause.

`just test` is failing because of #651 but will pass once that is merged & rebased.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/popularity/sql.py
+++ b/openverse_catalog/dags/common/popularity/sql.py
@@ -306,7 +306,7 @@ def create_audioset_view_query():
     """
     return dedent(
         f"""
-        CREATE VIEW public.{AUDIOSET_VIEW_NAME}
+        CREATE VIEW public.{AUDIO_VIEW_NAME}
         AS
           -- DISTINCT clause exists to ensure that only one record is present for a given
           -- foreign identifier/provider pair. This exists as a hard constraint in the API table
@@ -325,7 +325,7 @@ def create_audioset_view_query():
             ((audio_set ->> 'filetype'::text))::character varying(80) AS filetype,
             ((audio_set ->> 'thumbnail'::text))::character varying(1000) AS thumbnail,
             provider
-          FROM public.{AUDIOSET_VIEW_NAME}
+          FROM public.{AUDIO_VIEW_NAME}
           WHERE (audio_set IS NOT NULL)
           ORDER BY audio_set ->> 'foreign_identifier', provider, updated_on DESC;
         """  # noqa: E501
--- a/openverse_catalog/dags/common/popularity/sql.py
+++ b/openverse_catalog/dags/common/popularity/sql.py
@@ -306,7 +306,7 @@ def create_audioset_view_query():
     """
     return dedent(
         f"""
-        CREATE VIEW public.{AUDIO_VIEW_NAME}
+        CREATE VIEW public.{AUDIOSET_VIEW_NAME}
         AS
           -- DISTINCT clause exists to ensure that only one record is present for a given
           -- foreign identifier/provider pair. This exists as a hard constraint in the API table
--- a/openverse_catalog/dags/common/popularity/sql.py
+++ b/openverse_catalog/dags/common/popularity/sql.py
@@ -308,19 +308,26 @@ def create_audioset_view_query():
         f"""
         CREATE VIEW public.{AUDIOSET_VIEW_NAME}
         AS
-        SELECT DISTINCT
-            (audio_set ->> 'foreign_identifier')  :: varchar(1000) as foreign_identifier,
-            (audio_set ->> 'title')               :: varchar(2000) as title,
-            (audio_set ->> 'foreign_landing_url') :: varchar(1000) as foreign_landing_url,
-            (audio_set ->> 'creator')             :: varchar(2000) as creator,
-            (audio_set ->> 'creator_url')         :: varchar(2000) as creator_url,
-            (audio_set ->> 'url')                 :: varchar(1000) as url,
-            (audio_set ->> 'filesize')            :: integer       as filesize,
-            (audio_set ->> 'filetype')            :: varchar(80)   as filetype,
-            (audio_set ->> 'thumbnail')           :: varchar(1000) as thumbnail,
+          -- DISTINCT clause exists to ensure that only one record is present for a given
+          -- foreign identifier/provider pair. This exists as a hard constraint in the API table
+          -- downstream, so we must enforce it here. The audio_set data is chosen by which audio
+          -- record was most recently updated (see the final section of the ORDER BY clause
+          -- below). More info here:
+          -- https://github.com/WordPress/openverse-catalog/issues/658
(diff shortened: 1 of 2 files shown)


openverse_catalog/dags/common/popularity/sql.py
```diff
@@ -308,19 +308,26 @@ def create_audioset_view_query():
...
+            ((audio_set ->> 'creator'::text))::character varying(2000) AS creator,
+            ((audio_set ->> 'creator_url'::text))::character varying(2000) AS creator_url,
+            ((audio_set ->> 'url'::text))::character varying(1000) AS url,
+            ((audio_set ->> 'filesize'::text))::integer AS filesize,
+            ((audio_set ->> 'filetype'::text))::character varying(80) AS filetype,
+            ((audio_set ->> 'thumbnail'::text))::character varying(1000) AS thumbnail,
             provider
-        FROM public.{AUDIO_VIEW_NAME}
-        WHERE audio_set IS NOT NULL;
+          FROM public.{AUDIOSET_VIEW_NAME}
```
stacimc, line 328: This should be `FROM public.{AUDIO_VIEW_NAME}`, right?

stacimc (changes requested):
So excited for this fix! I think there may just be a typo in the view name there.

We can test this by running the `recreate_audio_popularity_calculation` DAG locally. Everything looks good after updating the view name!

openverse_catalog/dags/common/popularity/sql.py
AetherUnbound: Oh goodness, good catch!

openverse_catalog/dags/common/popularity/sql.py
```diff
@@ -306,7 +306,7 @@ def create_audioset_view_query():
     """
     return dedent(
         f"""
-        CREATE VIEW public.{AUDIOSET_VIEW_NAME}
+        CREATE VIEW public.{AUDIO_VIEW_NAME}
```
stacimc, line 309: Oh no sorry, I think _this_ one was correct as `AUDIOSET_VIEW_NAME`

AetherUnbound: Oh god, okay let me get it right this time 😅

stacimc (approved):
🎉

sarayourfriend:
@AetherUnbound can you explain why this fixes the issue? I'm not familiar enough with the audio ingestion flow to be able to review this in a timely manner without more explanation of the background of how it works to know what I am reviewing :confused:

AetherUnbound:
@sarayourfriend Sure! There's some explanation & context in the associated ticket: https://github.com/WordPress/openverse-catalog/issues/658
I was trying to avoid adding the same text from the issue into the PR, but I'll add some of that context to the PR description (and call out when checking out the original issue would be a good prerequisite in future PRs). If the issue itself also doesn't make sense, please let us know 🙂

sarayourfriend:
I did read the issue already, but I'm not able to connect the dots between the problem and the fix, sorry. Maybe someone with more background on the catalogue and the audio ingestion flow and data refresh flow would be a better fit to review this? I don't really have time right now to learn about this process as I try to wrap up other work and dealing with infrastructure issues before being AFK for three weeks.

@obulat would you be able to review this PR?

AetherUnbound:
Sounds good, definitely don't want this to detract from all the other stuff you're doing! I've re-assigned review to @obulat, if she also does not have time we can find someone else 🙂

openverse_catalog/dags/common/popularity/sql.py
```diff
@@ -308,19 +308,26 @@ def create_audioset_view_query():
...
-            (audio_set ->> 'filesize')            :: integer       as filesize,
-            (audio_set ->> 'filetype')            :: varchar(80)   as filetype,
-            (audio_set ->> 'thumbnail')           :: varchar(1000) as thumbnail,
+          -- DISTINCT clause exists to ensure that only one record is present for a given
+          -- foreign identifier/provider pair. This exists as a hard constraint in the API table
+          -- downstream, so we must enforce it here. The audio_set data is chosen by which audio
+          -- record was most recently updated (see the final section of the ORDER BY clause
+          -- below). More info here:
+          -- https://github.com/WordPress/openverse-catalog/issues/658
+          SELECT DISTINCT ON (audio_set ->> 'foreign_identifier', provider)
```
obulat, line 317: Can we keep the columns in the formatting to make this query easier to scan?

obulat (approved):
So sorry for late review! The fix looks great, I just have one non-blocking formatting comment inline.

Issue #588, opened 2022-07-01
Refactor Metropolitan Museum to use ProviderDataIngester


stacimc:
## Suggested Improvement

Refactor the Metropolitan museum provider script to use the new ProviderDataIngester base class

## Benefit

More details in #229

## Implementation

- [x] 🙋 I would be interested in implementing this feature.

rwidom:
I started working on this, and found a couple of things from [their repo documentation](https://metmuseum.github.io/).

- [4. At its discretion the Museum may restrict access to or discontinue The Met Collection API or any functionality associated with the API. The Museum may limit the number of transactions a user may send or receive through the API, which may, for example, ensure even performance for all users, or may in the future require access through assigned keys.](https://www.metmuseum.org/information/terms-and-conditions) Indeed, I had some access issues during testing.
- There is a csv file available for download [here](https://github.com/metmuseum/openaccess).

The csv is small enough that it would be easy enough to process with pandas. But I don't know how often it gets updated, it doesn't have image urls, etc. I'd be happy to reach out to someone at the museum, but I wonder if it would be more effective coming from staff. Tagging @AetherUnbound , @stacimc and @sarayourfriend  for your thoughts.

AetherUnbound:
Thanks for looking into this! What were you thinking of asking the museum?

AetherUnbound:
Making a note here that we'll want to be sure we incorporate the image titles in this refactor, see #641

rwidom:
> Thanks for looking into this! What were you thinking of asking the museum?

How often they update that extract (to evaluate the csv route), and if they have any more specific guidelines for rate limits that we could be using (if the API is really the better option).

> Making a note here that we'll want to be sure we incorporate the image titles in this refactor, see #641

Yes, will do! And I think we should at least consider adding the artist name and some other things to tags. I think it's currently under creator, but I don't know that creator is currently part of elastic search. Is it? So [this](https://wordpress.org/openverse/search/?q=Kiyohara%20Yukinobu) only brings back results from flickr, even though the artist is in some of our existing test cases, and [definitely at the met](https://www.metmuseum.org/search-results?q=Kiyohara+Yukinobu).

I have a first draft using the API workflow, that I could publish, but it needs some more testing, which I'm kind of blocked on. :/

AetherUnbound:
[...]

PR #674, opened 2022-08-18
Refactor Metropolitan Museum of Art to use ProviderDataIngester

### LINKED ISSUE #641: Correct title extraction for Metropolitan museum
## Problem

Many of the recently ingested items from the Metropolitan museum do not have titles in database. If you go to their foreign landing URL, you'll see that the items actually do have titles.

## Description

The Met script needs to be updated to always extract the title, if available.

## Additional context

Based on investigation in #524.

<details><summary>Here are some of the images </summary>

| title | foreign_landing_url                                    | created_on                    | updated_on                    | last_synced_with_source       | removed_from_source |
| --- | --- | --- | --- | --- | --- | 
|       | https://www.metmuseum.org/art/collection/search/855619 | 2022-07-20 09:29:52.842581+00 | 2022-07-20 09:29:52.842581+00 | 2022-07-20 09:29:52.842581+00 |  False               |
|       | https://www.metmuseum.org/art/collection/search/479    | 2020-07-01 18:22:30.284679+00 | 2020-07-08 06:49:27.92134+00  | 2020-07-08 06:49:27.92134+00  | False               |
|       | https://www.metmuseum.org/art/collection/search/747045 | 2022-03-26 00:03:07.946308+00 | 2022-07-31 00:08:14.325392+00 | 2022-07-31 00:08:14.325392+00 | False               |
|       | https://www.metmuseum.org/art/collection/search/5915   | 2020-07-01 18:22:30.284679+00 | 2020-07-03 06:28:16.151653+00 | 2020-07-03 06:28:16.151653+00 | False               |
|       | https://www.metmuseum.org/art/collection/search/4542   | 2020-07-01 18:22:30.284679+00 | 2020-07-03 06:28:16.151653+00 | 2020-07-03 06:28:16.151653+00 |  False               |
|       | https://www.metmuseum.org/art/collection/search/4565   | 2022-04-13 00:03:55.312458+00 | 2022-07-29 00:12:53.807042+00 | 2022-07-29 00:12:53.807042+00 | False               |
|       | https://www.metmuseum.org/art/collection/search/773049 | 2022-04-13 00:03:55.312458+00 | 2022-04-24 01:08:44.630586+00 | 2022-04-24 01:08:44.630586+00 |  False               |
|       | https://www.metmuseum.org/art/collection/search/51803  | 2020-07-03 06:28:16.151653+00 | 2020-07-03 06:28:16.151653+00 | 2020-07-03 06:28:16.151653+00 |False               |
|       | https://www.metmuseum.org/art/collection/search/747045 | 2022-03-26 00:03:07.946308+00 | 2022-07-31 00:08:14.325392+00 | 2022-07-31 00:08:14.325392+00 |  False               |
|       | https://www.metmuseum.org/art/collection/search/747045 | 2022-03-26 00:03:07.946308+00 | 2022-07-31 00:08:14.325392+00 | 2022-07-31 00:08:14.325392+00 |  False               |

</details>

## Implementation

rwidom:
## Fixes
Fixes #641 by @obulat and #588 by @stacimc

## Description
This refactors the metropolitan museum to use the provider data ingester class, and hopefully improve search results by populating title and tags more reliably.

## Testing Instructions
When I was testing the dag in airflow, I kept hitting non-responses, until I set the delay at 3 seconds, rather than the standard 1 second. I'm not sure how sustainable that will be for production, since we need to hit the API once for each and every record, including non-CC0 records that we do not load. So, I do recommend testing out the dag in airflow, but with the rate limit caveat.

image

```
*Provider*: `metropolitan_museum`
*Date range*: 2022-08-08 -> 2022-08-09
*Duration of data pull task*: _No data_
*Number of records upserted per media type*:
  - `image`: 9,857 _(985 duplicate foreign IDs, 2,758 duplicate URLs)_

[2022-08-17, 17:46:31 UTC] {taskinstance.py:1395} INFO - Marking task as SUCCESS. dag_id=metropolitan_museum_workflow, task_id=report_load_completion, execution_date=20220808T000000, start_date=20220817T174631, end_date=20220817T174631
[2022-08-17, 17:46:31 UTC] {local_task_job.py:156} INFO - Task exited with return code 0
[2022-08-17, 17:46:31 UTC] {local_task_job.py:273} INFO - 0 downstream tasks scheduled from follow-on schedule check
```

And then go to `db-shell` and check that title and tags are always populated, try clicking through some of the urls, and so on. For example...

image

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
```diff
@@ -6,138 +6,168 @@
...
+            {
+                "foreign_landing_url": object_json.get("objectURL"),
+                "image_url": img,
+                "license_info": self.DEFAULT_LICENSE_INFO,
+                "foreign_identifier": self._get_foreign_id(object_id, img),
+                "creator": artist,
+                "title": title,
+                "meta_data": meta_data,
+                "filetype": img.split(".")[-1],
+                "raw_tags": raw_tags,
```
obulat, line 109:
```suggestion
                "raw_tags": raw_tags,
```
If filetype is computed as the extension of the `image_url`, then it's best to not add it here. `ImageStore` will add the filetype, and will also unify some formats (i.e., use a single `jpg` instead of a mix of `jpeg` and `jpg`)

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
```diff
@@ -6,138 +6,168 @@
...
+                object_json.get("creditLine"),
+                object_json.get("period"),
+            ]
+            if tag
+        ]
+        if object_json.get("tags"):
+            tag_list += [tag["term"] for tag in object_json.get("tags")]
+        return tag_list
+
+    def _get_title(self, record):
```
obulat, line 146: Nice!

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
```diff
@@ -6,138 +6,168 @@
...
+            return record.get("title")
+        else:
+            return record.get("objectName")
+
+    def _get_artist_name(self, record):
+        artist = record.get("artistDisplayName")
+        # Treating "unidentified" the same as missing, but maybe it would be useful in
+        # in search? Maybe in the art world it has a more specific outsider art
+        # connotation?
+        if artist in ["Unidentified", "Unidentified artist"]:
```
obulat, line 157: Madison usually says that it's better to use a `set` when checking with `in` :)

tests/dags/providers/provider_api_scripts/resources/metropolitan_museum_of_art/read_met_csv.py
```diff
@@ -0,0 +1,158 @@
+"""
```
obulat, line 1: This is awesome, @rwidom!

obulat (approved):
Really nice and clean!

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
rwidom: Oh that's awesome! Thanks so much @obulat !

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
rwidom: 🤦  Yes, that makes total sense, I'll fix it tonight.

AetherUnbound: Ha, you know me too well @obulat 😁

tests/dags/providers/provider_api_scripts/test_metropolitan_museum.py
```diff
@@ -6,181 +6,164 @@
...
+def test_get_batch_data(response_json, expected):
+    actual = mma.get_batch_data(response_json)
+    assert actual == expected
+
+
+@pytest.mark.parametrize(
+    "case_name, response_json, expected",
+    [
+        (
+            "single image",
```
AetherUnbound, line 68: Do you think it would be reasonable to add the empty case for this and the next 3 tests? E.g. `response_json` is `{}` or `None`

tests/dags/providers/provider_api_scripts/test_metropolitan_museum.py
```diff
@@ -6,181 +6,164 @@
...
+            single_object_response,
+            single_expected_data[0].get("meta_data"),
+        ),
+        ("multi images", full_object_response, full_expected_data[0].get("meta_data")),
+    ],
+)
+def test_get_meta_data(case_name, response_json, expected):
```
AetherUnbound, line 75: It looks like in these tests that `case_name` isn't actually used. If you're looking to document the particular test cases, you can add comments to each parameter set:
```suggestion
    "response_json, expected",
    [
        # Single image response
        (
            single_object_response,
            single_expected_data[0].get("meta_data"),
        ),
        # Multi-image response
        (images", full_object_response, full_expected_data[0].get("meta_data")),
    ],
)
def test_get_meta_data(response_json, expected):
```

Here's another example: 

https://github.com/WordPress/openverse-catalog/blob/ee474f2cbf6f797092ec7be3f56fef580f9c0a38/tests/dags/providers/test_factory_utils.py#L97-L116

tests/dags/providers/provider_api_scripts/test_metropolitan_museum.py
```diff
@@ -6,181 +6,164 @@
...
+    if expected is None:
+        assert actual is None
+    else:
+        assert len(actual) == len(expected)
+        for i in range(len(actual)):
+            for key, value in expected[i].items():
+                if key == "license_info":
+                    assert actual[i].get(key) == CC0
+                else:
+                    assert actual[i].get(key) == value
```
AetherUnbound, line 169: Here's another way to approach this without doing list indexing using [`zip`](https://docs.python.org/3.3/library/functions.html#zip)!

```suggestion
        for actual_result, expected_result in zip(actual, expected):
            for key, value in expected_result.items():
                if key == "license_info":
                    assert actual_result.get(key) == CC0
                else:
                    assert actual_result.get(key) == value
```

AetherUnbound (commented):
This is fantastic, thank you for tackling it!

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
rwidom: But it actually does make a noticeable difference! ~7%!

```
(openverse) % python3 -m timeit '[n in [1, 2] for n in range(4)]'
1000000 loops, best of 5: 390 nsec per loop
(openverse) % python3 -m timeit '[n in {1, 2} for n in range(4)]'
1000000 loops, best of 5: 360 nsec per loop
```

tests/dags/providers/provider_api_scripts/resources/metropolitan_museum_of_art/read_met_csv.py
rwidom: Thanks so much!

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
```diff
@@ -6,138 +6,167 @@
...
+
+class MetMuseumDataIngester(ProviderDataIngester):
+    providers = {"image": prov.MET_MUSEUM_DEFAULT_PROVIDER}
+    endpoint = "https://collectionapi.metmuseum.org/public/collection/v1/objects"
+    DEFAULT_LICENSE_INFO = get_license_info(license_="cc0", license_version="1.0")
+
+    def __init__(self, date: str = None):
+        super(MetMuseumDataIngester, self).__init__(date=date)
+        self.retries = 5
+        self.delayed_requester._DELAY = 3
```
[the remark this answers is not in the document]
rwidom: Totally. Done.

tests/dags/providers/provider_api_scripts/test_metropolitan_museum.py
rwidom: I was trying to get the test case description to appear in the results if the test failed, but I think comments would get dropped in the abbreviation. I'll try it, or is there a better way to accomplish that? Maybe it's just making the values themselves describe the case, where there is a string value... ??

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
```diff
@@ -6,138 +6,167 @@
...
+                object_json.get("period"),
+            ]
+            if tag
+        ]
+        if object_json.get("tags"):
+            tag_list += [tag["term"] for tag in object_json.get("tags")]
+        return tag_list
+
+    def _get_title(self, record):
+        return record.get("title", record.get("objectName"))
```
rwidom, line 149: @AetherUnbound , I should have tested before committing the suggestion here. It looks like this really only works for `None`, where `""` evaluates to `False` as well. :/ Some of the API responses have all of the fields, but then an empty string, and I want to be able to handle that situation.

openverse_catalog/dags/providers/provider_workflows.py
```diff
@@ -116,7 +117,10 @@ def __post_init__(self):
     ),
     ProviderWorkflow(
         provider_script="metropolitan_museum",
-        start_date=datetime(2016, 9, 1),
+        ingester_class=MetMuseumDataIngester,
+        # changed for testing
+        # start_date=datetime(2016, 9, 1),
```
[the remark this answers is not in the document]
rwidom: Done.

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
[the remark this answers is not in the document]
AetherUnbound: Sweet, I made https://github.com/WordPress/openverse-catalog/issues/687 as well!

tests/dags/providers/provider_api_scripts/test_metropolitan_museum.py
AetherUnbound: Ah! If you're looking to edit the test IDs specifically and how they show up in the node names during testing, here's a few ways to do that: https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
AetherUnbound: Oh! That totally makes sense, you could potentially do `return record.get("title") or record.get("objectName")`. I was going to suggest that at first but thought that using the default might be better 😅 The `or` syntax will appropriately handle empty strings I believe!

tests/dags/providers/provider_api_scripts/test_metropolitan_museum.py
rwidom: Thanks so much!

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
rwidom: Yes! Thank you.

AetherUnbound (commented):
The new changes look great! I tried running this locally (with the same altered start date you originally committed) but every run I attempt has skipped 🤔 Does that happen on your end? This _should_ be pulling at least a little bit of data every day.

rwidom:
> The new changes look great! I tried running this locally (with the same altered start date you originally committed) but every run I attempt has skipped 🤔 Does that happen on your end? This _should_ be pulling at least a little bit of data every day.

Yes. It wasn't happening before, but things got weird after I merged in the changes from main. See https://github.com/WordPress/openverse-catalog/pull/674/commits/8bec026c591cb49afe0e2112d35553310db5bee6 which is there to fix an issue with the file name generation step. I'm going to put back the delay and the more recent start date and see if I can figure out what's going on.

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
```diff
@@ -6,138 +6,167 @@
...
-            creator=object_json.get("artistDisplayName"),
-            title=object_json.get("title"),
-            meta_data=meta_data,
+class MetMuseumDataIngester(ProviderDataIngester):
+    providers = {"image": prov.MET_MUSEUM_DEFAULT_PROVIDER}
+    endpoint = "https://collectionapi.metmuseum.org/public/collection/v1/objects"
+    DEFAULT_LICENSE_INFO = get_license_info(license_="cc0", license_version="1.0")
+
+    # adding args for automatically generated parameters from generate_tsv_filenames
+    def __init__(self, date: str = None, *args):
```
rwidom, line 48: Started seeing this when testing locally and moved this back to draft. based on [this](https://airflow.apache.org/docs/apache-airflow/stable/faq.html#what-s-the-deal-with-start-date), I think `datetime.datetime` might be better than `str` here, but I'm not sure if a one-of fix is better, or if I could be doing something more coordinated.

```
Broken DAG: [/usr/local/airflow/openverse_catalog/dags/providers/provider_workflow_dag_factory.py] Traceback (most recent call last): File "/usr/local/airflow/openverse_catalog/dags/providers/provider_dag_factory.py", line 165, in create_provider_api_workflow dag = DAG( File "/usr/local/airflow/.local/lib/python3.10/site-packages/airflow/models/dag.py", line 388, in __init__ if start_date and start_date.tzinfo: AttributeError: 'str' object has no attribute 'tzinfo'
```

stacimc: I'm not sure why you're seeing that error, but it looks like that's referring to the `start_date` for the DAG, rather than _this_ date, which is the date for ingestion.

The DAG's `start_date` is [defined in the provider_workflow](https://github.com/WordPress/openverse-catalog/pull/674/files#diff-c9fe199d287d2190c2b675f6056f2bd047894744aa2c891827f8b88a069c602bL135) as a datetime and looks good to me.

The `date` that we get here is generated [here](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_dag_factory.py#L185) as a string (and by default will be today's date). It needs to be a string because it's what we're passing to the provider API in query params.

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
```diff
@@ -6,138 +6,167 @@
...
+        super(MetMuseumDataIngester, self).__init__(date=date)
+        self.retries = 5
+
+        self.query_param = None
+        if self.date:
+            self.query_param = {"metadataDate": date}
```
stacimc, line 54:
```suggestion
        if date:
            self.query_param = {"metadataDate": date}
```

`self.date` has not been set, so these query params won't get set.

openverse_catalog/dags/common/loader/provider_details.py
```diff
@@ -20,6 +20,7 @@
 SMITHSONIAN_DEFAULT_PROVIDER = "smithsonian"
 BROOKLYN_DEFAULT_PROVIDER = "brooklynmuseum"
 CLEVELAND_DEFAULT_PROVIDER = "clevelandmuseum"
+MET_MUSEUM_DEFAULT_PROVIDER = "met"
```
stacimc, line 23: Nice, funny this one wasn't pulled out already! Can we use the full name `METROPOLITAN_MUSEUM_DEFAULT_PROVIDER`, just for consistency?

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
```diff
@@ -6,138 +6,167 @@
 Output:                 TSV file containing the image, their respective
                         meta-data.
 
-Notes:                  https://metmuseum.github.io/
-                        No rate limit specified.
+Notes:                  https://metmuseum.github.io/#search
+                        "Please limit requests to 80 requests per second." May need to
```
stacimc, line 10: Thank you for always adding excellent notes ✨

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
stacimc, line 48:
```suggestion
    def __init__(self, conf: dict = None, date: str = None, *args):
```

The `conf` was added to the base class in #650, so we'll need to add it here and in the call to `super` below. Sorry, lots of changes happening at once 😄

rwidom: Thank you so much! It was a total typo in the provider workflow. Now I just need to figure out why it's skipping.

rwidom: Got it! Yes, this helps a lot.

openverse_catalog/dags/common/loader/provider_details.py
rwidom: Done.

rwidom:
> The new changes look great! I tried running this locally (with the same altered start date you originally committed) but every run I attempt has skipped 🤔 Does that happen on your end? This _should_ be pulling at least a little bit of data every day.

Big thanks to @stacimc, this seems to be all set now. I just needed to properly add `conf` parameters from https://github.com/WordPress/openverse-catalog/pull/650

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
```diff
@@ -6,138 +6,167 @@
...
+                object_json.get("classification"),
+                object_json.get("objectDate"),
+                object_json.get("creditLine"),
+                object_json.get("period"),
+            ]
+            if tag
+        ]
+        if object_json.get("tags"):
+            tag_list += [tag["term"] for tag in object_json.get("tags")]
+        return tag_list
```
stacimc, line 149: I haven't looked into this very far, but when I ran the DAG locally and inspected the generated tsv in MinIO I see the `raw_tags` all look like this:

```
[{"name": "Asian Art", "provider": "met"}, {"name": "Gilt bronze", "provider": "met"}, {"name": "Korea", "provider": "met"}, {"name": "Figure", "provider": "met"}, {"name": "Sculpture", "provider": "met"}, {"name": "mid-7th century", "provider": "met"}, {"name": "Purchase, Walter and Leonore Annenberg and The Annenberg Foundation Gift, 2003", "provider": "met"}, {"name": "Three Kingdoms period (57 B.C.–A.D. 676)", "provider": "met"}, {"name": "Contemplation", "provider": "met"}, {"name": "Buddhism", "provider": "met"}, {"name": "Bodhisattvas", "provider": "met"}]
```

Any idea what's going on there?

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
```diff
@@ -6,138 +6,167 @@
...
-            title=object_json.get("title"),
-            meta_data=meta_data,
+class MetMuseumDataIngester(ProviderDataIngester):
+    providers = {"image": prov.METROPOLITAN_MUSEUM_DEFAULT_PROVIDER}
+    endpoint = "https://collectionapi.metmuseum.org/public/collection/v1/objects"
+    DEFAULT_LICENSE_INFO = get_license_info(license_="cc0", license_version="1.0")
+
+    # adding args for automatically generated parameters from generate_tsv_filenames
+    def __init__(self, conf: dict = None, date: str = None):
+        super(MetMuseumDataIngester, self).__init__(conf=conf, date=date)
```
AetherUnbound, line 49: FWIW in python 3 this can just be:
```suggestion
        super().__init__(conf=conf, date=date)
```

We don't need to specify the parent class unless the [Module Resolution Order](https://www.educative.io/answers/what-is-mro-in-python) needs to be specifically navigated:

> The parameterless call to super() is recommended and sufficient for most use cases, and needing to change the search hierarchy regularly could be indicative of a larger design issue. [^1]

[^1]: https://realpython.com/python-super/#a-super-deep-dive

AetherUnbound (approved):
Ran this locally with success! Looking great! Only one extra comment.

stacimc (changes requested):
This is looking so good! 🥳 Just a few notes, mostly the unexpected behavior with the tags.

I love the improvements to the data quality, although I'm a tiny bit hesitant to make too many extra changes while refactoring for fear of having to revert all of it if something unexpected happens. Since Metropolitan is `dated`, we should make a note that the changes won't be backfilled until reingestion workflows are up again. I think it should be fine in this case though 👍 

I'd just like to confirm a couple of things, maybe with some of the more frontend-knowledgeable folks:
* Is it okay to remove the information that was previously in `meta_data` and move it to `raw_tags`? I'm not 100% sure how that metadata was being used.
* Are there any concerns with the number and type of tags being added? I'd imagine the more relevant tags the merrier, but I'm pretty sure this would be a big change so may be worth mentioning.

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
stacimc: AH, my mistake. I was expecting a flat list, so `["Asian Art", "Gilt bronze", "Korea", "Figure", "Sculpture"...]` But I'm of course thinking of the _raw_ tags, not the enriched tags needed for the load step. This is working totally as expected, ignore me @rwidom!

stacimc:
@rwidom I take back the only blocking suggestion I had, as it was a misunderstanding on my part. I'm just going to try to verify my two questions about moving tags out of `meta_data` and populating a large list, and then this will LGTM! Thanks for your patience as always, I would be so excited to get these improvements in ✨

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
```diff
@@ -6,138 +6,167 @@
...
+        self.object_ids_retrieved = 0  # total object IDs based on date
+        self.non_cc0_objects = 0  # number checked and ignored because of licensing
+
+    def get_next_query_params(self, prev_query_params=None):
+        return self.query_param
+
+    def get_batch_data(self, response_json):
+        if response_json:
+            self.object_ids_retrieved = response_json["total"]
+            logger.info(f"Total objects found {self.object_ids_retrieved}")
```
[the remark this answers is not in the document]
rwidom: The met museum returns a json with two things: a list of object ids, and the length of that list. The additional API calls are just to get the details about any given object ID, not to get another list of object IDS. So, I think it's ok to just take this from the source once for each dag run.

rwidom:
> I'd just like to confirm a couple of things, maybe with some of the more frontend-knowledgeable folks:
> 
>     * Is it okay to remove the information that was previously in `meta_data` and move it to `raw_tags`? I'm not 100% sure how that metadata was being used.
> 
>     * Are there any concerns with the number and type of tags being added? I'd certainly _imagine_ the more relevant tags the merrier, but I'm pretty sure this would be a big change so may be worth mentioning. 🤞

Yes, absolutely. I would love to know the answers here as well. I somehow got the impression that metadata does not make it to elastic search, that title and tags are the default information in any search, but I could be totally wrong and it's definitely something I want to learn more about, @stacimc .

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
```diff
@@ -6,138 +6,167 @@
...
+        if object_json.get("tags"):
+            tag_list += [tag["term"] for tag in object_json.get("tags")]
+        return tag_list
+
+    def _get_title(self, record):
+        if record is not None:
+            # Use or to skip false-y (empty) titles: ""
+            return record.get("title") or record.get("objectName")
+
+    def _get_artist_name(self, record):
```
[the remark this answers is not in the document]
rwidom: Yes, I just tried to clarify. In the process it occurred to me that we could consider each object a separate "batch", but my understanding is that there are a bunch that only have one record and I don't know what I would call the initial list of objects then, so I'm leaving it as is otherwise.

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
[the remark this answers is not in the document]
stacimc: But this is reporting once for every _batch_, indicating the number of records in that batch as opposed to the total number ingested by the run. I see it multiple times in the logs for a particular dagrun. Is that the intention, or am I wrong about what that number means?

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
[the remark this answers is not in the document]
stacimc: Oh wow -- 'object' vs 'record' in this context _is_ confusing, I can see what you mean. Thanks for adding the comment, that's very clear now!

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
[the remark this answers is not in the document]
rwidom: Oh... Maybe I'm confused about what a batch means and how it's functioning here. My sense is that there should only be one batch per dag-run. I'll take another look.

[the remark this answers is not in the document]
rwidom: Hopefully this enforces the one batch per dagrun thing! https://github.com/WordPress/openverse-catalog/pull/674/commits/3f7765ebc23bae24499b4ae6caecea2e41b9a181

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
```diff
@@ -6,138 +6,167 @@
...
+                "raw_tags": raw_tags,
+            }
+            for img in image_list
+        ]
+
+    def _get_foreign_id(self, object_id: int, image_url: str):
+        unique_identifier = image_url.split("/")[-1].split(".")[0]
+        return f"{object_id}-{unique_identifier}"
+
+    def _get_meta_data(self, object_json):
```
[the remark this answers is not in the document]
rwidom: Huh, what are the benefits of making that declaration? The main one I'm familiar with is being able to call them from outside a specific instance of the class. Would that make testing more efficient in this case? Are there other benefits I'm not thinking of?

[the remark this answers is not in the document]
AetherUnbound: It's usually a good practice to make functions that you know won't need to access or modify the class's state via `self`, `staticmethod`s. This makes the scope clear to folks reading the function, can make testing easier in certain cases (since you don't need to instantiate a class in order to test that specific function), and prevents cases where a function that _seems_ like it should be static but is not explicitly set as such has access to `self` and can thus modify class state (either intentionally or unintentionally).

AetherUnbound:
> Is it okay to remove the information that was previously in `meta_data` and move it to `raw_tags`? I'm not 100% sure how that metadata was being used.

Looking at the infrastructure repo, it doesn't seem like we do any indexing on the `meta_data` and AFAICT the only thing it's used for is the authority boost:

```bash
$ rg -g "*.py" meta_data
ingestion_server/authority.py
12:The authority can be set from the catalog layer through the meta_data field

ingestion_server/elasticsearch_models.py
93:        meta = row[schema["meta_data"]]
135:    def get_license_url(meta_data):
140:        if meta_data and "license_url" in meta_data:
141:            return meta_data["license_url"]
146:    def get_maturity(meta_data, api_maturity_flag):
149:        :param meta_data: The metadata column, which may have a 'mature'
153:        we will ignore the meta_data column and mark the work 'mature'.
157:        if meta_data and "mature" in meta_data:
158:            _mature = meta_data["mature"]
164:    def get_authority_boost(meta_data, source):
166:        if meta_data and "authority_boost" in meta_data:
168:                authority_boost = float(meta_data["authority_boost"])
241:        meta = row[schema["meta_data"]]
323:        meta = row[schema["meta_data"]]
```

There are **zero** references to `authority_boost` in the catalog repo, so I suspect that's not something we're using at all currently. I think moving that info into the tags, if you believe it would help with search relevancy, is a good idea 🙂

stacimc (approved):
As far as I can _tell_, the changes from `meta_data` to `tags` should be okay -- and thanks @AetherUnbound for checking as well. Very exciting!

We'll keep an eye on this to be safe, but since this is a dated DAG and reingestion is not yet enabled, it will naturally roll out only to new data for now. 🤞 I hope this is the beginning of much needed relevancy improvements across more of our providers, thanks @rwidom!

PR #678, opened 2022-08-18
Standardize on datetime over pendulum

AetherUnbound:
## Fixes

This addresses a comment left by @sarayourfriend in https://github.com/WordPress/openverse-catalog/pull/632#discussion_r934777831

## Description

This PR standardizes all uses of a `datetime` or `datetime` related object/class on the standard library's `datetime` module rather than `pendulum`. Pendulum is a library that Airflow uses internally, but we don't specify it explicitly in our requirements. Additionally, the related `pendulum` analogs to `datetime` are actually subclasses of `datetime` objects, so it's perfectly safe to use the base classes everywhere as well.

## Testing Instructions

`just test` should be sufficient!

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/tests/dags/common/operators/test_postgres_result.py
+++ b/tests/dags/common/operators/test_postgres_result.py
@@ -5,7 +5,6 @@ https://airflow.apache.org/docs/apache-airflow/stable/best-practices.html#unit-t
 import datetime
 import os
 
-import pendulum
 import pytest
 from airflow import DAG
 from airflow.models import DagRun, TaskInstance
@@ -15,7 +14,7 @@ from airflow.utils.types import DagRunType
 from common.operators.postgres_result import PostgresResultOperator
 
 
-DATA_INTERVAL_START = pendulum.datetime(2021, 9, 13, tz="UTC")
+DATA_INTERVAL_START = datetime.datetime(2021, 9, 13, tzinfo=datetime.timezone.utc)
 DATA_INTERVAL_END = DATA_INTERVAL_START + datetime.timedelta(days=1)
 
 TEST_DAG_ID = "test_postgres_result_dag"
--- a/openverse_catalog/dags/providers/factory_utils.py
+++ b/openverse_catalog/dags/providers/factory_utils.py
@@ -1,6 +1,7 @@
 import inspect
 import logging
 import time
+from datetime import datetime
 from types import FunctionType
 from typing import Callable, Sequence
 
@@ -8,7 +9,6 @@ from airflow.models import DagRun, TaskInstance
 from airflow.utils.dates import cron_presets
 from common.constants import MediaType
 from common.storage.media import MediaStore
-from pendulum import datetime
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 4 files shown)


sarayourfriend (commented):
Nice :slightly_smiling_face: Looks like we were barely using pendulum anyway :sweat_smile:

AetherUnbound:
Yea, I was just trying to conform with the Airflow standard on a few more recent PRs 😅

AetherUnbound:
Ah shoot, it looks like we actually do need pendulum in some places:
```python
self = UtcDateTime(timezone=True)
value = datetime.datetime(2022, 8, 18, 21, 22, 22, 316222)
dialect = <sqlalchemy.dialects.postgresql.psycopg2.PGDialect_psycopg2 object at 0x7fc297b5b8e0>

    def process_bind_param(self, value, dialect):
        if value is not None:
            if not isinstance(value, datetime.datetime):
                raise TypeError('expected datetime.datetime, not ' + repr(value))
            elif value.tzinfo is None:
>               raise ValueError('naive datetime is disallowed')
```

I'll update for this specific case and add a comment.

sarayourfriend:
If we do need to keep it around can we add it to the requirements?

sarayourfriend:
If Airflow and the community use Pendulum I do think that's a pretty decent argument for using it. Especially if there are particular pitfall edge cases where Pendulum is required. It's a lot less mental overhead to just use one, either one, than get frustrated running into the edge cases suddenly (IMO). Especially if Pendulum is duck compatible with datetime.datetime for usages with libraries that don't know about pendulum.

obulat (approved):
LGTM!

AetherUnbound:
> Especially if there are particular pitfall edge cases where Pendulum is required.

Pendulum is only required in a [test that explicitly runs an Airflow DAG directly](https://github.com/WordPress/openverse-catalog/pull/678/commits/676c48d5a6a6a418e34ce5f2375f02907253de46#diff-176b6052f4f4d99c946d6cd87df291c47af669878344996e7a49829c495bb446R68), which is not an step we'd perform as part of our normal operational code. Furthermore, while pendulum can serve as a drop-in replacement for `datetime.datetime`, there are [code changes necessary for using pendulum's `duration` over `datetime.timedelta`](https://pendulum.eustace.io/docs/#interval). We use `timedelta` a substantial amount, so this would incur a non-trivial code change:

```bash
# Number of uses of datetime.timedelta
$ rg -g "*.py" timedelta | wc -l
115
```

This PR constitutes a smaller change than standardizing on pendulum, and would likely lead to greater consistency with the wider Python contributor community (who are more likely to use the standard library's `datetime` over an optimized `pendulum`).

tests/dags/common/operators/test_postgres_result.py
```diff
@@ -5,7 +5,6 @@
 import datetime
 import os
 
-import pendulum
```
krysal, line 8: Nice to keep only one time library in this file, and standardize on `datetime` in general.

krysal (approved):

dhruvkb (approved):
Looks good to me. It's nice that we don't need to use an indirect dependency.

Issue #209, opened 2021-10-04
Move API keys from .env to Airflow Variables


AetherUnbound:
## Current Situation

Provider API keys look to be set in the [.env file](./openverse_catalog/env.template).

## Suggested Improvement

These keys could be moved into Airflow Variables instead.

## Benefit

* As we expand our list of providers, the current strategy would require that we also bloat the environment with each of the new API keys.
* Airflow encrypts and hides Variables when they contain secrets.
* API keys can be edited without having to restart the service.

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

zackkrida:
Excellent suggestion!

AetherUnbound:
We've got a few options here, but my TL;DR for the time being is: keep things as is.

## Some potential approaches:
### Make Variables easy to access & edit

In the past, I've used Variables and Connections pretty heavily for API keys, secrets, etc. It's easy to use/update in the UI, provided that we have the "hide secrets" Airflow setting turned **off**. If that setting is on, it makes it much harder for engineers to introspect the keys and change things on the fly. That said, in my previous positions we also didn't have as many API keys to manage as we're likely to have here. This method also requires some manner of populating the Variables & Connections. Currently we're using env vars for this, which kind of defeats the purpose of _also_ putting them in Variables. A Variable means a call to the database, whereas an env var retrieval is bound to be much faster. Variables do allow us to change these values during runtime.

**Pros**:
- Easy for engineers to access & update API keys.
- Updates take affect immediately and apply to the next DAG run for a given service.
- Environment is not polluted with an ever-increasing number of API keys.

**Cons**:
- Values will need to be hand entered or have a programmatic creation in the `init` part of the entrypoint (work described in #227).
- In order to make this easiest for engineers, API keys would be exposed directly in the UI. This could be mitigated by creating multiple Airflow user roles with different permissions sets.
- Requires a DB call to retrieve the Variable. Particularly if this is added at the import-level along with everything else (#229), this will cause a significant slowdown in DAG processing time.

### Only use environment variables

This would mean all API keys end up in our environment somehow (either via `.env` or otherwise). With this approach, I think it might be best to standardize the names as `API_KEY_<provider>` so they all appear in the same location within the environment. This also opens us up for using an external secrets manager (like [AWS's option](https://aws.amazon.com/secrets-manager/)) to store those keys. The difficulty with the env vars is that we would have to restart the service/worker any time a key changes.

**Pros**:
- More secure.
[...]

PR #683, opened 2022-08-18
Add Openverse email to DAG default args

AetherUnbound:
## Description

This PR adds the `CONTACT_EMAIL` (presently openverse@wordpress.org) to the `email` section of the DAG default args. This is a precursor to getting alert emails working, since Airflow first needs an address to send said emails to.

## Testing Instructions

You should be able to go to the "Details" page of any DAG locally and see that `email` is now defined under the "Default Args" section.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/constants.py
+++ b/openverse_catalog/dags/common/constants.py
@@ -11,11 +11,14 @@ MEDIA_TYPES = [AUDIO, IMAGE]
 
 MediaType = Literal["audio", "image"]
 
+CONTACT_EMAIL = os.getenv("CONTACT_EMAIL")
+
 DAG_DEFAULT_ARGS = {
     "owner": "data-eng-admin",
     "depends_on_past": False,
     "start_date": datetime(2019, 1, 15),
     "email_on_retry": False,
+    "email": [CONTACT_EMAIL],
     "retries": 2,
     "retry_delay": timedelta(minutes=5),
     "execution_timeout": timedelta(hours=1),


zackkrida (approved):

sarayourfriend (approved):

sarayourfriend:
@AetherUnbound are Airflow variables available there? If so, what's the trade off to use an environment variable instead of the airflow variable? Do environment variables require a redeployment with the catalogue as well?

AetherUnbound:
Yes, environment variables would require a deployment to change whereas Airflow Variables can be changed on the fly. We had a longer discussion about this which you can see here: #209 

The environment variable I used here already existed and was in use elsewhere, so I just opted to use it here 🙂 we could just as easily hard-code it or use an Airflow Variable. I think the advantage to using an env var is that you don't have to have the Airflow metastore available when running a script directly. This is used in the [`provider_details.py` file](https://github.com/WordPress/openverse-catalog/blob/a2e1d50ff3c7b4c131a77313e11db63201b5c701/openverse_catalog/dags/common/loader/provider_details.py#L112), which _could_ be run outside of the Docker/Airflow setup.

sarayourfriend:
Gotcha.

> This is used in the [provider_details.py file](https://github.com/WordPress/openverse-catalog/blob/a2e1d50ff3c7b4c131a77313e11db63201b5c701/openverse_catalog/dags/common/loader/provider_details.py#L112), which could be run outside of the Docker/Airflow setup

What is the use case for running this or any other script outside of the deployed Airflow?

AetherUnbound:
Many of the provider scripts were (seemingly) originally set up to be able to run on their own, inside a Python virtual environment without interacting with the Airflow infrastructure. I presume this was so that they could be run outside the context of a DAG. More and more we're tying these provider scripts into Airflow infrastructure, so I don't believe it will be as relevant or necessary going forward.

However, Airflow Variables are useful for values we foresee changing frequently, or values where we wouldn't want to have to issue a deployment in order to change. Since this value already existed in the environment (so no deployment is necessary) and we're unlikely to change the value in the near future, I think it's fine as-is.

PR #694, opened 2022-08-24
Always record provider run duration

### LINKED ISSUE #692: Use try/finally for provider run duration
## Problem

The duration calculation step for provider API scripts runs right after the run func is completed, which means that it is _not_ computed for runs that fail:

https://github.com/WordPress/openverse-catalog/blob/83c688d8354bd663a22835864f7aa18e3bdc9286/openverse_catalog/dags/providers/factory_utils.py#L136-L140

## Description

Since we now know that `try`/`finally` will run in the event of an Airflow failure as well, we should be able to wrap the run func in a `try`/`finally` and perform the duration calculation & xcom-pushing within the `finally` so that it gets run even in the event of failures.

This should be a fairly small change! The hardest part will probably be adding appropriate tests 🙂 

## Implementation

AetherUnbound:
## Fixes

Fixes #692 by @AetherUnbound

## Description

This PR wraps the run function in a try/finally block so that the duration is always recorded & pushed. I'm not removing the machinery which expects that duration _could_ be `None` because there may still be circumstances where Airflow kills things even in the `finally` block or if the error occurs before that block.

## Testing Instructions

`just test`

Optionally, you could start a provider DAG and fail the data pull task after it starts actually saving data to disk. When viewing the logs for the `report_load_completion` task, the crafted message should still show a duration even though the upstream task failed.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/factory_utils.py
+++ b/openverse_catalog/dags/providers/factory_utils.py
@@ -136,11 +136,13 @@ def pull_media_wrapper(
     start_time = time.perf_counter()
     # Not passing kwargs here because Airflow throws a bunch of stuff in there that
     # none of our provider scripts are expecting.
-    data = run_func(*args)
-    end_time = time.perf_counter()
-    # Report duration
-    duration = end_time - start_time
-    ti.xcom_push(key="duration", value=duration)
+    try:
+        data = run_func(*args)
+    finally:
+        end_time = time.perf_counter()
+        # Report duration
+        duration = end_time - start_time
+        ti.xcom_push(key="duration", value=duration)
     return data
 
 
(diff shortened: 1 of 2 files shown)


stacimc (approved):
LGTM! Tested with a successful run, a run with an error manually raised during `pull_data`, and by manually marking the task as 'failed' in Airflow while it was running.

This is very nice and we've needed it for awhile! Excited to get duration data after manually pass/failing tasks now ✨

krysal (approved):
LGTM!

PR #706, opened 2022-08-31
Always use Jamendo's "streaming" audio

### LINKED ISSUE #667: Use new format for Jamendo urls
## Description

On the frontend we use the `url` to play audio previews. Currently, our urls from Jamendo have the format "https://prod-1.storage.jamendo.com/download/track/121465/mp32, where the track id and format are part of the path.

Jamendo is now recommending that we use a different endpoint using query params instead:"https://prod-1.storage.jamendo.com/?trackid=1279226&format=mp32"

Because the old format was causing errors on the frontend (https://github.com/WordPress/openverse-frontend/issues/1605), we've already updated all Jamendo records in place in the DB. **For this issue, we should update the Jamendo provider script to set the urls correctly during ingestion.** This will ensure that records ingested in the future don't cause the same issues.

When this is complete, we can turn the Jamendo workflow back on in production.

## Resolution

AetherUnbound:
## Fixes

Fixes #667 by @stacimc

## Description

This PR alters the logic for determining which URL to use for Jamendo audio results. Ingestions now always use the "streaming" URL rather than the "download" URL. Our primary use-case for the `url` column presently is to provide an audio file that the frontend can stream as a preview. Preferring the download URL was causing some issues on the frontend, see https://github.com/WordPress/openverse-frontend/issues/1605. 

As part of the API response we get back a `from` parameter that looks like the API or some other encoded information. Since we're already stripping `trackid` from a URL in another portion of the provider script, I opted to consolidate this logic and keep the original `_remove_<field>` functions around as a means for holding documentation specific to each action.

## Testing Instructions

If you undo the changes on the test file and run the tests, you can see in the failure diff that the provider script is now pulling out the preferred audio.

For more thorough testing:
1. `just recreate`
1. Enable the Jamendo DAG
1. After a minute or so, mark the `pull_audio_data` task as successful (this is not a class-based ingester yet! see #587)
1. Run `just db-shell`
1. Run `SELECT url FROM audio LIMIT 10;`

You should see results in the following format:
```
openledger> select url from audio limit 10;
+--------------------------------------------------------------+
| url                                                          |
|--------------------------------------------------------------|
| https://prod-1.storage.jamendo.com/?trackid=1091&format=mp32 |
| https://prod-1.storage.jamendo.com/?trackid=1092&format=mp32 |
| https://prod-1.storage.jamendo.com/?trackid=1093&format=mp32 |
| https://prod-1.storage.jamendo.com/?trackid=1094&format=mp32 |
| https://prod-1.storage.jamendo.com/?trackid=1095&format=mp32 |
| https://prod-1.storage.jamendo.com/?trackid=1096&format=mp32 |
| https://prod-1.storage.jamendo.com/?trackid=1097&format=mp32 |
| https://prod-1.storage.jamendo.com/?trackid=1098&format=mp32 |
| https://prod-1.storage.jamendo.com/?trackid=1099&format=mp32 |
| https://prod-1.storage.jamendo.com/?trackid=1137&format=mp32 |
+--------------------------------------------------------------+
```

In order to make sure they were parsed correctly, click on one and listen :smiling:

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/jamendo.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/jamendo.py
@@ -130,7 +130,7 @@ def _extract_audio_data(media_data):
         foreign_landing_url = media_data["shareurl"]
     except (TypeError, KeyError, AttributeError):
         return None
-    audio_url, duration, download_url = _get_audio_info(media_data)
+    audio_url, duration = _get_audio_info(media_data)
     if audio_url is None:
         return None
     item_license = _get_license(media_data)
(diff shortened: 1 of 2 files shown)


openverse_catalog/dags/providers/provider_api_scripts/jamendo.py
```diff
@@ -184,27 +184,50 @@ def _get_foreign_identifier(media_data):
...
+    """
+    if audio_url is None:
+        return
+    return _remove_param_from_url(audio_url, "from")
+
+
+def _get_audio_info(media_data: dict) -> tuple[Optional[str], int]:
     """Parses audio URL, audio download URL, audio duration
     If the audio does not allow download, we save the 'streaming'
     URL as the `audio_url`
```
stacimc, line 218: This part of the docstring should be updated.

Optionally, it seems reasonable to me to have the documentation for removing the 'from' parameter here and remove the `_remove_from` function, since `_get_audio_info` is itself very simple now. Total bikeshedding, but `_remove_from` looks confusing to me, I think because 'from' is a preposition and the documentation is in a different function.

`_remove_trackid` makes more sense to me and I'd be fine keeping it even if you do decide to remove `_remove_from`. Up to you though!

stacimc (changes requested):
Works great! I think just the one docstring update and then this looks fantastic. Removing the `from` parameter seems like a good idea 👍

openverse_catalog/dags/providers/provider_api_scripts/jamendo.py
```diff
@@ -184,27 +184,42 @@ def _get_foreign_identifier(media_data):
...
 
 
-def _get_audio_info(media_data):
-    """Parses audio URL, audio download URL, audio duration
-    If the audio does not allow download, we save the 'streaming'
-    URL as the `audio_url`
+def _remove_param_from_url(url: str, param: str) -> str:
+    """
+    Remove a parameter from a provided URL.
+    """
```
krysal, line 190: This function looks potentially useful for many providers!

AetherUnbound: That's a good point! If any of us comes across another scenario where this is needed, we should make it a general utility function 🙂

PR #713, opened 2022-09-02
Remove error swallowing during ingestion

### LINKED ISSUE #703: Determine provider discrepancies on `get_response_json` try/catch
## Description

As part of #555 and #645, we unified the provider ingester script query mechanisms and error handling. In the process however, it looks like we may have introduced error handling/swallowing where there previously was none:

https://github.com/WordPress/openverse-catalog/blob/eec83ac6970dd73cb5e5efec140c7a0f540c74f1/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py#L296-L306

As written, this will catch every failure that could reasonably happen from `get_response_json`, swallow it, and continue execution. For **most** cases, this is neither ideal or appropriate. However for some cases (e.g. [Brooklyn Museum](https://github.com/WordPress/openverse-catalog/pull/645/files#diff-7c4ffca320da8f2f4e9382b7a90945e5cad454dd4c633168ba9f79e4312e7da3)), this was the intended behavior. It seems that this probably _shouldn't_ be the case though, since #650 gave us the capacity to skip ingestion errors that occur within a batch. We can rely on this method for catching errors rather than adding an explicit try/catch here.

This will involve altering tests, as some providers expect errors to result in continued execution.

**In summary**, we want to change this to halt execution on errors that occur here, since we have higher-order mechanisms for skipping/handling errors in processing.

## Resolution

stacimc:
## Fixes

Fixes #703 by @AetherUnbound 

## Description

TL;DR: this PR removes error swallowing during `get_batch`, which I believe was only necessary in _legacy code_ in provider scripts that were not making use of updates to the `DelayedRequester`. It also updates some tests.

### Background

When the `ProviderDataIngester` was first implemented, I included error handling that I noticed in many of our provider scripts during the `get_batch` logic, and in particular in Cleveland (which was the first provider I happened to refactor). Cleveland, in fact, explicitly tests that errors/None responses are caught.

After auditing our provider scripts to understand why that handling was there in the first place, I believe all of this can be removed. The try/except is (or was, before refactoring) present in exactly half of our provider scripts. For the vast majority of these, I believe this is legacy code that was written before updates were made to the `DelayedRequester`. Here's a good example in [rawpixel](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_api_scripts/rawpixel.py#L28). The script uses `delayed_requester.get`, and then checks the status code and parsing the JSON within a try/except. This logic, however, is now available in [delayed_request.get_response_json](https://github.com/WordPress/openverse-catalog/blob/d828d257adc70121d0b2d8b926e3d625f700651c/openverse_catalog/dags/common/requester.py#L97), along with handling retries.

The `ProviderDataIngester` is already using `get_response_json` and can rely on its handling.

### Notes about Cleveland

Cleveland had explicit tests to ensure that errors/None responses from the API were caught. With the removal of the error swallowing, this is no longer true. I've decided this is the correct behavior and removed the erroneous tests. **On review of past runs of the Cleveland DAG, this situation has not arisen, so I do not believe this is something we _have_ to catch (for example because of flakiness).** If it becomes an issue in the future we can revisit it, and for any particular DagRun we can use `skip_ingestion_errors` to accomplish the same thing.

## Testing Instructions

`just test`, and confirm that you agree with my reasoning above 😄 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
@@ -6,10 +6,9 @@ from typing import Dict, List, Optional, Tuple
 
 from airflow.exceptions import AirflowException
 from airflow.models import Variable
-from common.requester import DelayedRequester, RetriesExceeded
+from common.requester import DelayedRequester
 from common.storage.media import MediaStore
 from common.storage.util import get_media_store_class
-from requests.exceptions import JSONDecodeError, RequestException
 
 
 logger = logging.getLogger(__name__)
@@ -298,17 +297,7 @@ class ProviderDataIngester(ABC):
         should_continue = True
 
         # Get the API response
-        try:
-            response_json = self.get_response_json(query_params)
-        except (
-            RequestException,
-            RetriesExceeded,
-            JSONDecodeError,
-            ValueError,
-            TypeError,
-        ) as e:
-            logger.error(f"Error getting response due to {e}")
-            response_json = None
+        response_json = self.get_response_json(query_params)
 
         # Build a list of records from the response
         batch = self.get_batch_data(response_json)
--- a/tests/dags/providers/provider_api_scripts/test_cleveland_museum.py
+++ b/tests/dags/providers/provider_api_scripts/test_cleveland_museum.py
@@ -145,27 +145,6 @@ def test_get_response_no_data():
     assert len(batch) == 0
 
 
-def test_get_response_failure():
-    query_param = {"cc": 1, "has_image": 1, "limit": 1, "skip": -1}
-    r = requests.Response()
-    r.status_code = 500
-    r.json = MagicMock(return_value={"error": ""})
-    with patch.object(clm.delayed_requester, "get", return_value=r) as mock_get:
-        batch, should_continue = clm.get_batch(query_param)
-
-    assert mock_get.call_count == 4
-    assert batch is None
-
-
-def test_get_response_None():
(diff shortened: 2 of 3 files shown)


stacimc:
For posterity, the list of providers that had some try/except logic in `get_batch`:
* Brooklyn
* Cleveland
* Europeana
* Museum Victoria
* NYPL
* Rawpixel
* Science Museum
* SMK
* Thingiverse
* WordPress

zackkrida (approved):
Great, I think your logic is sound 😅 .

krysal (approved):
Agree, the reasoning sounds solid. About that list...

> For posterity, the list of providers that had some try/except logic in get_batch:
>
>    Brooklyn
>    Cleveland
>    Europeana
>    ...

Will those individual error catching change with the DAG refactorings?

stacimc:
> Will those individual error catching change with the DAG refactorings?

@krysal Nope! Several of those providers have already been refactored. Those that _haven't_ been refactored have a variety of custom implementations of the requester logic, which I'm leaving untouched. When they get refactored to use the base class, they'll automatically get the new requester logic and there should ultimately be no difference in error handling.

AetherUnbound:
Thank you for the great explanation & context on this @stacimc! I feel confident about having removed it 😄 and it makes our code simpler!

PR #717, opened 2022-09-06
Terminate PG query when task is killed via Airflow

### LINKED ISSUE #690: Ensure PG backend query is terminated on failure
## Problem

This issue applies to tasks on Airflow that interact with postgres (e.g. `PythonOperator`s or `Postgres*Operator`s). 

Presently when a failure or execution timeout occurs, Airflow will raise an `AirflowException` within the running task. For most cases operating on the python level, this is sufficient to stop the task. In cases where the task does not stop when asked kindly, Airflow will then send a `SIGKILL` to the task and stop it.

When this happens to a task that involves postgres queries, the _python_ script running on Airflow will stop but the _postgres query_ running on the postgres backend will not. In some cases this can even cause the task to look like it's still running long after its execution timeout, because the Airflow task doesn't respond at all to the signals Airflow is sending and hangs until the query is complete (at least, to my recollection, we may need to confirm this). In cases where Airflow is able to advance, there are knock-on effects to the query still running.

**Example**: We had one instance where a load was occurring that had passed its execution timeout. I don't remember if we manually failed that task or if it failed on its own, but either way the DAG continued onto the next step which was to delete the intermediate load table. However, since the load to said intermediate table was still running on the postgres backend, the delete of the table hung until the former completed (or in this case, until we explicitly killed that query on the postgres backend). 

## Description

We should add a custom `on_failure_callback` (or similar cleanup method) to certain postgres steps that we know will be longer running which will actually terminate the postgres query on the backend upon failure. This will involve looking up the query that is being run, then issuing a `pg_terminate_backend` on the query ID.

```sql
-- Get running queries
SELECT pid, age(clock_timestamp(), query_start), usename, query 
FROM pg_stat_activity 
WHERE query != '<IDLE>' AND query NOT ILIKE '%pg_stat_activity%' 
ORDER BY query_start desc;

-- Kill query
SELECT pg_terminate_backend(<pid>);
```

## Additional context

This came up specifically in discussions around #549, since it is a longer-running DAG. However other DAGs (Finnish museums in particular) may also suffer from similar circumstances.

## Resolution

rwidom:
## Fixes
Fixes #690  by @AetherUnbound 

## Description
This summarizes a lot of discussion (and help from @stacimc and @AetherUnbound!):

- Big picture goal: make sure that airflow task execution_timeouts are enforced in the postgres database.
- Big picture design: create new postgres hook and operator that run `SET statement_timeout TO '{execution_timeout}s';` before any other SQL, so that postgres knows when to kill long-running queries.

## Testing Instructions

Switch to the branch and `just test`, and then run some dags. I ran the Cleveland Museum and iNaturalist (with small test files) locally and it went fine. 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


stacimc (commented):
This looks really great so far! The approach seems sound to me, although I agree this will be a tricky one to test.

I tried testing this by just adding a `SELECT pg_sleep(120)` to the `create_loading_table` task, and adjusting its execution timeout to 10 seconds. When I ran a provider script locally, though, the task hung for the full 120 seconds and only _then_ raised a Timeout error 🤔 Are you seeing this too?

> I wouldn't want to kill a long-running db job that just needs to run, but I would want to be able to kill other long-running things that might get in the first one's way.

Sorry, I didn't quite get this -- can you give an example of what you mean here?

rwidom:
Thanks @stacimc ! 
I really just meant "Are you _sure_ we should automatically kill all of these queries? Do we need any other filter?"
It may be that the way that airflow manages threads the `application_name` thing won't work though, because the airflow log says it executed the set application name statement, but the database isn't showing it in the query for PID. Also the task isn't marked as failed, I guess...
And maybe the answer is just "yeah, don't worry about it, kill them all!"
I'll commit the changes I made to generate these test results. Maybe I misunderstood what you did?
--R

<details>
<summary>Airflow log</summary>

```
*** Reading remote log from s3://openverse-airflow-logs/dag_id=phylopic_workflow/run_id=scheduled__2011-01-02T00:00:00+00:00/task_id=load_image_data.create_loading_table/attempt=1.log.
[2022-09-07, 23:21:56 UTC] {taskinstance.py:1179} INFO - Dependencies all met for <TaskInstance: phylopic_workflow.load_image_data.create_loading_table scheduled__2011-01-02T00:00:00+00:00 [queued]>
[2022-09-07, 23:21:56 UTC] {taskinstance.py:1179} INFO - Dependencies all met for <TaskInstance: phylopic_workflow.load_image_data.create_loading_table scheduled__2011-01-02T00:00:00+00:00 [queued]>
[2022-09-07, 23:21:56 UTC] {taskinstance.py:1376} INFO - 
--------------------------------------------------------------------------------
[2022-09-07, 23:21:56 UTC] {taskinstance.py:1377} INFO - Starting attempt 1 of 3
[2022-09-07, 23:21:56 UTC] {taskinstance.py:1378} INFO - 
--------------------------------------------------------------------------------
[2022-09-07, 23:21:56 UTC] {taskinstance.py:1397} INFO - Executing <Task(PythonOperator): load_image_data.create_loading_table> on 2011-01-02 00:00:00+00:00
[2022-09-07, 23:21:56 UTC] {standard_task_runner.py:52} INFO - Started process 2946 to run task
[2022-09-07, 23:21:56 UTC] {standard_task_runner.py:79} INFO - Running: ['***', 'tasks', 'run', 'phylopic_workflow', 'load_image_data.create_loading_table', 'scheduled__2011-01-02T00:00:00+00:00', '--job-id', '5', '--raw', '--subdir', 'DAGS_FOLDER/providers/provider_workflow_dag_factory.py', '--cfg-path', '/tmp/tmpnkh29a9t', '--error-file', '/tmp/tmpfuoqmyw_']
[2022-09-07, 23:21:56 UTC] {standard_task_runner.py:80} INFO - Job 5: Subtask load_image_data.create_loading_table
[2022-09-07, 23:21:56 UTC] {task_command.py:371} INFO - Running <TaskInstance: phylopic_workflow.load_image_data.create_loading_table scheduled__2011-01-02T00:00:00+00:00 [running]> on host 3b0b8f9e8fe3
[2022-09-07, 23:21:56 UTC] {taskinstance.py:1589} INFO - Exporting the following env vars:
AIRFLOW_CTX_DAG_EMAIL=openverse@wordpress.org
AIRFLOW_CTX_DAG_OWNER=data-eng-admin
AIRFLOW_CTX_DAG_ID=phylopic_workflow
AIRFLOW_CTX_TASK_ID=load_image_data.create_loading_table
AIRFLOW_CTX_EXECUTION_DATE=2011-01-02T00:00:00+00:00
AIRFLOW_CTX_TRY_NUMBER=1
AIRFLOW_CTX_DAG_RUN_ID=scheduled__2011-01-02T00:00:00+00:00
[2022-09-07, 23:21:56 UTC] {crypto.py:82} WARNING - empty cryptography key - values will not be stored encrypted.
[2022-09-07, 23:21:56 UTC] {base.py:68} INFO - Using connection ID 'postgres_openledger_upstream' for task execution.
[2022-09-07, 23:21:56 UTC] {dbapi.py:231} INFO - Running statement: SET application_name = 'phylopic_workflow__load_image_data.create_loading_table__20110102';
, parameters: None
[2022-09-07, 23:21:56 UTC] {base.py:68} INFO - Using connection ID 'postgres_openledger_upstream' for task execution.
[2022-09-07, 23:21:56 UTC] {dbapi.py:231} INFO - Running statement: 
  select pg_sleep(120);
  CREATE TABLE public.provider_data_image_phylopic_20110102T000000(
  foreign_identifier character varying(3000),
foreign_landing_url character varying(1000),
url character varying(3000),
thumbnail character varying(3000),
filetype character varying(5),
filesize integer,
license character varying(50),
license_version character varying(25),
creator character varying(2000),
creator_url character varying(2000),
title character varying(5000),
meta_data jsonb,
tags jsonb,
category character varying(80),
watermarked boolean,
provider character varying(80),
source character varying(80),
ingestion_type character varying(80),
width integer,
height integer);
, parameters: None
[2022-09-07, 23:23:57 UTC] {base.py:68} INFO - Using connection ID 'postgres_openledger_upstream' for task execution.
[2022-09-07, 23:23:57 UTC] {dbapi.py:231} INFO - Running statement: ALTER TABLE public.provider_data_image_phylopic_20110102T000000 OWNER TO ***;, parameters: None
[2022-09-07, 23:23:57 UTC] {base.py:68} INFO - Using connection ID 'postgres_openledger_upstream' for task execution.
[2022-09-07, 23:23:57 UTC] {dbapi.py:231} INFO - Running statement: 
CREATE INDEX IF NOT EXISTS provider_data_image_phylopic_20110102T000000_provider_key
ON public.provider_data_image_phylopic_20110102T000000 USING btree (provider);
, parameters: None
[2022-09-07, 23:23:57 UTC] {base.py:68} INFO - Using connection ID 'postgres_openledger_upstream' for task execution.
[2022-09-07, 23:23:57 UTC] {dbapi.py:231} INFO - Running statement: 
CREATE INDEX IF NOT EXISTS provider_data_image_phylopic_20110102T000000_foreign_identifier_key
ON public.provider_data_image_phylopic_20110102T000000 USING btree (provider, md5((foreign_identifier)::text));
, parameters: None
[2022-09-07, 23:23:57 UTC] {base.py:68} INFO - Using connection ID 'postgres_openledger_upstream' for task execution.
[2022-09-07, 23:23:57 UTC] {dbapi.py:231} INFO - Running statement: 
CREATE INDEX IF NOT EXISTS provider_data_image_phylopic_20110102T000000_url_key
ON public.provider_data_image_phylopic_20110102T000000 USING btree (provider, md5((url)::text));
, parameters: None
[2022-09-07, 23:23:57 UTC] {python.py:173} INFO - Done. Returned value was: None
[2022-09-07, 23:23:57 UTC] {taskinstance.py:1415} INFO - Marking task as SUCCESS. dag_id=phylopic_workflow, task_id=load_image_data.create_loading_table, execution_date=20110102T000000, start_date=20220907T232156, end_date=20220907T232357
[2022-09-07, 23:23:57 UTC] {local_task_job.py:156} INFO - Task exited with return code 0
[2022-09-07, 23:23:57 UTC] {local_task_job.py:273} INFO - 0 downstream tasks scheduled from follow-on schedule check
```
</details>


<details>
<summary>DB jobs query</summary>

```
SELECT pid, age(clock_timestamp(), query_start), usename, application_name, query 
 FROM pg_stat_activity 
 WHERE query != '<IDLE>' AND query NOT ILIKE '%pg_stat_activity%' 
 ORDER BY query_start desc;
+------+----------------+---------+------------------+---------------------------------------------------------------------+
| pid  | age            | usename | application_name | query                                                               |
|------+----------------+---------+------------------+---------------------------------------------------------------------|
| 31   | <null>         | <null>  |                  |                                                                     |
| 33   | <null>         | deploy  |                  |                                                                     |
| 29   | <null>         | <null>  |                  |                                                                     |
| 28   | <null>         | <null>  |                  |                                                                     |
| 30   | <null>         | <null>  |                  |                                                                     |
| 43   | 0:00:00.408960 | airflow |                  | COMMIT                                                              |
| 44   | 0:00:14.036838 | airflow |                  | COMMIT                                                              |
| 1363 | 0:00:52.850229 | airflow |                  | COMMIT                                                              |
| 49   | 0:00:52.851721 | airflow |                  | COMMIT                                                              |
| 1364 | 0:00:53.157167 | airflow |                  | COMMIT                                                              |
| 1365 | 0:00:53.159587 | airflow |                  | COMMIT                                                              |
| 1366 | 0:00:53.162553 | airflow |                  | COMMIT                                                              |
| 47   | 0:00:53.368939 | airflow |                  | COMMIT                                                              |
| 1766 | 0:01:01.635456 | deploy  |                  |                                                                     |
|      |                |         |                  |   select pg_sleep(120);                                             |
|      |                |         |                  |   CREATE TABLE public.provider_data_image_phylopic_20110109T000000( |
|      |                |         |                  |   foreign_identifier character varying(3000),                       |
|      |                |         |                  | foreign_landing_url character varying(1000),                        |
|      |                |         |                  | url character varying(3000),                                        |
|      |                |         |                  | thumbnail character varying(3000),                                  |
|      |                |         |                  | filetype character varying(5),                                      |
|      |                |         |                  | filesize integer,                                                   |
|      |                |         |                  | license character varying(50),                                      |
|      |                |         |                  | license_version character varying(25),                              |
|      |                |         |                  | creator character varying(2000),                                    |
|      |                |         |                  | creator_url character varying(2000),                                |
|      |                |         |                  | title character varying(5000),                                      |
|      |                |         |                  | meta_data jsonb,                                                    |
|      |                |         |                  | tags jsonb,                                                         |
|      |                |         |                  | category character varying(80),                                     |
|      |                |         |                  | watermarked boolean,                                                |
|      |                |         |                  | provider character varying(80),                                     |
|      |                |         |                  | source character varying(80),                                       |
|      |                |         |                  | ingestion_type character varying(80),                               |
|      |                |         |                  | width integer,                                                      |
|      |                |         |                  | height integer);                                                    |
| 1574 | 0:01:53.200383 | deploy  |                  |                                                                     |
|      |                |         |                  |   select pg_sleep(120);                                             |
|      |                |         |                  |   CREATE TABLE public.provider_data_image__7550167585681209474(     |
|      |                |         |                  |   foreign_identifier character varying(3000),                       |
|      |                |         |                  | foreign_landing_url character varying(1000),                        |
|      |                |         |                  | url character varying(3000),                                        |
|      |                |         |                  | thumbnail character varying(3000),                                  |
|      |                |         |                  | filetype character varying(5),                                      |
|      |                |         |                  | filesize integer,                                                   |
|      |                |         |                  | license character varying(50),                                      |
|      |                |         |                  | license_version character varying(25),                              |
|      |                |         |                  | creator character varying(2000),                                    |
|      |                |         |                  | creator_url character varying(2000),                                |
|      |                |         |                  | title character varying(5000),                                      |
|      |                |         |                  | meta_data jsonb,                                                    |
|      |                |         |                  | tags jsonb,                                                         |
|      |                |         |                  | category character varying(80),                                     |
|      |                |         |                  | watermarked boolean,                                                |
|      |                |         |                  | provider character varying(80),                                     |
|      |                |         |                  | source character varying(80),                                       |
|      |                |         |                  | ingestion_type character varying(80),                               |
|      |                |         |                  | width integer,                                                      |
|      |                |         |                  | height integer);                                                    |
| 1367 | 0:03:07.974316 | airflow |                  | COMMIT                                                              |
| 50   | 0:03:08.020433 | airflow |                  | COMMIT                                                              |
| 52   | 0:03:15.860842 | airflow |                  | COMMIT                                                              |
| 1359 | 0:03:53.310257 | deploy  |                  | COMMIT                                                              |
| 1236 | 0:05:50.341626 | airflow |                  | COMMIT                                                              |
| 1185 | 0:05:52.497251 | airflow |                  | COMMIT                                                              |
| 42   | 0:16:23.451864 | airflow |                  | COMMIT                                                              |
+------+----------------+---------+------------------+---------------------------------------------------------------------+
SELECT 22
Time: 0.024s
```
</details>

stacimc:
> So, I think that we want to kill jobs associated with the failed task, but not other jobs, associated with other ongoing tasks. [application_name](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-APPLICATION-NAME) and [task_instance_key_str](https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html) together could help, and that's the design represented here.

Got it, thank you! And I think you're right, we'll want to make sure we're only killing jobs associated with the failed task.

I checked out the temp commit you pushed to test a long-running query on `create_loading_table` (thank you!). That's exactly how I was initially testing as well! When I run it locally, I see that the entire `pg_sleep` runs, even though it exceeds the `execution_timeout` on the task 🤔 Said another way: the task ignores its `execution_timeout` and just hangs until the long-running query completes, so the new `on_failure_callback` is never hit.

I did a little bit of digging and got as far as finding a way of getting the timeouts to work in _PostgresOperators_, by passing in a `statement_timeout`. Here's an example task I made and added to the provider DAGs for testing:

```
sql_task = PostgresOperator(
    task_id="sql_task",
    postgres_conn_id=DB_CONN_ID,
    sql=(
        "select pg_sleep(120)"
    ),
    execution_timeout=timedelta(seconds=5) # This timeout is totally ignored
    runtime_parameters={'statement_timeout': '10s'}, # This timeout is respected
)

...

# How I added it to the DAG
sql_task >> [create_loading_table, copy_to_s3] >> load_from_s3
```

When I run that, the `execution_timeout` is still totally ignored, but the `statement_timeout` is respected and in fact the query is cancelled for me. In the logs I see:

```
psycopg2.errors.QueryCanceled: canceling statement due to statement timeout
```

I don't know how helpful this is, but maybe something to look into?

rwidom:
Thanks so much @stacimc ! 

I have code in a temporary test dag that I just committed that works for both setting the database `statement_timeout` to have postgres proactively kill the job, and `application_name` to enable airflow to go back and kill the job after the fact (e.g. if it fails for reasons other than timing out?).

So to the extent that my goal was to learn more about the broad array of possible ways to get this done, I feel pretty accomplished. 🤣 But I still am not 100% sure what would be the best thing to do for openverse. Here are some notes based on the different ways that the catalog is running postgres jobs, and what I'm learning about these different kill-the-job design options.

PG connection tool | Catalog Code | Notes
-- | -- | --
PostgresOperator | [inaturalist.py](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py#L141) | Postgres Operator won't parse airflow variables in the SQL statement, parameters, or runtime parameters, maybe as a protection against SQL injection. So, they have to be hard coded with the task and can't be differentiated across runs of the same task created by dag factories. However, since inaturalist is only a monthly run, and the postgres operator is used for a limited set of pre- and post- ingestion tasks, Airflow templating is not crucial here.
PostgresResultOperator | [class definition](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/operators/postgres_result.py) | We could modify this class to set the `application_name` and `statement_timeout` by default, but we still wouldn't be able to pass in the regular Airflow jinja templates.
PostgresResultOperator | [data refresh dag factory](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/data_refresh/dag_factory.py) | This is the only place that currently uses the result operator, and most of the underlying work for the refresh is done using a hook in popularity/sql.py.
PostgresHook | [common/loader/sql.py](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/loader/sql.py) | This gets called from a million different provider DAGs, which may have very different data volumes, timeouts, run frequencies, etc. So it seemed important to have a way to pass those details in to the database. It gets called only from common / loader / loader.py though, so those parameters will need to get passed through a bunch of layers.
PostgresHook | [common/popularity/sql.py](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/popularity/sql.py) | Seems like this is just run as part of the data refresh, but there are multiple calls to the PostgresHook for different steps along the way.
PostgresHook | [inaturalist.py](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py) | Main processing for this provider happens in SQL. Static preingestion steps use PostgresOperator, but the dynamic calls to get batches of data use the hook. Probably would rather not adjust database settings on each and every call, but probably could set things up to run once at initialization.
psycopg2 | [merge common crawl tags](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/commoncrawl/commoncrawl_scripts/scripts/merge_cc_tags.py) | Is this something that gets run regularly or was it a onetime thing as part of restructuring things?

So, yeah, I think that my next steps / open questions are:
- figuring out if/how we can pass the task-specific `execution_timeout` down to SQL loader, or finding out if there is a number that we feel comfortable using across all provider scripts, or maybe exploring numbers we could add to provider_ingestion_workflows as a parameter? How would that work for the data refresh?
- figuring out how we feel about risks around using `task_instance_key_str` and `ts_nodash` in database settings, how important killing jobs for non-timeout reasons is and whether pursuing the `application_name` approach even is, whether focusing on the `statement_timeout` for now would be better.
- figuring out how broadly we want this to apply: all of the examples in the table above? some of them? which? etc.

Thanks again!

stacimc:
This is such an interesting PR, @rwidom! I am learning a lot about how this works 😄 I _love_ that you've added the temporary test DAG, that makes it so easy to play around with. I have a bunch of thoughts, apologies for not organizing them super well. Let me know if this makes sense!

### Commoncrawl

For psycopg2, it looks like that's only used in the crawler as you've pointed out. The crawler hasn't been touched in a long time and is not running, and to my knowledge it's not been prioritized in any near term. I think we can avoid spending too much time on this part, but it's awesome that you've gotten it working!

### Custom operators

As I understand what you've said:  with `PostgresOperators`, the only problem is that to use the current approach, we still need to pass in an `application_name` and this is difficult because we can't use Jinja templating here
* Could we make a new operator, extending `PostgresOperator`? It could take care of adding the `application_name` and `statement_timeout`. I haven't fully fleshed this out, but I do know that from within the `execute` method, `PostgresOperator` has access to `self.dag_id` and `self.task_id`. There may be a way to access the `run_id` directly, and worst case you can probably query the dagruns to get it? Worth exploring a little deeper, maybe!

Just for the sake of discussion 😄, have you thought about making a custom operator for `PythonOperators` using `PostgresHook` to encapsulate this logic there as well? Curious what you think, but if we go down that road we should definitely separate the work out into multiple PRs/tickets to make it easier to manage!

### Approach/scope

> if there is a number that we feel comfortable using across all provider scripts, or maybe exploring numbers we could add to provider_ingestion_workflows as a parameter?

I think the way we've been writing these, it's safe to assume the `execution_timeout` for the task should be the `statement_timeout`. If down the line we found a case where those numbers should be separate, maybe the workflow configuration classes would be a cool place to store it :)

> whether pursuing the application_name approach even is, whether focusing on the statement_timeout for now would be better

Good question! Your idea that we'll need to be careful to kill only the query associated with the timed out task makes intuitive sense to me, but I'm really not very familiar with the database internals. Pinging @AetherUnbound for your thoughts on this especially 🧠 I do like that this approach lets you query for the specific query to kill using nothing but the task context. If we end up making our own custom operator, maybe there's something we can hook into to do clean up on timeout? I only looked briefly at the source, maybe [on_kill](https://github.com/apache/airflow/blob/main/airflow/models/baseoperator.py#L1138)?

> figuring out how broadly we want this to apply: all of the examples in the table above? some of them? which? etc.

Another good question 😄 I think this would be useful just about everywhere, certainly in the loader steps of the provider DAGs. It's totally reasonable for this to be split across multiple PRs, however! And as I said, the commoncrawl steps can be deprioritized since that's not running and we're not able to test it.

rwidom:
Ooh! I didn't know about `self.dag_id` or `self.task_id`! That makes so much sense. Does that mean that there is also a `self.execution_timeout` and a `self.ts_nodash`? I have to look at that, because if so, I do think that a custom hook and custom operator (maybe adding on to our existing custom pg operator?) could be a really nice and general solution (maybe in 2 separate PRs). And maybe I totally misunderstood the reason why the operator wasn't parsing variables the way I expected. Thanks @stacimc ! And thanks in advance for your thoughts @AetherUnbound !

AetherUnbound:
Sorry it took me so long to get to this! I _love_ the deep dive you've done Rebecca and I've learned a bunch 🚀 The `statement_timeout` piece is one I feel like I definitely should have been aware of before 😅 But nonetheless I'm so glad you've found it, it seems perfect!

One thing I'm getting caught up on as you're describing this:

> Postgres Operator won't parse airflow variables in the SQL statement, parameters, or runtime parameters, maybe as a protection against SQL injection.

This doesn't quite match my understanding of what Airflow is capable of, do you mind elaborating on this a bit? At least with the `PostgresOperator`, the `sql` parameter is a templated field so we should be able to do any manner of Airflow templating we'd like there:

https://github.com/apache/airflow/blob/84d915c249617aedaf0c451e9f9d19a4e7ff767b/airflow/providers/postgres/operators/postgres.py#L47

If we wanted other attributes (e.g. `runtime_parameters`) to be templated, we could subclass the `PostgresOperator` (which is an idea already floating around in this PR!) and add it to the list of `template_fields`.

Personally, I think that the best way to go _would_ be `runtime_parameters` exclusively. Don't get me wrong, the kill-based-on-application_name-and-task-context code you've drawn up is SO COOL. I'm way impressed that's possible at all and the code you've set up to accomplish that is really slick. But I do think that managing timeouts around the `runtime_parameters`'s `statement_timeout` would be the most ideal route. It feels like what we'll want down the line is to have that field automatically be populated for `PostgresOperator` steps based on the `execution_timeout`. 

Heck, we may even want a `PostgresHook` wrapper which brings all the logic in `PostgresOperator::execute` into `PostgresHook::run` so we can enforce timeouts there 😮 Even if we're running a `PythonOperator` that uses a `PostgresHook`, we'd still get into the scenario where a PG query either hangs or doesn't get terminated on the backend even though the python thread does. Jeez, now as I'm saying that I feel like those cases could make use of the kill-query-on-failure logic 😅 Perhaps we could at least start with utilizing `statement_timeout`, and see if we continue to encounter issues where we'd want the kill-query-on-failure logic then implement that down the line as needed?

rwidom:
Hey, @AetherUnbound, I was just thinking about returning to this. Sorry for the delay! Is it still high priority? Does https://github.com/WordPress/openverse-catalog/pull/939 make it irrelevant now? I am more familiar with xcoms and parameters now, so I might be better equipped to finish this up, but just trying to figure out if that makes sense right now.

AetherUnbound:
I don't _think_ this is made irrelevant by #939, but the diff might look a bit different because of the dependency changes there. I think terminating the query is still important and so could still be high priority, but I could also see it being moved down to medium since we haven't encountered this particular issue recently 🤔 What do you think @WordPress/openverse-catalog?

rwidom:
OK, I don't know if it's just that I haven't had enough real focus time on this, or that I'm really missing the boat on some basic airflow concepts or what, but here's where I think things are... 

- I tried (see the dag in `test_timeout_pg.py`), and I don't think I can recreate the original issue.
- However, that might be about failures in using `task_instance` attributes to identify processes using postgres `application_name`. In an earlier draft, I had planned to use `task_instance_key_str` (see https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html) but I was already concerned about tasks that run more than once a day, and now it definitely will not allow us to differentiate across mapped processes within a single task instance, which would really be a bummer if we start running mapped tasks in parallel like we've discussed for inaturalist. And/or, ideally we could pass the `task` info to set the database `statement_timeout` to match the `airflow_timeout`.
- I'm not 100% clear from the original issue (#690) on whether idle _state_ processes are the main concern or active running processes that have been abandoned. But for idle processes there might be a more general setting we could use, see https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/_modules/airflow/providers/postgres/hooks/postgres.html. And if there are really just some very long-running jobs that inspired this, maybe we should be defining separate connections for them with appropriate settings to automatically kill off idle processes?

So, I've been doing a lot of thinking in circles, and not making a ton of real progress, but now you have the latest @AetherUnbound  and @stacimc . Help?

AetherUnbound:
@rwidom and I had a quick synchronous discussion where we talked about the various possibilities for this endeavor and what might make the most sense moving forward. She prepared [this excellent doc](https://docs.google.com/document/d/121DwkExnRHhqf_0beXBjywbJ9Y8QpqxitXnQZGxBQYI/edit) describing what possibilities we have. Here are the outcomes of that discussion:

- Airflow can't be trusted to manage the timeout of the Postgres steps because it will just hang until the Postgres query is done executing. This makes trying to employ anything via the `on_failure_callback` useless since the callback won't run until the query is complete anyway. With this in mind we are **abandoning the approach of using `application_name` & other identifiers to find and stop the query from Airflow**.
- We want to use the [`statement_timeout` directive](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-STATEMENT-TIMEOUT) on the Postgres end for setting statement timeouts. We will create a _new_ `PostgresHook` (which will be a subclass of the `airflow.providers.postgres.hooks.postgres.PostgresHook`) which will allow both a hook-level default statement timeout AND a `statement_timeout` parameter to the `run` function on the hook.
- We will also add a new Operator (likely to be named `PgExecuteQueryOperator`) which will automatically hand the `execution_timeout` of the task down into the hook when it is instantiated. This will ensure that cases where we execute a query using this operator have the task's execution timeout applied as the statement timeout.
- We'll go in and add any clear/obvious statement timeouts to existing `PostgresHook::run` calls, but we won't spend too much time on that as part of this PR since we can always add those timeouts down the line.

I also shared this pseudo-code with Rebecca for what the hook might look like based off of [her initial code](https://github.com/WordPress/openverse-catalog/blob/37000dbbf61842d76e5e7b487d2d1eb170eb4a97/openverse_catalog/dags/common/sql_helpers.py#L13-L27):

```python
class TimedPostgresHook(PostgresHook):
    """
    PostgresHook that sets the database timeout on any query to match the airflow task
    execution timeout.
    """

    def __init__(self, *args, default_statement_timeout: str = None, **kwargs) -> None:
        self.default_statement_timeout = default_statement_timeout

    def get_pg_timeout_sql(self, statement_timeout: str) -> str:
        return f"SET statement_timeout TO '{statement_timeout}s';"

    def run(self, sql, statement_timeout: str = None, autocommit=False, parameters=None, handler=None):
        statement_timeout = statement_timeout or self.default_statement_timeout

        if statement_timeout:
            sql = [self.get_pg_timeout_sql(statement_timeout) + sql]
        super().run(sql, autocommit, parameters, handler)
```

I _think_ that covers everything we discussed, but Rebecca please chime in if I missed anything 😄

rwidom:
Thanks so much @AetherUnbound ! That's perfect!

AetherUnbound:
> One more thought on naming: I have these things in a file called sql_helpers, but I think sql_timeout would probably be more clear. Thoughts?

In line with some of our other packages naming conventions, maybe `sql_utils` would be best?

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
```diff
@@ -201,7 +215,7 @@ def compare_update_dates(
         raise AirflowSkipException("Nothing new to ingest")
 
     @staticmethod
-    def load_catalog_of_life_names(remove_api_files: bool):
+    def load_catalog_of_life_names(task: PythonOperator, remove_api_files: bool):
```
AetherUnbound, line 218: I believe the typing on this would be:
```suggestion
    def load_catalog_of_life_names(task: TaskInstance, remove_api_files: bool):
```
Of `airflow.models.TaskInstance`

openverse_catalog/dags/common/sql_helpers.py
```diff
@@ -0,0 +1,136 @@
...
+        **kwargs,
+    ):
+        statement_timeout = statement_timeout or self.default_statement_timeout
+
+        if statement_timeout:
+            sql = f"{self.get_pg_timeout_sql(statement_timeout)} {sql}"
+        return super().run(sql, autocommit, parameters, handler)
+
+    @staticmethod
+    def get_execution_timeout(task) -> float:
```
AetherUnbound, line 79: Same note here regarding typing, this could be `airflow.models.TaskInstance` I believe.

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
```diff
@@ -237,13 +251,18 @@ def load_catalog_of_life_names(remove_api_files: bool):
...
+        )
         COPY_SQL = (
             "COPY inaturalist.{} FROM STDIN "
             "DELIMITER E'\t' CSV HEADER QUOTE E'\b' NULL AS ''"
         )
         COUNT_SQL = "SELECT count(*) FROM inaturalist.{};"
         # upload vernacular names file to postgres
+        pg.run(
+            ""
+        )  # just to set the timeout, because it's not automatic for copy_expert
```
AetherUnbound, line 265: Maybe instead, we could have a separate function for cases like this that more explicitly runs the statement timeout? Like `pg.run_statement_timeout()`?

tests/dags/common/loader/test_sql.py
```diff
@@ -99,6 +103,16 @@ def create_query_values(
...
 
+@pytest.fixture
+def mock_dag() -> DAG:
+    return create_pg_timeout_tester_dag()
+
+
+@pytest.fixture
+def mock_task(mock_dag) -> BaseOperator:
+    return mock_dag.get_task("mock_task")
+
```
AetherUnbound, line 114: Since these fixtures are used in multiple places, you can actually set up a `conftest.py` file at a higher directory (whichever level in the hierarchy you feel most appropriate for these) which includes the fixtures. Pytest will then manage making them available to all tests within that directory tree! https://docs.pytest.org/en/7.1.x/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files

tests/dags/common/popularity/test_sql.py
```diff
@@ -7,14 +7,29 @@
...
 
 
+@pytest.fixture
+def mock_dag() -> DAG:
+    return create_pg_timeout_tester_dag()
+
+
+@pytest.fixture
+def mock_task(mock_dag) -> BaseOperator:
+    return mock_dag.get_task("mock_task")
```
AetherUnbound, line 30: Same note here RE sharing fixtures!

tests/dags/common/test_resources/dags/test_timeout_pg.py
```diff
@@ -0,0 +1,81 @@
...
+    task_id = "pg_hook_happy"
+    dag_run = HAPPY_DAG.create_dagrun(
+        run_id=f"test_{task_id}_{exec_dt.strftime('%y%m%d%H%M%S')}",
+        state="queued",
+        execution_date=exec_dt,
+        data_interval=(exec_dt - (2 * interval), exec_dt - interval),
+    )
+    ti = dag_run.get_task_instance(task_id)
+    execution_timeout = dag_run.get_dag().get_task(task_id).execution_timeout
+    print(execution_timeout.total_seconds())
```
AetherUnbound, line 81: I think this could be its own test too! Just as a possible smoketest for making sure it all runs fine 🙂

tests/dags/common/test_sql_helpers.py
```diff
@@ -0,0 +1,128 @@
...
+default_timeout_msg = (
+    f"DAG_DEFAULT_ARGS sets the default task execution timeout to "
+    f"{DAG_DEFAULT_ARGS['execution_timeout']}. Testing for {DEFAULT_TIMEOUT} seconds."
+)
+
+
+# PostgresHook works with or without explicit connection id, and with override conn id
+# Testing mainly for documentation purposes
+# Not using pg.run("select 1;") to test the actual connection, because we don't spin up
+# the databases for testing.
```
AetherUnbound, line 33: We do have a database for testing though, and this could be called! We do other operations with the database as part of our tests.

tests/dags/common/test_sql_helpers.py
```diff
@@ -0,0 +1,128 @@
...
+        pytest.param("xyzqwerty", 60, ("xyzqwerty", 60), id="override_both"),
+    ],
+)
+def test_PostgresHook_init_connection(conn_id, timeout, expected_results):
+    (expected_conn_id, expected_timeout) = expected_results
+    pg = PostgresHook(conn_id, timeout)
+    actual_conn_id = pg.postgres_conn_id
+    actual_timeout = pg.default_statement_timeout
+    assert actual_conn_id == expected_conn_id
+    assert actual_timeout == expected_timeout
```
AetherUnbound, line 63: This is definitely a stylistic nit, but this could be:
```suggestion
    assert pg.postgres_conn_id == expected_conn_id
    assert pg.default_statement_timeout == expected_timeout
```

tests/dags/common/test_sql_helpers.py
```diff
@@ -0,0 +1,128 @@
...
+from common.sql_helpers import PGExecuteQueryOperator, PostgresHook
+from psycopg2.errors import QueryCanceled
+
+from tests.dags.common.test_resources.dags.test_timeout_pg import (
+    TEST_SQL,
+    create_pg_timeout_tester_dag,
+)
+
+
+logger = logging.getLogger(__name__)
```
AetherUnbound, line 15: These tests rock!!

tests/dags/common/test_sql_helpers.py
```diff
@@ -0,0 +1,128 @@
...
+        sql=TEST_SQL,
+        execution_timeout=execution_timeout,
+    )
+    actual_result = operator.get_db_hook().default_statement_timeout
+    assert actual_result == expected_result
+
+
+# Happy path DAG works without error (Each task is a test case)
+# This generates a warning about running a dag without an explicit data interval being
+# deprecated, but I haven't found a way to get it through with this function.
```
AetherUnbound, line 126: If it becomes too much of a nuisance, you can use pytest to silence warnings within a test: https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html

tests/dags/providers/provider_api_scripts/test_inaturalist.py
```diff
@@ -167,6 +167,8 @@ def test_consolidate_load_statistics(all_results, expected):
     ],
 )
 def test_get_batches(batch_length, max_id, expected):
-    with mock.patch.object(PostgresHook, "get_records", return_value=max_id) as pg_mock:
-        actual = INAT.get_batches(batch_length, pg_mock)
-        assert actual == expected
+    task = "doesn't matter, but we have to pass something"
```
AetherUnbound, line 170: In this case, you can just use the following, as it gets the point across 🙂 
```suggestion
    task = mock.Mock()
```

openverse_catalog/dags/common/loader/sql.py
```diff
@@ -176,6 +191,7 @@ def load_s3_data_to_intermediate_table(
 
 def clean_intermediate_table_data(
     postgres_conn_id: str,
+    task: BaseOperator,
```
AetherUnbound, line 194:
```suggestion
    task: TaskInstance,
```

openverse_catalog/dags/common/popularity/sql.py
```diff
@@ -221,13 +235,21 @@ def create_media_popularity_constants_view(
...
+    )
     if media_type == AUDIO:
         popularity_constants = AUDIO_POPULARITY_CONSTANTS_VIEW
         popularity_constants_idx = AUDIO_POP_CONSTANTS_IDX
         popularity_metrics = AUDIO_POPULARITY_METRICS_TABLE_NAME
         popularity_percentile = AUDIO_POPULARITY_PERCENTILE_FUNCTION
+    # TO DO: How long does this usually take and how often is it usually run? Is the
+    # intention here to create the materialized view syntax without populating it? If
+    # so, maybe a much shorter timeout would be appropriate, but then I would expect to
+    # see the phrase "WITH NO DATA" here.
```
AetherUnbound, line 251: It looks like we renamed the tasks, so we don't get the durations in the grid view for this dag. I'm not sure if you have access to this Rebecca, but here's the task list for these tasks: https://airflow.openverse.engineering/taskinstance/list/?_flt_3_dag_id=recreate_image_popularity_calculation&_flt_3_task_id=create_image_popularity_constants_view

image

Durations ranged from 4-6 hours, but the last time this was run was almost a year ago so I'd expect it to be even higher. Perhaps 12 or 24 hours would make sense?

openverse_catalog/dags/common/popularity/sql.py
```diff
@@ -267,12 +289,19 @@ def create_media_popularity_constants_view(
...
+    task,
     media_type=IMAGE,
     popularity_constants_view=IMAGE_POPULARITY_CONSTANTS_VIEW,
 ):
     if media_type == AUDIO:
         popularity_constants_view = AUDIO_POPULARITY_CONSTANTS_VIEW
-    postgres = PostgresHook(postgres_conn_id=postgres_conn_id)
+    # TO DO see note above about when the materialized view gets populated. Not sure how
+    # often it is refreshed verses instantiated from scratch. And whether / how we might
+    # improve performance by integrating index updates/builds with that schedule.
```
AetherUnbound, line 300: It's only re-instantiated when we make a change to the schema that would require recreating it I believe. The refresh is run _much_ more frequently, I believe monthly as part of the data refresh now!

AetherUnbound (commented):
This is fantastic, thanks so much for all your efforts on this Rebecca! I have a lot of thoughts and comments, please let me know if any of them are unclear and I can expound on them more 😄

openverse_catalog/dags/common/sql_helpers.py
```diff
@@ -0,0 +1,136 @@
...
+        *args,
+        **kwargs,
+    ) -> None:
+        self.default_statement_timeout = default_statement_timeout
+        # TO DO: Is this really what we want? I think so, either explicitly override it,
+        # or you'll get the default task execution timeout, which is an hour currently.
+        if default_statement_timeout is None:
+            self.default_statement_timeout = DAG_DEFAULT_ARGS[
+                "execution_timeout"
+            ].total_seconds()
```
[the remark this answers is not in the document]
rwidom: Awesome! Yeah, I was thinking if we don't want that default to really be the default, we should change it. Right? Like if we find there are a ton of things that take an hour and fifteen minutes and it's a pain, maybe we just switch the default to 2 hours.

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
rwidom: This I'm pretty confident about. A task _instance_, does not necessary have the `execution_timeout`. That piece of information lives in the task (type = operator), or in the dag (type DAG, where it's the default for tasks within that dag), but not every task_instance object (at least when I was testing things) has all of the task info associated with it and not ever task has all of the dag info associated with it. I think that some of the challenge is that these are ultimately in different tables in a relational database and they don't always come together as a package when they're in python.  That's why I have `task` in the parameters rather than `ti`, and in some cases, both `task` and `ti`. :/

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
rwidom: That's better. I was considering when and whether I should get into the guts of more of the hook functions (like running a list of SQLs, which I didn't see in our code base or running copy_expert), but that's a nice solution to avoid the issue with copy expert at least.

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
```diff
@@ -110,7 +113,16 @@ def load_transformed_data(
...
+        # Considered using PostgresHook.get_execution_timeout(task), but not sure how
+        # that will work for mapped tasks. Is it total across all mappings or for each?
+        # This we know is for each iteration.
+        # In testing this locally, the longest full iteration took 39 minutes,
+        # median was 18 minutes. We should probably adjust the timeouts with
+        # more info from production runs.
+        pg = PostgresHook(
+            postgres_conn_id=POSTGRES_CONN_ID,
+            default_statement_timeout=timedelta(hours=1).total_seconds(),
+        )
```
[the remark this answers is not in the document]
rwidom: Yeah, it's just that the operator kwargs are stored a little differently in the [MappedOperator class](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/mappedoperator/index.html). But I added a branch to handle them separately.

[This](https://github.com/apache/airflow/blob/eb46eeb33d58436aa5860f2f0031fad3dea3ce3b/airflow/models/taskinstance.py#L463) is what makes me think that tasks are safer than task_instances for timeout purposes.

openverse_catalog/dags/common/sql_helpers.py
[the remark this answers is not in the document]
AetherUnbound: Definitely!

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
AetherUnbound: AAHHHHH, okay my mind spaced and I thought this was `task_instance` - didn't realize you could request `task` from the airflow context too!

AetherUnbound:
Shoot I forgot to mention this in my initial review, but due to the `task`'s proximity to `**context`/`kwargs` in general, I think it makes the most sense to add onto the _end_ of the parameter list in `loader.sql` and elsewhere in the functions it's required in, if that makes sense. I realize this is a preference piece but one way to think about it is `def func(<arguments the function needs and uses directly>, <other pieces it might need to divine other information>)`. A good example of this is something like slack's [send message function](https://github.com/WordPress/openverse-catalog/blob/3a963fce26d8dc081c6f64da656d8de76f3691a7/openverse_catalog/dags/common/slack.py#L310-L318).

rwidom:
Thank you so much for the awesome feedback, @AetherUnbound !

Here's where I think things are now:
- I renamed `common.sql_helpers` to just `common.sql`. 
- Sorry that makes everything outdated, but I put thumbs up on the comments where I made a change.
- TIL `conftest`!  So much better! Yay! (And I hope that the placement is ok.)
- I set 45 minutes for the inaturalist `load_transformed_data` tasks, but I think the default is probably better, so I'll do that.
- I haven't changed `task: BaseOperator` to `ti: TaskInstance`, because of [this](https://github.com/apache/airflow/blob/eb46eeb33d58436aa5860f2f0031fad3dea3ce3b/airflow/models/taskinstance.py#L463). Is that ok?
- On the order of args, putting `task` last was my initial thought, but then I got into a bunch of errors, because it came after optional arguments. And putting it as the last non-optional arg still means going back and double checking the order of everything. I could change it to `task: BaseOperator = None` to really go last, but if people want to override the default timeout I want that to work reliably, and I worry about where/when/how we would catch the error. I could do that plus `if task is None: raise TypeError` up front. I could even make `check_task_type` or `error_if_none` a static method in the Hook. Would that be the best way to go?
- Anything else?

I have learned so so much working on this, and I'm really so grateful for the opportunity and all of the guidance and help along the way!  I still feel nervous about merging, and like setting up some short, medium, and long-running connection IDs and putting them in the code manually could be safer in terms of a staged rollout and easier to maintain in the long run. Though I guess it would be [overloading the field](https://robertlambert.net/2014/09/a-field-guide-to-overloaded-data/), so to speak, and it would force a choice from a much shorter menu every time rather than using defaults and lots of flexibility to override them. There wouldn't be the override-within-a-single-hook-instance option that we currently have with `run(statement_timeout=...)`...

OK, so I think that's the latest. Feedback really really welcome!

AetherUnbound:
> * I haven't changed `task: BaseOperator` to `ti: TaskInstance`, because of [this](https://github.com/apache/airflow/blob/eb46eeb33d58436aa5860f2f0031fad3dea3ce3b/airflow/models/taskinstance.py#L463). Is that ok?

Yes, that's fine! Fault is mine for not realizing you could pull `task` from the [Airflow templates](https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html) (I think it might not have been there in previous versions 😅)

> * On the order of args, putting `task` last was my initial thought, but then I got into a bunch of errors, because it came after optional arguments. And putting it as the last non-optional arg still means going back and double checking the order of everything. I could change it to `task: BaseOperator = None` to really go last, but if people want to override the default timeout I want that to work reliably, and I worry about where/when/how we would catch the error. I could do that plus `if task is None: raise TypeError` up front. I could even make `check_task_type` or `error_if_none` a static method in the Hook. Would that be the best way to go?

I am fully in support of the `task: BaseOperator = None` + static method to check on the base class approach! I think that's perfect.

I'm so glad, this has been a great learning opportunity for me too!! Honestly, I feel comfortable moving forward with it as-is 😁 if we have to do some fast-follow PRs to adjust timeouts (or even revert temporarily), it wouldn't be too much extra effort!

openverse_catalog/dags/data_refresh/data_refresh_types.py
```diff
@@ -28,7 +28,9 @@ class DataRefresh:
                                        should be run.  Passed to the
                                        airflow.dag.DAG __init__ method.
     data_refresh_timeout:              int giving the amount of time in seconds a
-                                       given data pull may take.
+                                       given data pull may take. int rather than
+                                       than timedelta because it is used by HttpSensor
+                                       for scheduling, rather than PG for running tasks.
```
stacimc, line 33: Nice adding this clarification 👍

openverse_catalog/dags/database/image_expiration_workflow.py
```diff
@@ -21,7 +21,7 @@
     default_args={
         **DAG_DEFAULT_ARGS,
         "retry_delay": timedelta(seconds=15),
-        "execution_timeout": None,
+        "execution_timeout": timedelta(seconds=0),  # Just let these tasks run
```
stacimc, line 24: Possible clarification:
```suggestion
        "execution_timeout": timedelta(seconds=0),  # Let these tasks run with no timeout
```

openverse_catalog/dags/database/report_pending_reported_media.py
```diff
@@ -152,6 +155,10 @@ def create_dag():
                     "db_conn_id": OPENLEDGER_API_CONN_ID,
                     "media_type": media_type,
                 },
+                # TO DO: add a more specific timeout here?
+                # Seems like images and audio could have very different timeouts just
+                # based on data volumes. I was going to guess 2 hours, but the dag uses
+                # the default timeout of 1 hour, so I'm going to leave it at that.
```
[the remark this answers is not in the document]
stacimc: +1

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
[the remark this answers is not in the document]
stacimc: `sql.clean_intermediate_table_data` and `sql.upsert_records_to_db_table` are already using the timeout from the task, right? Testing locally, it looks like it is working in those cases.

But doesn't this mean that each of the individual SQL commands that are part of this task all have a separate 1-hr timeout, that could theoretically end up getting hit sequentially? Is that intended/something we have to live with?

openverse_catalog/dags/common/popularity/sql.py
```diff
@@ -129,6 +137,7 @@ def create_media_popularity_metrics(
 
 def update_media_popularity_metrics(
     postgres_conn_id,
+    task,
```
stacimc, line 140: This will cause an error in the `update_media_popularity_metrics_table` step of the data refresh, because it's passing `postgres_conn_id` and `media_type` in as _args_ rather than kwargs ([here](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/data_refresh/refresh_popularity_metrics_task_factory.py#L44)). So you'll get this error when trying to unpack:

`ValueError: The key 'task' in args is a part of kwargs and therefore reserved.`

One way to fix this is by updating the task to pass those in as `op_kwargs` instead, the way it's done in the provider dag factory.

openverse_catalog/dags/common/popularity/sql.py
```diff
@@ -267,12 +289,19 @@ def create_media_popularity_constants_view(
 
 def update_media_popularity_constants(
     postgres_conn_id,
+    task,
```
stacimc, line 287: Same comment as above, this will cause an error in the data refresh if we don't update the task to pass all of these in as kwargs.

openverse_catalog/dags/common/popularity/sql.py
```diff
@@ -383,10 +418,14 @@ def create_media_view(
 
 def update_db_view(
     postgres_conn_id,
+    task,
```
stacimc, line 413: The `refresh_matview` task also needs to be updated to prevent the `ValueError` here.

openverse_catalog/dags/common/loader/sql.py
```diff
@@ -345,8 +372,16 @@ def _delete_malformed_row_in_file(tsv_file_name, line_number):
                 write_obj.write(line)
 
 
-def expire_old_images(postgres_conn_id, provider, image_table=TABLE_NAMES[IMAGE]):
-    postgres = PostgresHook(postgres_conn_id=postgres_conn_id)
+def expire_old_images(
+    postgres_conn_id,
+    task,
```
stacimc, line 377: I left a comment below about needing to update tasks that use `op_args` rather than `op_kwargs`, or else the argument unpacking will break here. This one will also need to be updated.

stacimc (changes requested):
This is incredible work as usual :) I tested with most of our DAGs (set an INGESTION_LIMIT and turned everything on while I was reviewing code 😄), and it's looking really good!

Other than the comments that @AetherUnbound left, I left some thoughts but only have two real concerns:
* Some of the tasks in the data refreshes and image expiration flow will error right now due to the way the `task` is passed in, but it's an easy fix (I left comments on all the ones I found, but definitely double check me)
* I wanted your thoughts on the timeouts "stacking up" in tasks that have multiple SQL statements, each of which have the timeout applied (eg iNaturalist). Am I right that those timeouts would stack up? Is that a problem? If so, I don't think that has to block this PR, which brings *so* much value, but something to think about.

openverse_catalog/dags/providers/test_timeout_pg.py
```diff
@@ -0,0 +1,196 @@
...
+    ts_nodash: to append a timestamp to the above to identify an execution within a day.
+    execution_timeout:
+    """
+    # check that Airflow jinja is being parsed, raise an error if not.
+    if "ts_nodash" in ts_nodash or "task_instance_key_str" in task_instance_key_str:
+        raise ValueError(f"No jinja parsing: {ts_nodash=}, {task_instance_key_str=}")
+    db_settings = dict()
+    # application_name to identify job if we need to terminate it, particularly for
+    # failures that are not from timing out. the unique identifier from airflow only
+    # counts daily runs so adding timestamp.
```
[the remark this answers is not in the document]
rwidom: On the first question, I don't know. It kind of depends on how much is ever running at the same time in the catalog db, and how you all are administering it. I'm not sure if it's just my impulse to have all the data all the time, or if there are enough troubling weird straggler things going on that some labeling would help sort things out.

On the second, yes, exactly.

openverse_catalog/dags/providers/test_timeout_pg.py
```diff
@@ -0,0 +1,196 @@
...
+    ts_time_only = ts_nodash[8:]
```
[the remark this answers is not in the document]
rwidom: That makes total sense, yes.

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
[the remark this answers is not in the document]
AetherUnbound: Hmm, that's a good point. I think relying on the default task timeout here is OK, even if it means _cumulatively_ that the task may go over its official 1 hour limit. In the worst case, the task takes double the applied time out (2 hours) if we assume that a 1h timeout query is executed in the last minute of the task's total 1h timeout. Given that this is a significant improvement over no timeout at all and the query hanging indefinitely (or at least until it resolves), I'm fine leaving it as-is! Especially since it'd take some effort to parse out reasonable sub-timeouts for each command, which I'm not sure is worth it IMHO.

openverse_catalog/dags/common/popularity/sql.py
AetherUnbound: I think this may be addressed by moving `task` into the kwargs anyway, which is something we spoke about above! Great call out.

openverse_catalog/dags/common/sql.py
```diff
@@ -106,6 +113,9 @@ def get_execution_timeout(task) -> float:
     def get_pg_timeout_sql(statement_timeout: float) -> str:
         return f"SET statement_timeout TO '{statement_timeout}s';"
 
+    def run_statement_timeout(self):
```
AetherUnbound, line 116: Do you mind adding a docstring to this?

tests/dags/common/conftest.py
```diff
@@ -54,3 +60,77 @@ def identifier(request):
...
+
+
+@pytest.fixture(scope="session")
+def mock_dag():
+    return create_pg_timeout_tester_dag()
+
+
+@pytest.fixture(scope="session")
+def mock_task(mock_dag) -> BaseOperator:
+    return mock_dag.get_task("pg_hook_happy")
```
AetherUnbound, line 136: Since these fixtures are at almost the highest level `conftest.py` that means that they will be available to **all** tests in `tests/dags/common`. It might be prudent to have more specific names than `mock_dag` and `mock_task`, which speak to something much more generic than what's being described here and could be confusing for folks writing tests elsewhere. Maybe `mock_pg_timeout_dag` and `mock_pg_hook_task`?

AetherUnbound (commented):
I responded a bit to @stacimc's rolled-up-timeouts concern in a comment. TL;DR I think it's OK in the cases where we have it here, it will net us an improvement anyway, and at worst it means we'll be at `2 x timeout`.

openverse_catalog/dags/common/loader/sql.py
stacimc: Actually this can likely be ignored as this DAG is being retired!

stacimc:
> I responded a bit to @stacimc's rolled-up-timeouts concern in a comment. TL;DR I think it's OK in the cases where we have it here, it will net us an improvement anyway, and at worst it means we'll be at 2 x timeout.

@AetherUnbound I don't see this comment, did it get posted? Sorry if I'm looking right at it 😅 I thought the iNaturalist `load_transformed_data` tasks had more sequential statements with the timeout applied. I don't think that should be a big problem though (especially considering how smoothly iNaturalist is running now 😄), more something to keep in mind for future use cases.

But just to reiterate I couldn't agree more -- as this stands this is a **huge** improvement even in those very few cases, and _if_ we decide it's worth adding any further handling I think we'd definitely want to do that in a separate PR.

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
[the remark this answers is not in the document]
rwidom: Yeah, my assumption was that most of the time we could live with this for a first go. If there are places where we think a shorter timeout is in order, the task level timeout is just the default for the hook and we (but not me, honestly!) could always go back and use different `statement_timeout`s in the `run` commands within a given function.

rwidom:
Hello!

I think I responded above about the timeouts stacking, but yes, it's something to be aware of. It's really per SQL statement, not per task, and a task can have multiple SQL statements. But maybe future PR(s) can add `statement_timeout`s to the `run` statement where needed, if the intention is to divide timeouts differently among statements within a task?

Also, yes, thanks so much for catching those `op_args` vs `kw_args` issues!  I think I got them all, but we use the PostgresHook in _a lot_ of places in the catalog. I really appreciate the second and third set of eyes. Maybe future work could implement/standardize some of the related ideas we've talked about here?

- using the hook default postgres_conn_id, so that it doesn't need to be passed as a parameter so many times
- using `task: BaseOperator = None` and implementing type checking to handle times when the task doesn't get passed to the function that instantiates the Hook, so that `task` can always be the last arg
- other things?

For now, I did a lot of switching to kwargs in the function calls, mainly because it seemed like the path of least resistance and most clarity, but let me know if I should be doing something else instead.

AetherUnbound:
I think I'd still prefer the `task` attribute being moved down in the argument order as part of this PR, that way we aren't changing function parameter orders significantly twice for all of these functions. Is that something you're comfortable changing @rwidom?

rwidom:
> I think I'd still prefer the `task` attribute being moved down in the argument order as part of this PR, that way we aren't changing function parameter orders significantly twice for all of these functions. Is that something you're comfortable changing @rwidom?

My only concern is that I have already done a bunch (if not all, I think all) of those changes in the function calls and tests, so I think I would need to now go back through all of those changes and change them again. Also, I'm not entirely sure how we want the error handling to work if the task is None. Just set the default task timeout and act like everything is ok? Add some info to the log? Raise an error? Like, how important is it that the task timeouts we're setting actually get implemented (setting aside stacking concerns) at the level of detail we're setting?

AetherUnbound:
Per an offline discussion that we had, I took on the effort of moving the function parameters around so that the `task` param was at the end. @rwidom please check out that commit and let me know what you think 🙂 I also got a little lost in warning land and found some crucial mistakes I had made in #985 when removing the old warnings. 

I'm going to try to run a bunch of DAGs locally just to be sure, but I think this should be good to go now!

AetherUnbound (approved):
Thanks so much Rebecca for all your effort with this one! Can't wait to have it in 😄

stacimc (approved):
LGTM! Lowered the ingestion limit and ran just about all of our DAGs including the data refreshes. The changes to the parameters look great @AetherUnbound 💯 

Thanks so much again for your incredible work here @rwidom! I'm so excited to see this in 🎉

rwidom:
This is a really small nit, but I think that technically tasks can be either `BaseOperator` or `MappedOperator` and for reasons I don't understand, the latter is not a subclass of the former. Maybe we should change`BaseOperator` to `AbstractOperator`, per this https://airflow.apache.org/docs/apache-airflow/stable/_modules/airflow/models/mappedoperator.html ? It looks like that's the parent class of BaseOperator as well (https://airflow.apache.org/docs/apache-airflow/stable/_modules/airflow/models/baseoperator.html).

rwidom:
Ugh, I just started to make that change myself and got annoyed with myself about navigating the imports and wondering is `(BaseOperator | MappedOperator)` would be better. Maybe given that these things aren't really enforced we should just leave it as is and merge?

AetherUnbound (approved):
I think the `AbstractOperator` typing looks great! This is ready to me - [merge is "go" on your command](https://www.youtube.com/watch?v=l78dlQP8kmw&t=182s) 🚀😁

Issue #696, opened 2022-08-24
Add translations for top level taxonomies in inaturalist


rwidom:
## Problem
All of the descriptive text about photos from inaturalist is in Latin, which will make the search results less useful for users who are unfamiliar with the scientific nomenclature.

## Description
Taxonomies are trees though, so we could at least look up some higher levels terms (e.g. "birds" and terms in languages from the community rather than "aves") to include in our inaturalist metadata.

## Alternatives
We could try some kind of automated translation service, but it might not be that helpful. See https://wwf.panda.org/discover/knowledge_hub/teacher_resources/webfieldtrips/scientific_nomenclature/

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

obulat:
I wonder if we could do some data augmentation and get a short summary about the species from Wikipedia and use it as a description. Wikipedia site is also great for looking up the names of species in various languages.

For instance, the title of the first result in the API search filtered by source (https://api.openverse.engineering/v1/images/?source=inaturalist) is _Buteo jamaicensis_. If you search for it in Wikipedia, you get redirected to the [page](https://en.wikipedia.org/wiki/Red-tailed_hawk) titled _Red-tailed hawk_. From there, you can get the name of the species in other languages simply by selecting the language in the sidebar or changing the URL to add the locale prefix. So, the Spanish version is at https://es.wikipedia.org/wiki/Buteo_jamaicensis. 

This is how I usually look up the species names in other languages, it is much better than a dictionary :)

AetherUnbound:
That's true! I wonder if we could have a pre-ingestion step which sets up a lookup table in the `inaturalist` schema based on some public data. Looks like [folks have asked this before](https://biology.stackexchange.com/questions/52990/where-can-i-get-a-file-list-of-the-common-and-scientific-names-of-species) but unfortunately most of the links on that post are dead 😭

obulat:
We could use the data from the Catalog of life. The [downloads](https://www.catalogueoflife.org/data/download) page has a `ColDP` zip archive. 
First, we would need to find the species code in the `NameArchive.tsv`, `46VWN` here:
| col:ID | ... | col:basionymID	| ... |
| --- | --- | --- | --- |
| 46VWN | ... | Neotrigonia margaritacea | ... |	

Then we can find the English "vernacular" name in the `VernacularName.tsv` file using that code. Simplified data example:

| col:taxonID | col:name | col:language | 
| --- | --- | --- |
[...]

Issue #689, opened 2022-08-23
Investigate converting iNaturalist to an incremental DAG


AetherUnbound:
## Description

The iNaturalist DAG (#549) is a different beast from all our other provider scripts in that each run processes a massive amount of data. The first run of this DAG is already proving that the process will take quite a while to complete.

The DAG is presently set up to check when the upstream S3 files have been modified and re-run the entire ingestion. It would be ideal (and likely more performant) if we could run this update _incrementally_, i.e. only process the records that have been updated since the last run. This would either require some marker upstream being changed (like `observed_on` date for instance), or us keeping track of what records we've processed and performing a diff before ingestion. The latter seems like it might be quite intensive both on the data storage and compute end, so we should explore the feasibility of the former first.

## Additional context

CC @rwidom 

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

AetherUnbound:
Thanks to @rwidom's incredible work on #745 (which has now been run successfully) and the fact that the iNaturalist DAG now also [checks for updates to the S3 files before attempting to run again](https://github.com/WordPress/openverse-catalog/blob/af205795b412da9f3d0d7f6d7e066b7d039b1763/openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py#L182-L201), I think this can be closed. The iNaturalist data doesn't have any date fields we could use to do an incremental update, and even if we operated only on new IDs, we might lose any updates that had happened since the previous run. So we'll leave things as-is for now!

PR #745, opened 2022-09-28
iNaturalist in-SQL loading

rwidom:
## Related to
This relates to #689 in that my understanding is that we have not yet completed a full load of iNaturalist data, and if we were going to start loading the data incrementally, we would probably want a full load to start from. In practice, it also runs the cleaning and upsert steps on smaller batches, rather than all at once across 120+ million records.

It also integrates Catalog of Life data to address #696.

## Description
The existing iNaturalist dag essentially follows these steps: load normalized data from s3 to PostgreSQL, pull integrated JSON out in batches of 10,000 joined records, and then follow the rest of the steps in any provider dag, to reload the JSON into PostgreSQL.

The initial load of normalized data takes on the order of 10-15 minutes for approximately 120,000,000 records in `inaturalist.photos', and somewhat fewer in the other inaturalist tables. However, the dag has timed out after several days on the step where joined the data is being pulled out of PostgreSQL. 

This PR moves all of the processing to SQL. The latest local run timed out after 24 hours and 87,943,312 records loaded to `public.image`. I increased the timeout to be 48 hours, but/and we could consider longer. I estimate that it should complete within 36 hours. More on that [on the chart 12-14-2022 tab in this google sheet](https://docs.google.com/spreadsheets/d/1gaVsvFnsYby2iwzRm0Ta9IPhvNixCpQdHByGLo1r_pg/edit#gid=108859884). The workbook also contains visualizations for other test runs during this development process.

### Issues I will make and add to comments for future work

- Improve Airflow reporting, right now it only reports the number of records loaded
- Consider adding indices to the temp loading table _after_ loading data into it.
- Figure out a better way to store batch limit, for DRYer code
- Keep copies of the most recent source files on s3.

## Testing Instructions
I increased the disk image size in my local docker to 192 GB. You should be able to run the whole thing with a very small sample of test data in the repo, and I made some changes to `tests/dags/providers/provider_api_scripts/resources/inaturalist/pull_sample_records.py` to make it easier/faster to pull a test sample of your choice. And then poke around for data quality/formatting stuff. [Instructions to download the inaturalist data.](https://github.com/inaturalist/inaturalist-open-data/tree/main/Metadata/Download) 

I could really use a second set of eyes to double check that I have implemented all of the python data cleaning steps (initial attempt [on the "image table ddl" tab here](https://docs.google.com/spreadsheets/d/1gaVsvFnsYby2iwzRm0Ta9IPhvNixCpQdHByGLo1r_pg/edit#gid=2082506887), in the SQL. After running the large majority of the dataset, the log for the python cleaning step looks promising to me, but without more on #898, a lot of this has to happen manually. 

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


AetherUnbound:
This is some really exciting work! I took a look at your first commit, I think that would probably be the most ideal route. Ultimately that was what we had discussed early on I believe - circumventing the python-based provider data ingestion steps entirely and _only_ doing the operations in SQL. The license table you've set up is probably something we'd want to hard code for all SQL-only loads, and I wonder what other cleaning operations/checks we could standardize in SQL-land which would match what we currently have in python land too. Maintaining that parity will probably be the most difficult piece going forward, but it's worth it if that is the best way of doing bulk imports IMO 🙂

rwidom:
Thanks @AetherUnbound ! To help me make sure that all of the data quality / processing steps in python get moved to SQL here, I made [this sheet](https://docs.google.com/spreadsheets/d/1gaVsvFnsYby2iwzRm0Ta9IPhvNixCpQdHByGLo1r_pg/edit#gid=2082506887) to try to list out where cleaning is happening in the catalog and API repos. Maybe it would help with https://github.com/WordPress/openverse/issues/244 as well. I wonder if you and/or @krysal  would have a second to take a look at it and give me feedback? I haven't finished the API part, but it's a start there. I'm also thinking about the performance issues that @obulat raised, and the impact of having so many "[TOASTed](https://www.postgresql.org/docs/current/storage-toast.html)" fields in such a large table. I'm wondering how to navigate the trade-offs between complete data (e.g. repeating the provider name for each and every tag in the intermediate table), and performance of the load. More to learn there!

openverse_catalog/dags/common/loader/sql.py
```diff
@@ -72,7 +72,7 @@ def create_loading_table(
     columns_definition = f"{create_column_definitions(loading_table_columns)}"
     table_creation_query = dedent(
         f"""
-    CREATE TABLE public.{load_table}(
+    CREATE UNLOGGED TABLE public.{load_table}(
```
AetherUnbound, line 75: Verrrrrry cool 😮

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
```diff
@@ -144,6 +143,73 @@ def compare_update_dates(
         raise AirflowSkipException("Nothing new to ingest")
 
     @staticmethod
+    def load_catalog_of_life_names():
+        COL_URL = "https://api.checklistbank.org/dataset/9840/export.zip?format=ColDP"
+        # Tried importing common.loader.paths.STAGING_DIRECTORY but it didn't work in
+        # local environment, TO DO: find a better place for this process.
+        DATA_DIR = Path(__file__).parents[4]
```
AetherUnbound, line 150: Ah, we no longer have `STAGING_DIRECTORY` there! Best place for it might be [`OUTPUT_DIR`](https://github.com/WordPress/openverse-catalog/blob/b4ef93ce5f825e4ae0e3ba3814340ec42ade7fa2/docker/airflow/Dockerfile#L23)? Although would we want to make sure this gets cleaned up, or have it stick around for the lifetime of the container (e.g. however long it is until the next deployment)? I see you have a note about getting this into S3, would that be more resilient?

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
```diff
@@ -144,6 +143,73 @@ def compare_update_dates(
...
+        local_zip_file = "COL_archive.zip"
+        name_usage_file = "NameUsage.tsv"
+        vernacular_file = "VernacularName.tsv"
+        # download zip file from Catalog of Life
+        if (DATA_DIR / local_zip_file).exists():
+            logger.info(
+                f"No download from Catalog of Life. {DATA_DIR}/{local_zip_file} exists."
+            )
+        else:
+            with requests.get(COL_URL, stream=True) as r:
```
AetherUnbound, line 160: Ooo, this is a new request pattern! Unfortunately we don't have a way to handle streaming yet in the [`DelayedRequester` machinery](https://github.com/WordPress/openverse-catalog/blob/c1b970b11dbf63a45a8c2d6a33d6058a662b7ac9/openverse_catalog/dags/common/requester.py#L103). Something to think about down the line maybe!

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/create_schema.sql
```diff
@@ -13,25 +13,138 @@ https://github.com/inaturalist/inaturalist/blob/d338ba76d82af83d8ad0107563015364
...
+    (1, 'Animals'),
+    (47118, 'Spider'),
+    (47124, 'Flowers'),
+    (47126, 'Plants'),
+    (47157, 'Butterflies and Moths'),
+    (47158, 'Insect'),
+    (47167, 'Mushroom'),
+    (47434, 'Grass'),
+    (47604, 'Daisy'),
+    (48662, 'Monarch Butterfly');
```
AetherUnbound, line 149: Is there any concern these `taxon_ids` could change? Would it be better to map the Latin to the English here and then link that to IDs via a join?

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/taxa.sql
```diff
@@ -3,51 +3,131 @@
...
+            '{"name": "'||string_agg(DISTINCT
+                case
+                when v.name_language <> 'eng' then v.taxon_name
+                end,
+                '", "provider": "inaturalist"}, {"name": "')
+            ||', "provider": "inaturalist"}' tags_nonenglish_vernacular
+        FROM inaturalist.col_name_usage n
+            INNER JOIN inaturalist.col_vernacular v on v.taxonid = n.id
+        where length(n.id) <= 10
+        group by cast(md5(n.scientificname) as uuid)
```
AetherUnbound, line 80: Simple SQL literacy question, could this just be `group by 1`?

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/transformed_table.template.sql
```diff
@@ -33,19 +43,14 @@ INSERT INTO {intermediate_table}
...
-        -- TO DO: should there be a timestamp or anything in the metadata or is null ok?
-        null::json as META_DATA,
-        -- TO DO: confirm format here, is provider name integrated, and if so how?
-        array_to_json(string_to_array(string_agg(
-            INATURALIST.TAXA.ancestor_names,
-            '|')
-        ,'|')) as TAGS,
+        taxa_enriched.title,
+        LICENSE_CODES.license_url_metadata as META_DATA,
+        taxa_enriched.tags,
```
AetherUnbound, line 48: The tag enrichment is _exactly_ the kind of behavior I was thinking we'd need to have SQL equivalents for 😅 This is great for now, but definitely as we start to have more SQL-only ingestions we'll want some mechanism for abstracting this! Maybe a set of defined functions?

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
```diff
@@ -144,6 +143,73 @@ def compare_update_dates(
...
+                r.raise_for_status()
+                with open(DATA_DIR / local_zip_file, "wb") as f:
+                    for chunk in r.iter_content(chunk_size=8192):
+                        f.write(chunk)
```
AetherUnbound, line 164: Really smart idea to chunk this all 😄

AetherUnbound (commented):
Whew, so much of this SQL magic is _way_ over my head 🤯 I'm excited to see you're making awesome progress on this though!! Please feel free to ping us if there's anything you'd like us to look at or test specifically as you chug along 😄 🚋 🔬

openverse_catalog/dags/common/loader/sql.py
rwidom: Yay!!! Right???

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
rwidom: Yeah, I'm not sure. I did a project once that would download tables via an API (kind of like the Catalog of Life data), and store them on S3 under a dated prefix, just because I was a hard core data hoarder, and then Redshift would pretty much only have the current day's data except for very special cases where if we wanted to know when something changed it wasn't impossible to go back through the archive on S3. But I think that's might overkill for here. (?) Maybe it would make sense to have something like a `last_inaturalist_data_load/` prefix in the bucket where the tsv files normally go, where the raw source files could go? And yeah, changing this to `OUTPUT_DIR` sounds great.

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
rwidom: Totally! I thought about trying to use `DelayedRequester`, and I think it does take `kwargs`, so technically it could just pass along the `stream=True`, but then it wouldn't have the rest of the pattern. It seems like it's more designed for a bunch of smaller requests than the one big one, but maybe I should at least be more explicit about retries? Or maybe it would be better to break this into smaller tasks and let airflow handle it?

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/create_schema.sql
rwidom: That is so much a better idea!!!

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/taxa.sql
rwidom: This is such a great question! Yeah, I feel like there are some dialects or configurations where that doesn't work and others where it's more performant, but I think our postgres is one of the latter, so yeah, I'll totally change it.

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/transformed_table.template.sql
rwidom: Yeah...  I worry about navigating costs of context switching with user defined functions in SQL, particularly for functions that will be run many times over a table, but that might just be something I need to learn to do better in postgres. dbt lets you write jinja macros (e.g. [pivot](https://github.com/dbt-labs/dbt-utils/blob/main/macros/sql/pivot.sql)) for this kind of thing so that you're always running something in SQL, but your code can still be DRY. 

And I guess for this one in particular, I have a bit of a fantasy that we will make a separate `tags` table (e.g. with `media_type, provider, identifier, tag_name` and maybe `authority` if that makes sense) and stop trying to handle them in JSON at all.

But yeah, at a high level, I totally agree that this is not ideal for the longer term. I'm adding it to the list at the top of the PR for now.

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
rwidom: Thanks!

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
AetherUnbound: I could see that working! Especially with the cadence this DAG will run at, putting this where we'd store the other data (either TSV or otherwise) sounds like a good call 🙂

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
AetherUnbound: I think this is solid as is, because even with `DelayedRequester`'s retries we'd still need to change a LOT in order to get the context manager + `stream` to work. Something to look at again if it's a pattern we notice though!

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/taxa.sql
AetherUnbound: Nice! I was mostly thinking about not having to declare the same functions on the column, but it'd be even better if that is more performant 🏎️

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
AetherUnbound: It looks like this would be possible _outside_ the context manager, but we would need to ensure the connection is closed once the operation is complete:
```python
r = None
try:
    r = self.delayed_requester.get(COL_URL, stream=True)
    with open(DATA_DIR / local_zip_file, "wb") as f:
        for chunk in r.iter_content(chunk_size=8192):
            f.write(chunk)
finally:
    if r is not None:
        r.close()
```

sarayourfriend (commented):
This is so cool. I don't know enough SQL or Airflow to feel comfortable commenting much on this PR. I have some clarifying questions, however, to help me understand what the end result of this approach is and to help me refine my mental model of how to use Airflow.

With this work completed, does it mean that the iNaturalist ingestion is _entirely_ done in SQL? I'm personally not a super fan of doing everything in SQL mostly because I find it harder to test and maintain compared to Python, but this is just my own personal preference. Other questions I have about it though are, for example, does this make it harder for iNaturalist and other "non-API" based provider DAGs to be ported to a system that doesn't rely on Postgres for catalogue data storage? If we moved to Parquet, for example, would this create a significant difference or would it just add a new required step of exporting the data from the resulting PG tables out to parquet (or other hypothetical non-relational DB technology)?

Finally, this approach goes against my instinct of how to usually increase data throughput performance: parallelisation. I suppose Postgres will handle a _lot_ of that for us by implementing it this way. I wonder though if there is a reasonable alternative where data is initially loaded in PG from the dumps and then instead of using JSON to transport to DAGs, there's a dispatching job that runs through the tables and triggers additional tasks to run that in-parallel transform the data and loads it into its final home?

As I said in another comment, forgive me if my comments are plainly wrong, I'm just trying to get a handle on how best to use Airflow and I love learning about it through these very interesting and complex provider PRs.

Thanks for your excellent work and documentation, Rebecca.

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
```diff
@@ -63,43 +73,72 @@ def get_next_query_params(self, prev_query_params=None, **kwargs):
...
+                f"Inserted {transformed_records} into {intermediate_table}. "
+                f"Last photo_id loaded was {max_id_loaded}, as of {page_log}."
+            )
+            # Run standard cleaning on this batch of records
+            sql.clean_transformed_provider_s3_data(
+                identifier=identifier,
+                postgres_conn_id=postgres_conn_id,
+            )
+            # Add transformed records to the target catalog image table.
+            total_records += sql.upsert_records_to_db_table(
```
stacimc, line 133: Can these be split into individual tasks? I _think_ we could even still use dynamic task mapping with a resulting [TaskGroup](https://docs.astronomer.io/learn/task-groups#generate-task-groups-dynamically-at-runtime)

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
```diff
@@ -63,43 +73,72 @@ def get_next_query_params(self, prev_query_params=None, **kwargs):
...
+        pg = PostgresHook(postgres_conn_id)
+        sql_template = (SCRIPT_DIR / sql_template_file_name).read_text()
+        total_records = 0
+        # get the number of "pages" in the photos table
+        max_id = pg.get_records("SELECT max(photo_id) FROM inaturalist.photos")[0][0]
+        total_pages = int(max_id / page_length) + 1
+        logger.info(f"Loading {intermediate_table}...")
+        logger.info(f"Max photo_id={max_id}. Transform {page_length} ids / iteration.")
+        # then cycle through the pages
+        for page_start in range(0, max_id, page_length):
```
stacimc, line 112: I haven't used this myself yet so this idea would need exploration, but I wonder if this would be a good place to try using [dynamic tasks](https://docs.astronomer.io/learn/dynamic-tasks). Using task mapping, I think we could:
* have a task to get the number of "pages"
* expand a `loader` task to create a separate task instance for each "page"

That would increase the opportunity for parallelization (maybe, depending on the handling of intermediate tables 🤔), but even without concurrency it would be nice to let us clear/retry individual page tasks rather than having one giant task.

openverse_catalog/dags/providers/provider_dag_factory.py
```diff
@@ -302,7 +302,14 @@ def create_provider_api_workflow_dag(conf: ProviderWorkflow):
     )
 
     with dag:
-        ingest_data, ingestion_metrics = create_ingestion_workflow(conf)
+        # @stacimc is this stretching your beautiful default class a bit too far?
```
stacimc, line 305: :) I think it's a great idea as implemented!

Interestingly, I feel like the ingestion workflow isn't _that_ different from the normal one. Or at any rate, there's enough duplication that it's got me thinking of ways we could reuse the logic in `provider_dag_factory` -- especially since it's easy to imagine that we'll eventually add more providers that use this workflow.

This is more the case on `main` after #914, which broke out the `load_from_s3` task into smaller tasks including ones that could be reused here. The `create_loading_table`, `clean_data`, `upsert_data`, `drop_loading_table`, and `report_load_completion` tasks could all be reused, I think.

It would be cool to make `create_ingestion_workflow` a little more configurable by allowing individual steps to be overridden or skipped.  We have several other DAGs right now that are having to 'batch' their ingestion by splitting up the ingestion day into time intervals, that I think something like the dynamic tasks might even be helpful for our 'normal' workflows.

stacimc (commented):
This is incredible progress! I will have to set aside more time to look at the SQL as it's a bit over my head at first glance 😅

I have some early questions/suggestions about how we might reuse some of the existing factory code and possibly splitting up the `load_transformed_data` step. I think it would be a good idea to rebase this asap to bring in #914.

docker-compose.override.yml
```diff
@@ -50,6 +50,9 @@ services:
     image: minio/mc:latest
     env_file:
       - .env
+    environment:
+      MINIO_ROOT_USER: ${AWS_ACCESS_KEY}
+      MINIO_ROOT_PASSWORD: ${AWS_SECRET_KEY}
```
AetherUnbound, line 55: Why did this need to be added? We're adding the credentials as part of the init on L71 and this container doesn't exist for very long IIRC. The `s3` container should be how we access minio which _does_ need these set (and currently does have them set).

openverse_catalog/dags/common/loader/sql.py
AetherUnbound: [According to the documentation](https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-UNLOGGED), the biggest downsides to an unlogged table are that they 1) are not crash resistant and will be truncated on an unclean shutdown and 2) are not replicated. Since this is a transient table (and we don't do replication anyway), we should be able to recover it if postgres shuts down by re-running this task. I don't think there's any additional concern about having this one be unlogged, even if we don't move to some other data storage.

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
```diff
@@ -19,42 +19,52 @@
...
+from common.constants import POSTGRES_CONN_ID, XCOM_PULL_TEMPLATE
+from common.loader import provider_details, reporting, sql
 from providers.provider_api_scripts.provider_data_ingester import ProviderDataIngester
 
 
+logger = logging.getLogger(__name__)
+
 AWS_CONN_ID = os.getenv("AWS_CONN_ID", "test_conn_id")
-PROVIDER = prov.INATURALIST_DEFAULT_PROVIDER
+OPENVERSE_BUCKET = os.getenv("OPENVERSE_BUCKET", "openverse-storage")
```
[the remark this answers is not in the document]
AetherUnbound: Great question! There is a distinction between an _environment variable_ and an _Airflow Variable_ that might be a bit difficult to suss out. We can define Airflow Variables using env vars, but Airflow Variables can also be defined in the UI and overwritten/changed at any time. Additionally, storing secrets (e.g. API keys) in Airflow Variables will [automatically mask them in the logs](https://airflow.apache.org/docs/apache-airflow/stable/security/secrets/mask-sensitive-values.html). Lastly, we can use the `deserialize_json` flag to convert Airflow Variable values to native python types, which was something I mentioned on #904. Airflow Variables can be changed at any time whereas environment variable changes require a deployment (or at least a service restart). There's some more context for this in https://github.com/WordPress/openverse-catalog/issues/209.

> I recently learned in #904 that we can use template strings for a lot of environment variables.

To be clear, the templating that I described in #904 only applies to Airflow Variables, not environment variables. The former we wished to defer because they issued a database query on ever `Variable.get`. Environment variables are much cheaper to retrieve.

I think this particular value is fine to use from the environment because:
1. It does not change readily (or ever), so we would not need the ability to update it without a restart as a Variable.
2. It is [already defined as an environment variable](https://github.com/WordPress/openverse-catalog/blob/f42555b19ea2488feebe7ea01836e652a14158d8/env.template#L104-L103).
3. It is [used directly elsewhere](https://github.com/WordPress/openverse-catalog/blob/f654446ae34bb2351b5e52456404efa8c6002812/openverse_catalog/dags/providers/provider_dag_factory.py#L90) so changing this value would require a deployment/service restart anyway.

I hope that helps, let me know if you think it would be useful to codify some of this into an FAQ or as part of our README/contributing guide.

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
```diff
@@ -135,31 +243,125 @@ def create_preingestion_tasks():
                         f"{file_name}.csv.gz" for file_name in SOURCE_FILE_NAMES
                     ],
                 },
+                doc_md="Has iNaturalist published new data to s3?",
```
AetherUnbound, line 246: Do you mind rephrasing this in a more declarative form? 🙂

AetherUnbound (commented):
I'm so glad to see progress on this! Similarly I'll try to dive into the SQL a bit more, but I wanted to share some initial thoughts.

To respond to @sarayourfriend, I do believe that the implication of merging this would be that most of the transformations would happen in SQL. IIRC one of the constraints/requirements for doing so was that the data provided to us was in a relational form and had to be merged/joined to get the standard view we're used to. Doing so in our existing Postgres database seemed the easiest way to address these constraints. It would pose a problem if we moved to a different data source and I'm glad you called this out - we'd likely face a similar problem for other providers that expose their data in a relational manner like iNaturalist does. That's something we'll need to consider when exploring the other data sources; perhaps we will need a (possibly transient?) postgres database for these tasks.

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/taxa.sql
```diff
@@ -3,51 +3,165 @@
...
+            v.name_language,
+            v.taxon_name,
+            row_number() OVER (partition by n.scientificname
+                order by case when v.name_language='eng' then 0 else 1 end,
+                v.taxon_name)
+                as name_rank,
+            count(case when v.name_language='eng' then 1 end)
+                OVER (partition by n.scientificname
+                rows between unbounded preceding and unbounded following)
+                as english_names
```
AetherUnbound, line 88: Do you mind adding comments to each of these two lines describing what they're doing? I'm having trouble parsing it out 😅

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/taxa.sql
```diff
@@ -3,51 +3,165 @@
...
+        FROM inaturalist.col_name_usage n
+            INNER JOIN inaturalist.col_vernacular v on v.taxonid = n.id
+        where length(n.id) <= 10
```
AetherUnbound, line 91: Is this just getting the first 10 rows? Is the intent here to limit the number of tags we add? If so, do you mind adding that as a note at the top?

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/taxa.sql
```diff
@@ -3,51 +3,165 @@
...
+    (
+        SELECT
+            md5_scientificname,
+            string_agg(DISTINCT
+                case when name_language = 'eng' then taxon_name end,
+                ', ') name_english,
+            jsonb_agg(DISTINCT cast((taxon_name, 'inaturalist') as openverse_tag))
+                FILTER (where name_rank > english_names) as tags_nonenglish_vernacular
+        FROM catalog_of_life_names
+        where name_rank <= 20
```
AetherUnbound, line 103: Similar comment here as above, are we just limiting the number of potential tags so they don't get too large?

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/taxa.sql
```diff
@@ -3,51 +3,165 @@
...
+            (
+                coalesce(child.inaturalist_name_tag, to_jsonb(array[]::openverse_tag[]))
+                || coalesce(jsonb_agg(DISTINCT
+                    cast((ancestors.title,'inaturalist') as openverse_tag))
+                    FILTER (where ancestors.title is not null),
+                    to_jsonb(array[]::openverse_tag[]))
+                || coalesce(child.tags_nonenglish_vernacular,
+                    to_jsonb(array[]::openverse_tag[]))
+            ),
+            '$[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19]')
```
AetherUnbound, line 155: This would seem to match up with the `name_rank <= 20` above, is that the case? If so, can you add a comment here pointing that connection out?

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/taxa.sql
```diff
@@ -3,51 +3,165 @@
...
-)
-UPDATE inaturalist.taxa
-SET ancestor_names = aggregated.ancestor_names
-from aggregated
-where taxa.taxon_id = aggregated.taxon_id;
+        inaturalist.taxa_with_vernacular child
+        left join inaturalist.taxa_with_vernacular ancestors
+        on (ancestors.taxon_id = ANY (string_to_array(child.ancestry, '/')::int[]))
+    group by child.taxon_id, child.title,
+        child.inaturalist_name_tag, child.tags_nonenglish_vernacular
```
AetherUnbound, line 162: I'm having a particularly hard time grokking all these tables 😅 Would it be possible/reasonable to add an ERD as part of the documentation describing the relationships? Or some other way of describing these relationships?

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/taxa.sql
```diff
@@ -3,51 +3,165 @@
...
+COMMIT;
+
+/*
+           ********** Create enriched table with ancestry tags *********
+
+Join each record to all of its ancestor records and aggregate ancestor titles into the
+tags along with json strings from the enriched data above.
+    + expand each ancestry string into an array
+    + get the taxa record for each value of the array
+    + aggregate back to the original taxon level, with tags for ancestor names (titles)
```
AetherUnbound, line 137: Would you be able to provide some examples of data at each of these stages? E.g. `expand each ancestry string into an array ('reptilia/crocodilia/somethingelseia' -> ['reptilia', 'crocodilia', 'somethingelsia'])`.

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
rwidom: Ooooh! I didn't know about loops within DAGs / across tasks. Neat! I'll take a look.

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
```diff
@@ -63,43 +73,72 @@ def get_next_query_params(self, prev_query_params=None, **kwargs):
...
-        return record_data
+        raise NotImplementedError("TSV files from AWS S3 processed in postgres.")
 
     def get_media_type(self, record):
         # This provider only supports Images via S3, though they have some audio files
         # on site and in the API.
-        return "image"
+        # I'm using a constant here, which maybe should just be a class constant, but
+        # it's there to work with the static methods.
+        return MEDIA_TYPE
```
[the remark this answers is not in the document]
rwidom: Oops! Done.

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
rwidom: Right. I guess before I dig in too far, how much do we want to get into parallelism outside of pg basics? I guess I feel like we stopped at ~15 million records in 5 days or something like that with the current code, and this step gets us to ~120 million records in 2 days, maybe we could just call it a win for now, and circle back if needed? I'm definitely curious about dynamically generating tasks, but I'm really trying to avoid space constraints, so right now the one temp table is really treated as a temporary pass through space and emptied out after each page is upserted, which definitely will not work if we're trying to run pages in parallel. Just inaturalist is ~15% of the existing `image` table size and I wasn't able to make holding two copies work locally. And creating separate temp tables per page feels like it would be a big lift in terms of the current code, and maybe not worth it? Though I do see the potential for real reporting and operational wins of separate tasks even if they are run sequentially. ???

AetherUnbound: That's some helpful perspective Rebecca, thanks! I think with that in mind the current changeset is worth merging as it stands, and perhaps that other parallelization can be explored down the road.

stacimc: Definitely agree about parallelization being a v2 concern (_if_ it is necessary at all -- the estimated times you've given are great and it's not at all clear to me that it's worth it yet! 🤩), I mean to suggest that the _opportunity_ for easier future parallelization was one (small) advantage of splitting this up. The much greater immediate advantage in my mind is the ability to retry/inspect individual tasks. I don't think that would require dealing with any of the tricky space constraints or handling parallel load tables.

As it stands, this is such a phenomenal improvement that I do think it would be worth merging on its own. As we move to splitting the typical ingestion flow into smaller tasks (splitting the load and cleaning steps for example), and as we've been dealing with a lot of long-running DAG failures, I'm just slightly cautious about very large tasks.

rwidom: Yeah, I 100% hear you on the opportunity to look at the different tasks, @stacimc. I tried to address that need with some logging, but it's clunky, no question.

I just started to try coding this, and it occurred to me that for it to work, I would need a separate dag for the preingestion tasks rather than a task group within `inaturalist_workflow`. The number of pages (and therefore the number of tasks we would want to see in the dag) is based on the `max(photo_id)` loaded to `inaturalist.photos`, but with a single dag, that load would not have happened yet when airflow puts the dag task list together.

That said, I think that it's probably pretty common practice to separate out staging and target loads into separate dags, and we could have something like `inaturalist_staging_workflow` and `inaturalist_target_workflow`. But I wonder if that could be a v2 improvement?

rwidom:
Hey @sarayourfriend , thanks so much for the candor and thoughtfulness about SQL and the infrastructure implications here.

> With this work completed, does it mean that the iNaturalist ingestion is _entirely_ done in SQL? 

Not exactly. There are still dag steps that use the python operator rather than the postgres operator, and we're still using the existing cleaning and upsert functions. So in that sense, there is still python. But as @AetherUnbound notes, python-only doesn't seem practical for a normalized dataset of this size with our current infrastructure. I will say that in this process I've learned something about the efficiency of sets versus lists in python, which did make pulling a sample dataset with referential integrity much faster, but you know, many humans have been optimizing SQL databases for many decades and I'm not personally going to be able to do better than them for this kind of use case.

> I'm personally not a super fan of doing everything in SQL mostly because I find it harder to test and maintain compared to Python, but this is just my own personal preference.

It's so interesting to me, to think about _what_ we're testing when we're testing python vs SQL. I wonder a little bit if the ease that you're referring to here is really about the coding language, or more about the difference between testing _logic_ and testing _data_. It's definitely easier to test logic with python, but what happens when a provider changes something about the API or the CSVs? Then the sample data that we're using for our python tests becomes obsolete, and the tests aren't as useful either. There's more on this in #898 . 

>Other questions I have about it though are, for example, does this make it harder for iNaturalist and other "non-API" based provider DAGs to be ported to a system that doesn't rely on Postgres for catalogue data storage? 

This is a little tricky to answer in the abstract. In my experience, the devil is always in the details with these things. Here's kind of a high level narrative of the things that made for the performance improvement we're seeing here, and I haven't done a careful enough analysis of which pieces saved the most time:

- switching the chunks to fixed photo ID ranges (with some variation in number of records), rather than asking postgres to figure out how to pull an exact number of records for each chunk;
- moving the dataset between postgres and s3 just once in compressed form, rather than once compressed, and then twice more in a massively expanded json format; and
- switching the temp table to nologging; 

I'm saying this because _the core challenge with iNaturalist is joining/merging two very large tables_. The solution is always going to be some flavor of "break one of the tables into smaller chunks and find the relevant records from the other table", but _how_ that gets done and the levers available to improve performance along the way will be different with different tools. And not all tools are up to the task -- e.g. the files are just too big for pandas. I'm not sure what to do about the chicken-egg problem with infrastructure, other than say that this PR is built to work with the infrastructure we have now. 

>If we moved to Parquet, for example, would this create a significant difference or would it just add a new required step of exporting the data from the resulting PG tables out to parquet (or other hypothetical non-relational DB technology)?

I'm not super familiar with Parquet, but my first quick guess is that yes, the workflow would still have to look something like read stuff from S3, join tables / denormalize it (could be in a temporary postgres instance), and export it to parquet.

> Finally, this approach goes against my instinct of how to usually increase data throughput performance: parallelisation. I suppose Postgres will handle a _lot_ of that for us by implementing it this way. I wonder though if there is a reasonable alternative where data is initially loaded in PG from the dumps and then instead of using JSON to transport to DAGs, there's a dispatching job that runs through the tables and triggers additional tasks to run that in-parallel transform the data and loads it into its final home?

I guess from my perspective if your goal is to tie bows on 100 gifts, you can probably do it faster if you give 10 people 10 gifts and 10 bows each. But you might first want to check in with the people to confirm that they each are just going to tie bows on gifts, as opposed to unpacking each gift, painting it purple, waiting for the paint to dry, repacking it, and then tying the bow on. Forgive me for the weird metaphor, and I love parallelism, but it's not the only lever for performance improvement.

> Thanks for your excellent work and documentation, Rebecca.

Thank you so much for hanging in there with me on this, and creating such a welcoming environment to play with data and code, @sarayourfriend !

docker-compose.override.yml
rwidom: Oops! I must have added it at some point because of some error, but it doesn't seem to be necessary now. Thanks for flagging it!

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
rwidom: Done.

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/taxa.sql
rwidom: I added a bunch of comments, but one key question which might make this all simpler: I got the impression somewhere that the python code I was trying to mimic was taking only the first 20 tags for any image. Did I make that up somehow? Should I just take all the tags? (Just FYI, with vernacular names, a few taxa would have thousands of tags.)

AetherUnbound: Ah, I don't think we have a limit on tags currently! There is this note in Flickr though:

https://github.com/WordPress/openverse-catalog/blob/5ae98b0190c8d0971ac900b053ce1bd60dcaacc8/openverse_catalog/dags/providers/provider_api_scripts/flickr.py#L251-L253

I think 20 is good, but we should definitely note it somewhere!

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
stacimc: Separate DAGs would definitely be a heavy lift, but I think task mapping by itself should accomplish what we want here. Task mapping allows you to generate the task list totally dynamically, so you _don't_ need to know the number of tasks you want to see in the DAG in advance. The [documentation](https://airflow.apache.org/docs/apache-airflow/2.3.0/concepts/dynamic-task-mapping.html#task-generated-mapping) even mentions building the list of parameters from API calls or DB queries. There's really good instructions [here](https://docs.astronomer.io/learn/dynamic-tasks#map-inputs-when-both-tasks-are-defined-with-traditional-operators).

I wouldn't say this is a blocker, but it would be nice to have if it's not too much lift. I played around with this just to make sure I understood and I was able to get a rough version working. Basically you just have a function that gets the list of pages:

```
    @staticmethod
    def get_pages(
        intermediate_table: str,
        page_length: int,
        postgres_conn_id=POSTGRES_CONN_ID,
    ):
        pg = PostgresHook('postgres_conn_id')
        # get the number of "pages" in the photos table
        max_id = pg.get_records("SELECT max(photo_id) FROM inaturalist.photos")[0][0]
        total_pages = int(max_id / page_length) + 1

        # Return the list of pages. Note the weird formatting because this is being passed to op_args, which expects each arg to be a list
        return [[x] for x in range(0, max_id, page_length)]
```

And then a function that takes in `page_start` and just does the processing for a single page:

```
    @staticmethod
    def load_transformed_data(
        page_start,
        intermediate_table: str = '',
        page_length: int = 0,
        postgres_conn_id=POSTGRES_CONN_ID,
        sql_template_file_name="transformed_table.template.sql",
        identifier: str = None,
    ):

        pg = PostgresHook(postgres_conn_id)
        sql_template = (SCRIPT_DIR / sql_template_file_name).read_text()
        total_records = 0
        page_end = page_start + page_length - 1
        page_log = f"page {int(page_start / page_length) + 1}"
        logger.info(f"Starting at photo_id {page_start}, {page_log}")
        (transformed_records, max_id_loaded) = pg.get_records(
...and so on, truncating the table at the end
```

And then when you define the tasks, you use `expand` to dynamically map a bunch of `load_transformed_data` tasks using the result of the `get_pages` task:

```
                from airflow import XComArg

                get_pages = PythonOperator(
                    task_id="get_pages",
                    python_callable=INaturalistDataIngester.get_pages,
                    op_kwargs={ <omitted for brevity> }
                )

                load_transformed_data = PythonOperator.partial(
                    task_id="load_transformed_data",
                    python_callable=INaturalistDataIngester.load_transformed_data,
                    op_kwargs={
                        "intermediate_table": sql._get_load_table_name(
                            LOADER_ARGS["identifier"]
                        ),
                        "page_length": 2_000_000,
                        "postgres_conn_id": LOADER_ARGS["postgres_conn_id"],
                        "identifier": LOADER_ARGS["identifier"],
                    },
                ).expand(op_args=XComArg(get_pages))  # This part expands it using the return_value of get_pages
```

And then in the UI it looked like this for me:

Screen Shot 2022-12-21 at 1 24 25 PM

What do you think? I'll defer to you here 😄 We can certainly iterate.

openverse_catalog/dags/common/loader/sql.py
rwidom: Yeah, great question, @sarayourfriend , I was kind of concerned at first when I saw what a big difference it made, but then... well, what @AetherUnbound  said. :)

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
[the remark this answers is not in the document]
rwidom: Honestly, I'm pretty sure I was just copying what I had seen elsewhere, but I'm super grateful for the explanation here, @AetherUnbound !

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
rwidom: So, I guess I did know about dynamic tasks because we're using them for [pulling raw inaturalist tables](https://github.com/WordPress/openverse-catalog/blob/382192584f6004537985f754f991e4d51cdbfa85/openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py#L146), but I don't think they would work here for pages/chunks of iNaturalist photos, in a single dag. (More on this in a comment below.)

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/taxa.sql
stacimc: +1 for limiting to 20!

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
rwidom: Oh wow, I swear I at least skimmed that article the first time you linked to it, but I totally missed the xcom args part. Yes, this is a million times better. Let's do this. I'll get right on it.

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/taxa.sql
rwidom: Ah, ok and so then the question becomes, is it practical or just absurdly xenophobic to write code to prioritize / select for English language vernacular in those first 20? I thought about just doing some kind of alphabetical order or leaving order out of it entirely (though I need to check if `row_number()` requires an `order by`)... ???

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/taxa.sql
rwidom: Yes. Does the example below ([here](https://github.com/WordPress/openverse-catalog/blob/4f01d3b0378f2f0e33e51ebf92c1a3bcea32cd28/openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/taxa.sql#L147)) help?

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
rwidom: OK, and just to confirm, two follow-up questions:

- we'll use `max_active_tis_per_dag` to make sure that only one of these things is running at a time (see https://docs.astronomer.io/learn/dynamic-tasks), so they don't truncate each other, and it's ok to set it for all dags? and 
- this only applies to tasks, not task groups, so right it still needs to have all the load --> clean --> upsert --> truncate steps within that one function.

True?

stacimc: > we'll use max_active_tis_per_dag to make sure that only one of these things is running at a time

Yep! You can set this _on the actual task_, as in passing it in to the `PythonOperator` for the `load_transformed_data` task definition [Example](https://docs.astronomer.io/learn/airflow-scaling-workers#task-level-airflow-settings). So that way you're only configuring the max active task instances of this _particular_ task, per DagRun.

> this only applies to tasks, not task groups, so right it still needs to have all the load --> clean --> upsert --> truncate steps within that one function.

I think you can actually generate [whole task groups dynamically](https://docs.astronomer.io/learn/task-groups#generate-task-groups-dynamically-at-runtime)... but I think you need to be using the TaskFlow api. I suspect the UI will also get pretty confusing. So I think it makes total sense to leave this as one function for now!

You might've seen this already, but I also found this discussion about [consolidating XComs from dynamic tasks](https://forum.astronomer.io/t/dynamic-task-mapping-xcom-handling/2003/2) which might be helpful for summing the counts from each page.

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
stacimc: Just linking my [other comment](https://github.com/WordPress/openverse-catalog/pull/745/files/5437d7ba599d28c3be8afbcf5a2f5a5b07847106#r1054917849) for clarity that this can definitely be something we look at in a future iteration :)

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/taxa.sql
AetherUnbound: That's....a good question 😅 Do you know if they already appear in some hierarchical order or anything?

rwidom:
Thanks so much for this amazing feedback @sarayourfriend , @stacimc , and @AetherUnbound !

I believe that I've addressed the comments above (Do you think so?), but/and there are a couple of remaining issues I would love some feedback on:

- I made a bunch of changes to the taxa transformation so that the results are more inclusive and the code is more clear (hopefully!). I'm still not 100% sure what would be best from the perspective of elastic search or the UI, though, so would love any thoughts on that stuff.

- I did make a small change so that `common.sql.create_loading_table` returns the loading table name, so that future tasks will be able to use it without calling `common.sql._get_load_table_name`. Though I did not get into redefining the loading table name so that it can be different for different dynamically generated subtasks. How to ensure that loading table names are unique when we want them to be feels like a bigger set of questions to me. So, right now the subtasks exist (and are _so so great_ for testing/monitoring purposes!!!) but it won't work to run them in parallel. :(

- I did think of one more thing to test, in terms of performance, and I realized that this is probably not ready to run as is. :/ Almost all of the test that I've been doing so far have been of inserting new records, since the test db environment always starts with an empty `image` table. But I was doing some analysis related to the taxa stuff and I wanted to pull a single `image` record by `foreign_identifier` and it took more than 5 minutes to do so, and that made me realize, oh wait, how long will the upsert take if there are already inaturalist records in the image table? So I tested and the first batch which previously had taken 5 minutes, now took 49 minutes. I have some ideas about ways to address this (including running batches in parallel!), but given the infrastructure questions above, I'm wondering how much sense it would make to focus on that right now.

OK, and I know it's a holiday, and I hope that all of you are getting all of the vacation and chillaxing you so deserve! But this is the latest from me.

--R

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/taxa.sql
```diff
@@ -63,60 +64,82 @@ COMMIT;
...
+            /*
+            name_string_length -- Cumulative length of the title string (taxon names
+                plus comma and space) up to and including the current name, within each
+                scientific name. There is no hierarchy / priority field within the
+                vernacular table, so the names are ordered with English names first, and
+                then alphabetical order.
+                In the target table the title can be no more than 5,000 characters long.
+                So, if there are vernacular names that would make the title more than
+                5,000 characters they will be used as tags instead of title below.
+            */
```
AetherUnbound, line 96: This is SO helpful, thank you!

tests/dags/providers/provider_api_scripts/test_inaturalist.py
```diff
@@ -107,3 +107,39 @@ def test_get_media_type():
     expected = "image"
     actual = INAT.get_media_type({"some test": "data"})
     assert actual == expected
+
+
+ti_mock = mock.MagicMock(spec=TaskInstance)
```
AetherUnbound, line 112: Probably best to have this as a fixture so it's recreated on every test! If we wanted to check call count or anything, we'd get results from all the previous tests that used this object too.

Alternatively, since you're just using it in the test below, you can probably instantiate it within the test and remove it from the parametrizations 🙂

tests/dags/providers/provider_api_scripts/test_inaturalist.py
```diff
@@ -107,3 +107,39 @@ def test_get_media_type():
     expected = "image"
     actual = INAT.get_media_type({"some test": "data"})
     assert actual == expected
+
```
AetherUnbound, line 110: Do you mind also adding tests for the new `get_batches` function?

AetherUnbound (commented):
Rebecca, this is **awesome**!

The taxonomy docs you added are SO helpful, they make it much clearer for me.

I see what you mean about the parallelization regarding table names. That's unfortunate, _but_ I feel like the mapped instance setup does put us in a better position for parallelizing it down the line! I don't want that to be another barrier for merging this.

 I'm running this locally with the full data on my desktop to see what it all looks like, I'll update later with the results! I may also try a second run to see what affects the upsert (vs insert) might have. I have a few small comments, barring anything concerning about the local testing this is almost good to go!

AetherUnbound:
Okay I tried the full dataset locally and it got 2 hours in then failed due to my disk running out of space 😅 all in all it had gotten to batch 23 and had created 66GB worth of data. Our production instance only has 64GB of disk space (at this time at least) for produced TSV files. Rebecca, do you have any sense of how much space would be needed to process everything? Have you been able to do a whole run successfully?

rwidom:
> Okay I tried the full dataset locally and it got 2 hours in then failed due to my disk running out of space 😅 all in all it had gotten to batch 23 and had created 66GB worth of data. Our production instance only has 64GB of disk space (at this time at least) for produced TSV files. Rebecca, do you have any sense of how much space would be needed to process everything? Have you been able to do a whole run successfully?

Hmmm... not a totally full answer?

- The raw normalized compressed files are on the order of 8GB, and I wouldn't be surprised if our transformations multiply that space by 8. _However_, this is all happening within the database, not on the airflow server where tsv files sit during transformations, so maybe that's ok? Or I'm being silly and of course you're talking about free db space now?
- I run this with 192GB for the docker disk image locally.
- I have loaded everything into the temp table successfully once, but I haven't finished a full run yet. On 12/14, I got ~88 million records (~75% done, though I was using June data which is old now) into `public.image` before it timed out at 24 hours. I'm downloading the latest data now, and will kick it off again tonight and see if I can be patient enough this weekend. :)

rwidom:
I'm tracking my test runs [here](https://docs.google.com/spreadsheets/d/1gaVsvFnsYby2iwzRm0Ta9IPhvNixCpQdHByGLo1r_pg/edit#gid=1426633184). I haven't done a full load to `image` locally yet, and I'm not 100% sure what's going on with docker that made airflow come down on this most recent run. But, this last commit has a couple of improvements that might help:

- I realized that the catalog of life data does have an indicator for which country/ies use a particular vernacular name. It's only populated about 2/3 of the time, but I'm guessing / hoping that the number of countries is a reasonable proxy for the popularity/importance of the vernacular name.
- Given that, I felt more comfortable limiting the titles to 255 characters rather than 5,000, which depending on how many photos the longer-titled taxa (and their children taxa) have, could save a lot of space.

But, on this test run, airflow completely crashed before the DB had a chance to run out of space. :/ Maybe you can test it locally again and/or offer advice @AetherUnbound ?

AetherUnbound:
Ah okay, that's all helpful! I had assumed that it was the Airflow disk space being filled up because of the TSVs, but on second glance that seems incorrect since we're loading directly from PG->PG now! The production database has _plenty_ of space so that shouldn't be a problem 😄 It's odd that Airflow went down on that last run. Would you be able to try another and see what happens? If the scheduler crashes again, I'd love to see the last few lines of output of `just logs scheduler` to see what might have caused the scheduler to crash.

rwidom:
> Would you be able to try another and see what happens? If the scheduler crashes again, I'd love to see the last few lines of output of `just logs scheduler` to see what might have caused the scheduler to crash.

Yeah, I can give it a try this weekend, just it would have to wait until then, because I need to kind of leave my computer alone for a couple of days while it's running locally. I did try `just logs` but I didn't add scheduler, so it might have been confused. I think I used to have all of airflow in one container, and I wouldn't think that would matter here, but maybe it does?

krysal:
I just started reviewing this, and I have to say, this is such an incredible work of investigation and improvements! I haven't read the whole thread or code yet but observing the following make it very promising:

> - switching the chunks to fixed photo ID ranges (with some variation in number of records), rather than asking postgres to figure out how to pull an exact number of records for each chunk;
> - moving the dataset between postgres and s3 just once in compressed form, rather than once compressed, and then twice more in a massively expanded json format; and
> - switching the temp table to nologging;

I ran the DAG as is and I got 14 rows in the `image` table, which I assume is the current test data. Could you add more instructions on how to test it with a bigger/more updated iNaturalist dataset? I downloaded `inaturalist-open-data-latest.tar.gz`. Should I copy the files into the `/tests/s3-data/inaturalist-open-data` as specified in #549? Truly apologize for the wait; lot of urgent things appeared near the end of last year's holidays, but thank you a lot for the hard work put in here, I'm eager to deep through it!

AetherUnbound:
> I did try `just logs` but I didn't add scheduler, so it might have been confused. I think I used to have all of airflow in one container, and I wouldn't think that would matter here, but maybe it does?

It used to be all in one container! But now it's spread across `webserver` and `scheduler` as part of the changes from #874. When docker spits out its logs, I've found that it tends to split them out sometimes _by service_ rather than _chronologically_ (which can be very frustrating for debugging!). `just logs scheduler` makes sure that only the one services logs are shown which helps a bit in piecing things apart (if the problem is actually the scheduler that is 😅).

rwidom:
I added comments with more details on testing. And then, it occurred to me that I sometimes comment out the post-ingestion parts of the dag for testing, so that I can compare the raw inaturalist and catalog of life tables against the target image table. Maybe I'll take a look at how to set up a run time variable in airflow to skip those steps for testing. Hmmmm... Overkill? Helpful? What do you think @krysal and @AetherUnbound ?

AetherUnbound:
Ah, good question! While that would be possible certainly, I think it's okay to just add that as an additional comment to the most recent testing documentation you added 🙂

rwidom:
I started looking at how to do use a runtime [parameter](https://airflow.apache.org/docs/apache-airflow/stable/concepts/params.html) today, just because I was curious, and in the process realized that I had commented out the code to delete the catalog of life downloads. :/ So this uses the parameter for both the catalog of life files, _and_ the inaturalist db schema. But it also raises the question -- like catalog of life could be it's own dag in some ways. Anyway, I don't know, I can roll back this commit. I have some concerns about side effects given that I had to set it at the dag level rather than the task level. Or maybe instead I should move the COL file removal into its own separate post-ingestion task, so that the parameter can just be for the ShortCircuit step rather than the whole dag... ???

AetherUnbound:
Ooo, very cool! I think your current implementation looks great - defaulting to prod behavior (removing the files) was all I wanted to assure and it seems that's the case 🙂

rwidom:
I finished a full local run with data downloaded on 1/1/2023. Some details and visualizations are [here](https://docs.google.com/spreadsheets/d/1gaVsvFnsYby2iwzRm0Ta9IPhvNixCpQdHByGLo1r_pg/edit#gid=2030685519).

```
[2023-01-09, 11:30:46 UTC] {python.py:177} INFO - Done. Returned value was: 
*DAG*: `inaturalist_workflow`
*Date range*: _all_
*Duration of data pull tasks*: 1 day, 21 hours, 57 mins, 46 secs
*Number of records upserted per media type*:
  - `image`: 152,187,801
```

Reflections:
- The run paused twice because the database ran out of space. Each time, I did `just down` (no `-v`!!!), went into docker resources to increase space, then `just up`, and worked great (64GB --> 192GB --> 296GB). Pretty awesome.
- I ran [this super helpful sql](https://wiki.postgresql.org/wiki/Disk_Usage) to get a sense of how much space the inaturalist data would take up in the prod database (top row of output below). Total space is ~200GB, or on the order of 20 times the size of the raw (normalized, compressed) files. 

```
+--------------------+--------------------------+--------------+------------+------------+------------+------------+------------------------+
| table_schema       | table_name               | row_estimate | total      | index      | toast      | table      | total_size_share       |
|--------------------+--------------------------+--------------+------------+------------+------------+------------+------------------------|
| public             | image                    | 143581550.0  | 198 GB     | 50 GB      | 8192 bytes | 148 GB     | 0.9999561837655186     |
```

- I'm not sure that I really understand how postgres is counting toast data here or in production, and I want to look at [this](https://www.enterprisedb.com/blog/postgresql-vacuum-and-analyze-best-practice-tips) to see if it can help me at all. 
- This was still starting from an empty image table. I am thinking to try an additional run with the tiny test dataset to get a sense of how updates might go, though I want to give the database a minute to finish any automatic processes after the load. Question: have we considered [partitioning](https://www.postgresql.org/docs/9.1/ddl-partitioning.html) the image table by provider?

And, I guess bigger picture, I keep coming back to @sarayourfriend 's question about infrastructure. There are some important differences between document-based systems and relational systems. My experience is really with relational systems, but I'm psyched enough to learn more about document-based ones that I wouldn't want to sound like I'm really pulling for one over the other. I don't think that either is a magic pill, both can be used more or less effectively, etc etc, yada yada. But to the extent that we're staying in a relational db / Postgres, I think we really might want to consider ways we could save space, increase consistency, and make updates faster, like with a small license table (rather than repeating the license url on each record) and using the existing provider field in the image table (rather than repeating the provider name with each and every single tag).

AetherUnbound (approved):
That info is _so_ helpful, thank you @rwidom!! I'm feeling good about this based on your data, and your changes to the tests look great! Note that there are likely to be some conflicts with #939 because of the executor changes. I think I'm going to merge that PR shortly so if it's OK with you I'll also rebase this PR and address any merge conflicts that might come up with that one.

**Thank you** for all your continued efforts on this PR! It's exciting to think how much we'll be increasing our catalog's data size 💥

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/transformed_table.template.sql
```diff
@@ -0,0 +1,78 @@
...
+            as DIRECT_URL,
+        -- TO DO #810: Add the thumbnail url here.
+        null::varchar(10) as THUMBNAIL,
+        -- only jpg, jpeg, png & gif in 6/2022 data, all in extensions.py for images
+        lower(INATURALIST.PHOTOS.EXTENSION) as FILETYPE,
+        null::int as FILESIZE,
+        INATURALIST.LICENSE_CODES.OPENVERSE_CODE as LICENSE,
+        INATURALIST.LICENSE_CODES.LICENSE_VERSION,
+        COALESCE(INATURALIST.OBSERVERS.LOGIN, INATURALIST.PHOTOS.OBSERVER_ID::text)
+            as CREATOR,
```
krysal, line 43: The name of the observer will be preferred over their username.
```suggestion
        COALESCE(INATURALIST.OBSERVERS.NAME, INATURALIST.OBSERVERS.LOGIN)
            as CREATOR,
```

krysal (approved):
After taking the time to review this in detail, I have to say this is **impressive**! Watching all the parts involved, like [importing from the S3 CSV files](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_PostgreSQL.S3Import.html#aws_s3.table_import_from_s3), the communication between different tasks in the DAG, the use of SQL templates, and even more! (I won't be able to name them all). Seriously huge kudos for transforming iNaturalist to almost a purely SQL load DAG provider 🎉

I left a few comments on the fields used that are quick to fix, so I don't want to hold more onto this PR! Just being able to finish a DAG run is amazing 😄 and as you already noted, we can work on improvements in separate issues/PRs. For Example, one of the things we introduced a while was [standardized](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/extensions.py) [filetypes](https://github.com/WordPress/openverse-catalog/blob/3b1c4db61858e8f77245f73da594504f8b883d46/openverse_catalog/dags/common/storage/media.py#L278-L289), would be nice to port that to SQL, and avoid having mixed `jpg` `jpeg` and `JPG` but it's no super necessary right now.

openverse_catalog/dags/providers/provider_csv_load_scripts/inaturalist/transformed_table.template.sql
rwidom: Ooh! Great catch! I think I must have had a left join to observers in a prior version and maybe I was worried about some names being truncated or something? I don't know. Hopefully my fix doesn't over-complicate things.

openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
```diff
@@ -1,105 +1,185 @@
...
+from common.loader import provider_details, reporting, sql
 from providers.provider_api_scripts.provider_data_ingester import ProviderDataIngester
 
 
+logger = logging.getLogger(__name__)
+
 AWS_CONN_ID = os.getenv("AWS_CONN_ID", "test_conn_id")
-PROVIDER = prov.INATURALIST_DEFAULT_PROVIDER
+OPENVERSE_BUCKET = os.getenv("OPENVERSE_BUCKET", "openverse-storage")
+PROVIDER = provider_details.INATURALIST_DEFAULT_PROVIDER
```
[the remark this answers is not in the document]
rwidom: Oops! Removed.

PR #746, opened 2022-09-29
Refactor Freesound to use ProviderDataIngester

### LINKED ISSUE #586: Refactor Freesound to use ProviderDataIngester
## Suggested Improvement

Refactor the Freesound provider script to use the new ProviderDataIngester base class

## Benefit

More details in #229

## Implementation

AetherUnbound:
## Fixes

Fixes #586 by @stacimc

## Description

This PR refactors the Freesound provider script to use the `ProviderDataIngester` class.

It's a pretty standard refactor, aside from some more unique functions that have additional retries built in.

One thing I noticed is that the original provider script is set up to allow processing a subset of data, but it has been run in such a way that it consumes the entire dataset. I think this is fine for now, but **as soon as we have a full, successful run**, this should be turned into a daily dated DAG. This will be a much more efficient way to run the DAG! Getting a successful run will require some additional error handling for failures that have occurred in the past, which I'm hoping to enumerate in another ticket.

## Testing Instructions

1. `just recreate && just test`
1. Run the DAG locally and verify that it respects whatever ingestion limit is set (note that there are 3 licenses it iterates through, so that ingestion limit will be hit 3 times)

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/freesound.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/freesound.py
@@ -85,6 +85,42 @@ class FreesoundDataIngester(ProviderDataIngester):
         return None
 
 
+    @staticmethod
+    def _get_creator_data(item):
+        if creator := item.get("username"):
+            creator = creator.strip()
+            creator_url = f"https://freesound.org/people/{creator}/"
+        else:
+            creator_url = None
+        return creator, creator_url
+
+
+    @staticmethod
+    def _get_metadata(item):
+        metadata = {}
+        fields = [
+            "description",
+            "num_downloads",
+            "avg_rating",
+            "num_ratings",
+            "geotag",
+            "download",
+        ]
+        for field in fields:
+            if field_value := item.get(field):
+                metadata[field] = field_value
+        return metadata
+
+
+    @staticmethod
+    def _get_license(item):
+        item_license = get_license_info(license_url=item.get("license"))
+
+        if item_license.license is None:
+            return None
+        return item_license
+
+
 def main(date="all"):
     """This script pulls the data for a given date from the Freesound,
     and writes it into a .TSV file to be eventually read
@@ -285,39 +321,6 @@ def _get_audio_files(media_data):
     main_file["audio_url"] = main_file.pop("url")
     main_file["filesize"] = _get_audio_file_size(main_file["audio_url"])
     return main_file, alt_files
-
-
-def _get_creator_data(item):
-    if creator := item.get("username"):
-        creator = creator.strip()
-        creator_url = f"https://freesound.org/people/{creator}/"
-    else:
-        creator_url = None
-    return creator, creator_url
-
-
-def _get_metadata(item):
-    metadata = {}
-    fields = [
-        "description",
-        "num_downloads",
-        "avg_rating",
-        "num_ratings",
-        "geotag",
-        "download",
-    ]
-    for field in fields:
-        if field_value := item.get(field):
-            metadata[field] = field_value
-    return metadata
-
-
-def _get_license(item):
-    item_license = get_license_info(license_url=item.get("license"))
-
-    if item_license.license is None:
-        return None
-    return item_license
 
 
 if __name__ == "__main__":
--- a/openverse_catalog/dags/providers/provider_api_scripts/freesound.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/freesound.py
@@ -8,8 +8,9 @@ Output:                 TSV file containing the image, the respective
 
 Notes:                  https://freesound.org/apiv2/search/text'
                         No rate limit specified.
+                        This script can be run either to ingest the full dataset or
+                        as a dated DAG.
 """
-import copy
 import functools
 import logging
 from datetime import datetime
@@ -65,6 +66,14 @@ class FreesoundDataIngester(ProviderDataIngester):
     def get_next_query_params(self, prev_query_params: dict | None, **kwargs) -> dict:
         license_name = kwargs.get("license_name")
         if not prev_query_params:
+            start_date = "*"
+            # Allow self.date to be undefined, necessary for the first full, successful
+            # run of Freesound but can be changed to dated afterwards.
+            # TODO: Madison you need to mention this in the PR description :)
+            if self.date:
+                start_date = datetime.strftime(
+                    datetime.fromisoformat(self.date), "%Y-%m-%dT%H:%M:%SZ"
+                )
             return {
                 "format": "json",
                 "token": self.api_key,
@@ -74,6 +83,7 @@ class FreesoundDataIngester(ProviderDataIngester):
                 "filesize,bitrate,bitdepth,duration,samplerate,pack,username,"
                 "num_downloads,avg_rating,num_ratings,geotag,previews",
                 "license_name": license_name,
+                "filter": f"created:[{start_date} TO NOW]",
                 "page": 1,
             }
         else:
@@ -81,9 +91,9 @@ class FreesoundDataIngester(ProviderDataIngester):
 
     def get_batch_data(self, response_json):
         if response_json:
-            return response_json.get("results")
+            # Freesound sometimes returns results that are just "None", filter these out
+            return [item for item in response_json.get("results") if item is not None]
         return None
-
 
     @staticmethod
     def _get_creator_data(item):
@@ -93,7 +103,6 @@ class FreesoundDataIngester(ProviderDataIngester):
         else:
             creator_url = None
         return creator, creator_url
-
 
     @staticmethod
     def _get_metadata(item):
@@ -111,7 +120,6 @@ class FreesoundDataIngester(ProviderDataIngester):
                 metadata[field] = field_value
         return metadata
 
-
     @staticmethod
     def _get_license(item):
         item_license = get_license_info(license_url=item.get("license"))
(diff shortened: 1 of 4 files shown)


openverse_catalog/dags/providers/provider_api_scripts/freesound.py
```diff
@@ -8,294 +8,249 @@
...
-):
-    default_query_params = default_query_params or DEFAULT_QUERY_PARAMS
-    return {
-        **default_query_params,
-        "page": str(page_number),
-        "license": license_name,
+
+class FreesoundDataIngester(ProviderDataIngester):
+    batch_limit = 150
+    RETRIES = 3
```
stacimc, line 33:
```suggestion
    retries = 3
```
But actually this can be removed, since that's the default number of retries!

openverse_catalog/dags/providers/provider_api_scripts/freesound.py
```diff
@@ -8,294 +8,249 @@
...
+    providers = {"audio": prov.FREESOUND_DEFAULT_PROVIDER}
+    flaky_exceptions = (SSLError, ConnectionError)
+    preview_bitrates = {
+        "preview-hq-mp3": 128000,
+        "preview-lq-mp3": 64000,
+        "preview-hq-ogg": 192000,
+        "preview-lq-ogg": 80000,
     }
 
+    HEADERS = {
```
stacimc, line 45:
```suggestion
    headers = {
```
Currently this isn't used anywhere.

openverse_catalog/dags/providers/provider_api_scripts/freesound.py
```diff
@@ -8,294 +8,249 @@
...
-            page_number=page_number,
-        )
-        batch_data = _get_batch_json(query_param=query_param)
-        if batch_data:
-            item_count = _process_item_batch(batch_data)
-            page_number += 1
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.api_key = Variable.get("API_KEY_FREESOUND")
+        self.headers = {"api_key": self.api_key}
```
stacimc, line 52: It looks like the api key is passed in through the query params for Freesound and does not need to be sent as a header. This also overrides the `"Accept": "application/json"`

openverse_catalog/dags/providers/provider_api_scripts/freesound.py
```diff
@@ -8,294 +8,249 @@
...
+                )
+            return {
+                "format": "json",
+                "token": self.api_key,
+                "query": "",
+                "page_size": self.batch_limit,
+                "fields": "id,url,name,tags,description,created,license,type,download,"
+                "filesize,bitrate,bitdepth,duration,samplerate,pack,username,"
+                "num_downloads,avg_rating,num_ratings,geotag,previews",
+                "license_name": license_name,
```
stacimc, line 84: It looks like this query param should be 'license' rather than 'license_name'

openverse_catalog/dags/providers/provider_api_scripts/freesound.py
```diff
@@ -8,294 +8,249 @@
...
-    # for download (and requires a user to be authenticated to download)
-    try:
-        main_audio, alt_files = _get_audio_files(media_data)
-    except FLAKY_EXCEPTIONS:
-        logger.warning(f"Unable to get file size for {foreign_landing_url}, skipping")
-        return None
-    if main_audio is None:
+    def get_batch_data(self, response_json):
+        if response_json:
+            # Freesound sometimes returns results that are just "None", filter these out
```
stacimc, line 93: Nice, I like handling this here!

openverse_catalog/dags/providers/provider_api_scripts/freesound.py
```diff
@@ -8,294 +8,249 @@
...
+            }
+        ]
+        previews = media_data.get("previews")
+        # If there are no previews, then we will not be able to play the file
+        if not previews:
+            return None
+        main_file = self._get_preview_filedata(
+            "preview-hq-mp3", previews["preview-hq-mp3"]
+        )
+        main_file["audio_url"] = main_file.pop("url")
```
stacimc, line 193: I can see this was in the old code, but I'm not sure why `_get_preview_filedata` doesn't just name `url` `audio_url`?

stacimc (changes requested):
Looking very good! Just a few minor things, mostly the `license` param -- although I did get data locally, so that's kind of interesting!

> I think this is fine for now, but as soon as we have a full, successful run, this should be turned into a daily dated DAG.

This is really exciting! 😍

openverse_catalog/dags/providers/provider_api_scripts/freesound.py
obulat: I'm not 100% sure, but I think it's because the URL field is called `url` for `alt_files`, and `audio_url` for the main audio in the AudioStore. And `_get_preview_filedata` is mainly used for populating the `alt_files`, not for the `main` file.

openverse_catalog/dags/providers/provider_api_scripts/freesound.py
AetherUnbound: Oops! Thanks for catching this 😄

openverse_catalog/dags/providers/provider_api_scripts/freesound.py
AetherUnbound: It's definitely `audio_url` in the `AudioStore` for the main file:

https://github.com/WordPress/openverse-catalog/blob/4a9c008a150b2fc5a88b8f9c64f5bd5beef66ac9/openverse_catalog/dags/common/storage/audio.py#L46

I checked the database and it does look like we're using `url` in the `alt_files` dict:
```
deploy@localhost:openledger> select url, alt_files from audio where provider='freesound' limit 1;
-[ RECORD 1 ]-------------------------
url       | https://freesound.org/data/previews/100/100041_1578278-hq.mp3
alt_files | [{"url": "https://freesound.org/apiv2/sounds/100041/download/", "bit_rate": "0", "filesize": "5056680", "filetype": "wav", "sample_rate": "44100"}]
SELECT 1
```

openverse_catalog/dags/providers/provider_api_scripts/freesound.py
```diff
@@ -8,294 +8,247 @@
...
+        return constants.AUDIO
+
+    def ingest_records(self, **kwargs):
+        for license_name in [
+            "Attribution",
+            "Attribution Noncommercial",
+            "Creative Commons 0",
+        ]:
+            logger.info(f"Obtaining audio records under license '{license_name}'")
+            super().ingest_records(license_name=license_name, **kwargs)
```
krysal, line 62: AFAIK Freesound provides only CC-licensed tracks so is this custom code needed at all? They also [mention](https://freesound.org/help/faq/#licenses) they used to include the [Sampling+ license](http://creativecommons.org/licenses/sampling+/1.0/) (now retired) and we have ingested a few of those as well but now will be omitted here.

What do you think?

krysal (commented):
I let run the DAG for +10 minutes and got nothing, not sure why. I have an API key and confirmed it worked by doing a few requests with curl.

AetherUnbound:
Huh, I was able to run this just fine locally 😮 Can anyone else with an API key confirm that they're receiving records when they run this @WordPress/openverse-catalog?

openverse_catalog/dags/providers/provider_api_scripts/freesound.py
AetherUnbound: I'm glad you pointed this out! I looked more into their [API's `filter` notes](https://freesound.org/docs/api/resources_apiv2.html?highlight=license#request-parameters-text-search-parameters), and apparently we're using this entirely wrong! The `license` param is completely ignored in the request (I tried `Attribution` locally and saw that I got back a ton of different license types within the first 25 results:

```
In [14]: {x["license"] for x in r[0]}
Out[14]: 
{'http://creativecommons.org/licenses/by-nc/3.0/',
 'http://creativecommons.org/licenses/by/3.0/',
 'http://creativecommons.org/publicdomain/zero/1.0/',
 'https://creativecommons.org/licenses/by-nc/4.0/',
 'https://creativecommons.org/licenses/by/4.0/'}
```

As you mention, I don't think this logic is needed at all (and we obviously aren't even using it presently). I think it'd be best to remove it, but we could also create an issue for the stability milestone to have this filter by license type correctly. Given the nature of the script I don't think there's any advantage to doing so. What do you think @WordPress/openverse-catalog?

krysal:
I tried rerunning it for some 20 minutes and got the results! I was too impatient 😆 So looks great! As I mentioned early, I just think we can avoid the whole license loop given all the tracks are CC licensed.

AetherUnbound:
Thanks for checking @krysal! I've gone ahead and removed that logic - if folks have any concerns though, feel free to voice them 🙂

openverse_catalog/dags/providers/provider_api_scripts/freesound.py
stacimc: I can see that the AudioStore expects `audio_url`, and that `alt_files` uses `url` (notably the `audio_url` also gets converted to `url` for the main file in the actual DB which can be a little confusing, I opened #784 for it).

But it looks like `_get_preview_filedata` is only ever used once, to get the url and other data for the main file? I don't see why it couldn't just return the key as `audio_url` so it doesn't have to get renamed here. Unless I'm missing another place where it's being used.

openverse_catalog/dags/providers/provider_api_scripts/freesound.py
stacimc: It makes sense that it wasn't doing anything, especially considering we were previously using `license_name` as the param 😮 I guess the intention must have been to add it to the filter? But even so, you're right, it would be unnecessary!

I think removing this as part of this PR is completely fine, because as you point out it doesn't change the _behavior_ of the script at all 😄 Well spotted!

openverse_catalog/dags/providers/provider_api_scripts/freesound.py
AetherUnbound: OH, I see what you're saying now, apologies! I'm not sure it's necessary either, I'll rework this a bit so it's clearer!

krysal (approved):
Exellent!

stacimc (approved):
🎉 Looks great! 🎉

PR #747, opened 2022-09-29
Refactor Phylopic to use ProviderDataIngester

### LINKED ISSUE #591: Refactor Phylopic to use ProviderDataIngester
## Suggested Improvement

Refactor the phylopic provider script to use the new ProviderDataIngester base class

## Benefit

More details in #229

## Implementation

AetherUnbound:
## Fixes

Fixes #591 by @stacimc

## Description

This PR refactors the Phylopic provider script to use the new `ProviderDataIngester` class. 

This refactor is actually pretty significant because the query params are **not** the defining feature for determining the values to retrieve, the `endpoint` is! I had to usurp existing functions a bit in order to get this all to work with the ingester class. Let me know if the current setup is too confusing or difficult to work with. Frankly we can strip out the logic for processing the entire dataset if we'd like and just make this a dated DAG. 

## Testing Instructions

1. `just recreate && just test`
1. Set the class's `delay` attribute to `0.5` and turn the DAG on. It will take about 5 runs (due to catchup), but the 5th run should actually capture data.
1. Run `just run python openverse_catalog/dags/providers/provider_api_scripts/phylopic.py --date-start 2020-11-15`, this should ingest 6 records
1. (Optionally) Run `just run python openverse_catalog/dags/providers/provider_api_scripts/phylopic.py`, this should attempt to ingest everything, but let it go for only a few iterations before stopping

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -13,272 +13,282 @@
...
+    providers = {"image": prov.PHYLOPIC_DEFAULT_PROVIDER}
+    batch_limit = 25
+    # Default number of days to process if date_end is not defined
+    default_process_days = 7
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # This is made an instance attribute rather than passed around via query params
+        # because it actually comprises the URL we hit (not the params) depending on
+        # whether we're running a dated DAG or not.
```
sarayourfriend, line 39: Nit: I think the word comprise is not used correctly here. This confused me for a bit (had to look up the word to make sure I was understanding correctly). Is this more accurate?

```suggestion
        # This is made an instance attribute rather than passed around via query params
        # because the URL we hit includes it in the path (not the params) depending on
        # whether we're running a dated DAG or not.
```

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -13,272 +13,282 @@
...
+        Given an ISO formatted date string and a number of days, compute the
+        ISO string that represents the start date plus the days provided.
+        """
+        date_end = date.fromisoformat(date_start) + timedelta(days=days)
+        return date_end.isoformat()
+
+    def get_media_type(self, record: dict) -> str:
+        return constants.IMAGE
+
+    def get_response_json(
```
sarayourfriend, line 54: Would it makes sense for `get_next_query_params` to still just return a configuration object that then `get_response_json` converts into a usable endpoint to call `super().get_response_json` with?

I think that might tighten up the implementation and make it seem a little more "normal" and easier to understand. At least I think it would from my perspective, coming into this out of pure curiosity :slightly_smiling_face: 

Another thing that crossed my mind was whether some flag other than empty query params (maybe a sentinel object?) could be used to indicate that the batch operation was complete. It would make the process more explicit for other scripts too as well as allowing this one to avoid the awkward dict copying that requires a whole extra comment to explain it.

stacimc (commented):
I haven't gotten a chance to review this closely yet, but I'm curious if you know why the script supports PhyloPic being run as a dated DAG _and_ as "non-dated"? My assumption is that this is run weekly as a dated dag and processes 7 days at a time, but the intention is that a reingestion workflow would need to be able to run it for single individual days as normal.

I'm not sure why it's doing that as opposed to just running on a `@daily` schedule, for a single day like other dated DAGs 🤔 It seems like this introduces a lot of complexity. Were you able to find a reason that it was originally implemented that way?

Obviously this isn't introduced by your refactor! But curious if most of the complexity is coming from that, or if that's incidental and it's mostly the way the endpoints are structured?

AetherUnbound:
After looking at the Stocksnap refactor (#601) for some other reasons, I think Rebecca's approach of having `endpoint` be a computed property is probably the best way to go! I also want to incorporate some of the other feedback provided, so I'll be taking a second look at this next week and making further changes to the `endpoint` setup 🙂

AetherUnbound:
I should be able to get to this tomorrow but I'm going to go ahead and draft it until it's ready!

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
AetherUnbound: I've used a totally different method of constructing the endpoints, let me know if this new approach makes more sense!

AetherUnbound:
I tested this locally as a dated DAG and it worked great, I also set `dated=False` and ran again and the iteration behaved as expected!
```
[2022-10-07, 20:16:08 UTC] {provider_data_ingester.py:151} INFO - Begin ingestion for PhylopicDataIngester
[2022-10-07, 20:16:08 UTC] {phylopic.py:76} INFO - Constructed endpoint: http://phylopic.org/api/a/image/list/0/25
[2022-10-07, 20:16:21 UTC] {provider_data_ingester.py:165} INFO - 25 records ingested so far.
[2022-10-07, 20:16:21 UTC] {phylopic.py:76} INFO - Constructed endpoint: http://phylopic.org/api/a/image/list/25/25
[2022-10-07, 20:16:34 UTC] {provider_data_ingester.py:165} INFO - 50 records ingested so far.
[2022-10-07, 20:16:34 UTC] {phylopic.py:76} INFO - Constructed endpoint: http://phylopic.org/api/a/image/list/50/25
[2022-10-07, 20:16:48 UTC] {provider_data_ingester.py:165} INFO - 74 records ingested so far.
[2022-10-07, 20:16:48 UTC] {phylopic.py:76} INFO - Constructed endpoint: http://phylopic.org/api/a/image/list/75/25
[2022-10-07, 20:17:01 UTC] {provider_data_ingester.py:165} INFO - 99 records ingested so far.
[2022-10-07, 20:17:01 UTC] {phylopic.py:76} INFO - Constructed endpoint: http://phylopic.org/api/a/image/list/100/25
[2022-10-07, 20:17:02 UTC] {media.py:219} INFO - Writing 100 lines from buffer to disk.
[2022-10-07, 20:17:14 UTC] {provider_data_ingester.py:165} INFO - 124 records ingested so far.
[2022-10-07, 20:17:14 UTC] {phylopic.py:76} INFO - Constructed endpoint: http://phylopic.org/api/a/image/list/125/25
[2022-10-07, 20:17:27 UTC] {provider_data_ingester.py:165} INFO - 149 records ingested so far.
[2022-10-07, 20:17:27 UTC] {phylopic.py:76} INFO - Constructed endpoint: http://phylopic.org/api/a/image/list/150/25
[2022-10-07, 20:17:40 UTC] {provider_data_ingester.py:165} INFO - 171 records ingested so far.
[2022-10-07, 20:17:40 UTC] {phylopic.py:76} INFO - Constructed endpoint: http://phylopic.org/api/a/image/list/175/25
[2022-10-07, 20:17:53 UTC] {provider_data_ingester.py:165} INFO - 196 records ingested so far.
[2022-10-07, 20:17:53 UTC] {phylopic.py:76} INFO - Constructed endpoint: http://phylopic.org/api/a/image/list/200/25
```

AetherUnbound:
@stacimc It's not clear to me why this was set up to run in both manners, it seems like that's how it was from the beginning: https://github.com/WordPress/openverse-catalog/commit/1518383943e806e95b0727e815efc0d2af86906c

I chose `@weekly` for this DAG rather than `@daily` because there were many days which had no updates so the runs would skip. It seemed like running this once a week rather than daily meant we'd have a smaller impact on the API while ensuring more of the times we actually ran picked up data.

I think once we get a reingestion workflow set up for this provider, we can do away with the reprocessing altogether! That, OR, our "reingestion" workflow can just reconsume the entire dataset using the offset + limit approach 🤔 🤷🏼‍♀️

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -13,272 +13,263 @@
...
+        data = self._get_response_data(response_json)
+
+        if not data:
+            logger.warning("No content available!")
+            return None
+
+        return data
+
+    @staticmethod
+    def _image_url(_uuid: str) -> str:
```
sarayourfriend, line 121: Is `uuid` reserved? Why does it have a leading `_`? I don't _see_ it imported in this module, but maybe it's being defined globally somehow? Is this an Airflow specific thing?

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -13,272 +13,263 @@
...
+                    img_url = f"{PhylopicDataIngester.host}{img_url}"
+                    width = image.get("width")
+                    height = image.get("height")
+
+        return img_url, width, height
+
+    @staticmethod
+    def _get_taxa_details(result: dict) -> tuple[list[str] | None, str]:
+        taxa = result.get("taxa", [])
+        # [0].get('canonicalName', {}).get('string')
```
sarayourfriend, line 155: Is this comment leftover? If it is still relevant, can you add some explanation for it, please? It is not clear what it's meant to indicate at the moment.

It looks like from the code below it was maybe scratch-pad type stuff, so adding a suggestion to just remove it in case it is no longer relevant.

```suggestion
```

sarayourfriend (commented):
Just a couple comments but the endpoint handling makes way more sense, nice work :cowboy_hat_face:

openverse_catalog/dags/common/loader/provider_details.py
```diff
@@ -33,6 +33,7 @@
 WORDPRESS_DEFAULT_PROVIDER = "wordpress"
 FREESOUND_DEFAULT_PROVIDER = "freesound"
 INATURALIST_DEFAULT_PROVIDER = "inaturalist"
+PHYLOPIC_DEFAULT_PROVIDER = "phylopic"
```
stacimc, line 36: Funny this one wasn't set up here!

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -13,272 +13,263 @@
...
+    base_endpoint = f"{host}/api/a/image"
+    providers = {"image": prov.PHYLOPIC_DEFAULT_PROVIDER}
+    batch_limit = 25
+    # If a date is supplied to the ingester in the form of YYYY-MM-DD, this is the
+    # number of days from that date to process. By default,this DAG runs weekly
+    # and so default_process_days is 7.
+    default_process_days = 7
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
stacimc, line 39: Can we make Phylopic accept `process_days` as an argument here instead of overriding the `default` when we want to change it (as is done in the `main` method)?

This will be necessary for the reingestion workflows, where we want to be able to run Phylopic for individual days instead of a week at a time. Even with that change, this still introduces a complication for reingestion since those workflows are factory-generated. We would need to add an `args` option to the ProviderReingestionWorkflow configuration objects or something similar, that we could pass in through the factory 😬 

Outside of the scope of this PR, I'd like to at least consider making this a daily DAG. If it is giving us a small enough amount of data that this is unreasonable, then alternatively maybe it shouldn't be dated?

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -13,272 +13,263 @@
...
+                ]
+            )
+        }
+        endpoint = f"{self.base_endpoint}/{_uuid}"
+        response_json = self.get_response_json(params, endpoint)
+        result = self._get_response_data(response_json)
+        if not result:
+            return None
+
+        return self._get_meta_data(result)
```
stacimc, line 248: Total nit, but `_get_meta_data` is slightly confusing because `meta_data` is itself just one of the fields being extracted.

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -13,272 +13,263 @@
...
+            "meta_data": meta_data,
+        }
+
+    def get_record_data(self, data: dict) -> dict | list[dict] | None:
+        _uuid = data.get("uid")
+        if not _uuid:
+            return
+        logger.debug(f"Processing UUID: {_uuid}")
+        params = {
+            "options": " ".join(
```
stacimc, line 227: I really like this improvement for readability 😄

stacimc (approved):
Nice one, I think this might be the weirdest one to refactor! I ran this locally while reviewing the code and forgot that `catchup` was turned on, so it ended up running quite a backfill with no problems 😆

I made #813 to track pulling out the logic for the 'non-dated' version. I have some reservations about the unusual setup with running `@weekly`, and how we will tweak this for reingestion -- but either way I think it's out of scope here. This looks good to me!

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
stacimc: That _is_ strange. Looks like this comes from the original code and also applied to some other variables like `_endpoint`, which definitely shouldn't be reserved. Maybe meant to denote an internal variable, but it's not done consistently 🤔

I don't see any reason we couldn't omit the leading `_`, we have variables named `uuid` elsewhere in provider scripts. Probably a leftover artifact of refactoring multiple times!

AetherUnbound: Oop, definitely holdover from the original code. I think it's more convention than anything, since `uuid` _is_ a module and it might be ideal to make the distinction between what this is and that module. That said, we aren't using that module here as y'all point out so I'll change it 🙂

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
AetherUnbound: Ah, good eye!

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
AetherUnbound: I tried doing this initially, and I ran into some issue with handling `args` vs `kwargs` vs this parameter in the various places that the provider ingester class is passed around. I think having it as `@weekly` was fine initially, but if that's going to create more problems for the reingestion workflow, then maybe this would just be better as `@daily` 😄 It'd prevent us from having to figure out all these additional workarounds for `process_days` as well. I'll go ahead and make that change as a commit, but if we do want to pursue just keeping this as a weekly DAG before merging I can revert that commit.

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
AetherUnbound: I've removed this function entirely, it wasn't really serving a useful extra distinction and it made the class method calls easier!

AetherUnbound:
I've updated this to be an `@daily` DAG and removed all the logic for processing more than one day at a time. I've also tested this locally to ensure that the `catchup=True` nature of this DAG won't result in it trying to run daily versions for all of the days which are _not_ Sunday (i.e. when the `@weekly` runs would kick off). 

Here are some local runs, you can see the logical date is eveyr 7 days until the 21st, where it starts running daily after that!

image

stacimc:
Looks great to me! Switching to `@daily` seems like the easiest path forward and will definitely make reingestion easier 👍

> I've also tested this locally to ensure that the catchup=True nature of this DAG won't result in it trying to run daily versions for all of the days which are not Sunday (i.e. when the @weekly runs would kick off).

Kudos for thinking to test this, what a good catch!

openverse_catalog/dags/providers/provider_workflows.py
```diff
@@ -178,8 +179,9 @@ def __post_init__(self):
     ),
     ProviderWorkflow(
         provider_script="phylopic",
+        ingestion_callable=PhylopicDataIngester,
```
obulat, line 182: Do we know what's the actual start_date for Phylopic images? I've run this for several days in 2011 and got no return. Does it make sense to update it closer to the actual start date for Phylopic to prevent unnecessary requests?

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -13,272 +13,238 @@
...
+
+    @staticmethod
+    def _get_creator_details(result: dict) -> tuple[str | None, str | None, str | None]:
+        credit_line = None
+        pub_date = None
+        creator = None
+        submitter = result.get("submitter", {})
+        first_name = submitter.get("firstName")
+        last_name = submitter.get("lastName")
+        if first_name and last_name:
```
obulat, line 166: I think it would be nice to add the submitter if there is a first_name OR last_name. Does that ever happen?

obulat (approved):
It makes sense to standardize all the DAGs to be daily when the date is provided.
I've added a couple of nitty comments inline. Other than that, everything works and looks great!

openverse_catalog/dags/providers/provider_workflows.py
AetherUnbound: When this was weekly, it would only take a few runs to actually pull data (see the durations [in this screenshot above](https://github.com/WordPress/openverse-catalog/pull/747#issuecomment-1287482934) as a proxy for when data starts coming in). We could bump it up to that week I suppose!

PR #792, opened 2022-10-12
Add condition to not notify slack when don't have files in dags folder

### LINKED ISSUE #785: Only report a deployment when DAG files are affected
## Description

Presently we get deployment messages whenever the repo is updated (see the [`dag-sync.sh` file](https://github.com/WordPress/openverse-catalog/blob/28bfd1696fb362a5c7a8c6c46d603354920b6efa/dag-sync.sh) and [deployment documentation](https://github.com/WordPress/openverse-catalog/blob/28bfd1696fb362a5c7a8c6c46d603354920b6efa/DEPLOYMENT.md).

However, there are often cases where PRs that affect only dependencies, tests, or documentation are merged. These cases have no outstanding effect on the operational code which Airflow uses, and thus reporting their deployment is not as useful (and regarding dependencies, actually inaccurate since the dependency update won't be deployed until a release is cut and the whole service is redeployed).

We should modify the DAG sync script so it only issues a slack message if files within the `dags/` folder is modified, otherwise the slack message step should be skipped.

## Implementation

ilitotor:
## Fixes

Fixes #785 785 by @AetherUnbound 

## Description
This PR gets the last git log and checks if have files inside the `dags` folder to notify slack. 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/dag-sync.sh
+++ b/dag-sync.sh
@@ -22,9 +22,9 @@ new=$(git rev-list --reverse --topo-order HEAD..origin/main | head -1)
 [ -z "$new" ] && exit
 
 # Verify if have /dags/ in the last commit
-have_dag=$(git log -p -1 origin/main  --pretty=format: --name-only | grep "/dags/")
+have_dag=$(git log -p -1 $new  --pretty=format: --name-only | grep "/dags/")
 # If there is no files under /dags/ folder, no need to notify, quit early
-![ "$have_dag" ] && exit
+[ -z "$have_dag" ] && exit
 
 #Move ahead to this new commit
 git reset --hard "$new"
--- a/dag-sync.sh
+++ b/dag-sync.sh
@@ -22,7 +22,7 @@ new=$(git rev-list --reverse --topo-order HEAD..origin/main | head -1)
 [ -z "$new" ] && exit
 
 # Verify if have /dags/ in the last commit
-have_dag=$(git log -p -1 $new  --pretty=format: --name-only | grep "/dags/")
+have_dag=$(git log -p -1 "$new"  --pretty=format: --name-only | grep "/dags/")
 # If there is no files under /dags/ folder, no need to notify, quit early
 [ -z "$have_dag" ] && exit
 
--- a/dag-sync.sh
+++ b/dag-sync.sh
@@ -21,16 +21,15 @@ new=$(git rev-list --reverse --topo-order HEAD..origin/main | head -1)
 # If there is no new commit hash to move to, nothing has changed, quit early
 [ -z "$new" ] && exit
 
+# Move ahead to this new commit
+git reset --hard "$new"
 # Verify if have /dags/ in the last commit
 have_dag=$(git log -p -1 "$new"  --pretty=format: --name-only | grep "/dags/")
 # If there is no files under /dags/ folder, no need to notify, quit early
 [ -z "$have_dag" ] && exit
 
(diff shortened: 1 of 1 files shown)


dag-sync.sh
```diff
@@ -20,7 +20,13 @@ git fetch origin
 new=$(git rev-list --reverse --topo-order HEAD..origin/main | head -1)
 # If there is no new commit hash to move to, nothing has changed, quit early
 [ -z "$new" ] && exit
-# Move ahead to this new commit
+
+# Verify if have /dags/ in the last commit
+have_dag=$(git log -p -1 origin/main  --pretty=format: --name-only | grep "/dags/")
+# If there is no files under /dags/ folder, no need to notify, quit early
+![ "$have_dag" ] && exit
```
AetherUnbound, line 27: This may not work as expected if several PRs are merged in quick succession - `origin/main` will be several commits ahead, but the next commit being merged (defined on line 20) will only be one commit ahead. In cases where the next commit doesn't have a DAGs change but the tip of `origin/main` does, this would incorrectly report the next commit (and all commits unconditionally until the codebase had caught up with `origin/main`). Similarly, if the tip of `origin/main` doesn't have DAG changes, no commits along the way would be reported even if they did have DAG changes. I tested this locally, and I think what we'll want is:
```suggestion
# Verify if have /dags/ in the last commit
have_dag=$(git log -p -1 $new  --pretty=format: --name-only | grep "/dags/")
# If there is no files under /dags/ folder, no need to notify, quit early
[ -z "$have_dag" ] && exit
```

AetherUnbound (changes requested):
Thank you for submitting this PR @ilitotor! This is a great approach. It looks like the linting is failing, I think the proposed suggestion below should address the issue 😄

dag-sync.sh
ilitotor: Thanks, @AetherUnbound! It makes total sense.

dag-sync.sh
```diff
@@ -20,7 +20,13 @@ git fetch origin
 new=$(git rev-list --reverse --topo-order HEAD..origin/main | head -1)
 # If there is no new commit hash to move to, nothing has changed, quit early
 [ -z "$new" ] && exit
-# Move ahead to this new commit
+
+# Verify if have /dags/ in the last commit
+have_dag=$(git log -p -1 "$new"  --pretty=format: --name-only | grep "/dags/")
+# If there is no files under /dags/ folder, no need to notify, quit early
+[ -z "$have_dag" ] && exit
```
AetherUnbound, line 27: One more change I just realized (sorry!), this should go _after_ the `git reset --hard "$new"` line below. If we don't move forward on each commit, `$new` will always be the next commit ahead and we will run into the same loop of logic for skipping it if it happens to not include DAGs. We only want to exit prior to sending the Slack notification if there aren't any DAGs.

AetherUnbound (changes requested):
The new changes look great, but apologies, one more thing before this is ready!

krysal (approved):
Looks good. Thank you for this valuable contribution to the Openverse's catalog @ilitotor!

AetherUnbound (approved):
Looks great, congrats on your first contribution! 😄

PR #806, opened 2022-10-18
Allow default retry count to be determined by environment variable

### LINKED ISSUE #675: Allow default retry count to be determined by environment variable
## Description

Most of our DAGs share default arguments, which can be found here: 

https://github.com/WordPress/openverse-catalog/blob/492ae8b521dd2cf7356b1295c68c72feaf7d491c/openverse_catalog/dags/common/constants.py#L14-L23

By default (and unless explicitly stated at the DAG or task level), DAGs which use these arguments have their retries set to `2`, which means that the task will run 3 total times before it eventually fails.

For local development, these retries can be cumbersome and time-consuming. It would be great if we could have that value configured from an environment variable which defaults to `2` if not set.

## Implementation

Pmeet:
Modified the DAG_DEFAULT_ARGS dictionary's retries record and added the DEFAULT_RETRY_COUNT to 2 in env.template file and reformatted both files according to flake8

## Fixes

Fixes #675  by @AetherUnbound 

## Description

I fixed this issue by adding the environment variable for the retries in env.template and fetching the same variable in constants.py file using os.getenv(<variable-name>). 

## Testing Instructions

Not Applicable

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/constants.py
+++ b/openverse_catalog/dags/common/constants.py
@@ -22,9 +22,7 @@ DAG_DEFAULT_ARGS = {
     "email_on_failure": False,
     "email_on_retry": False,
     "email": [CONTACT_EMAIL],
-    "retries": os.getenv(
-        "DEFAULT_RETRY_COUNT"
-    ),  # Default retries is set to 2 in .env file,
+    "retries": int(os.getenv("DEFAULT_RETRY_COUNT", 2)),
     "retry_delay": timedelta(minutes=5),
     "execution_timeout": timedelta(hours=1),
     "on_failure_callback": slack.on_failure_callback,
--- a/openverse_catalog/dags/common/constants.py
+++ b/openverse_catalog/dags/common/constants.py
@@ -22,7 +22,9 @@ DAG_DEFAULT_ARGS = {
     "email_on_failure": False,
     "email_on_retry": False,
     "email": [CONTACT_EMAIL],
-    "retries": 2,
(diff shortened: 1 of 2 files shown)


openverse_catalog/dags/common/constants.py
```diff
@@ -22,7 +22,9 @@
     "email_on_failure": False,
     "email_on_retry": False,
     "email": [CONTACT_EMAIL],
-    "retries": 2,
+    "retries": os.getenv(
+        "DEFAULT_RETRY_COUNT"
+    ),  # Default retries is set to 2 in .env file,
```
AetherUnbound, line 27: In the case where this environment variable isn't set, we'd still like this to default to 2. We can change change that by doing:
```suggestion
    "retries": os.getenv("DEFAULT_RETRY_COUNT", 2),
```

AetherUnbound (changes requested):
Thanks for submitting this PR @Pmeet! I have one suggested change (in some environments we don't specify all env vars or use the `.env` file at all, so we want to make sure we have that case covered).

AetherUnbound (approved):
This looks great, thanks @Pmeet and congrats on your first contribution! 😄

stacimc (approved):
Looks great 😄 Thanks for the contribution!

Issue #26, opened 2021-04-21
[Bug] Flickr Provider API Script doesn't retrieve all unique items for a given day (original #326)


obulat:
This issue has been migrated from the [CC Search Frontend repository](https://github.com/creativecommons/cccatalog/issues/326)

```
Author: mathemancer
Date: Fri Mar 13 2020
Labels: Hacktoberfest,help wanted,🙅 status: discontinued,🛠 goal: fix
```
## Bug Description

When the Flickr Provider API Script is run for a given day, the output is a copy of the rows for the same ~4,500 items over and over.  Instead, it should collect around 80,000-90,000 unique items per day.

## To Reproduce

Run `python flickr.py --date 2018-08-20`, notice that the output is 90,000 lines long, but there are only 4,376 unique photos (repeated over and over).

## Expected behavior

There should be *far* fewer repeated photos in the output, and *far* more unique results per day.

## Additional context

This appears to be due to some bug in the API.  Basically, it's actually *returning* the same ~4,500 articles over and over when called for an entire day's worth of data and paging through the result.

I tried splitting the day into hours, half hours, and 15 minute blocks, to see what the results would be (i.e., I looped through the same day's data by hours, half hours, etc.).  The results were confusing, to say the least.  All these results are from `--date 2018-08-20`:
```text
|  split  |  rows   | unique |
+---------+---------+--------+
| day     |  90,000 |  4,376 |
+---------+---------+--------+
| hour    | 106,457 | 82,008 |
+---------+---------+--------+
| 30 min  | 115,723 | 84,326 |
+---------+---------+--------+
| 15 min  | 140,966 | 82,484 |
+---------+---------+--------+
```
It seems like splitting things up by some time interval is a good idea, but it's not clear what the appropriate time interval might be.

I also tried separating out the licenses (currently we ask for *all* CC licensed items in the same query), and looping through 15-minute intervals.  This resulted in  

129,435 rows, with 78,210 unique results.

--------
### Original Comments: 

*mathemancer* commented on Fri Mar 13 2020:

>We need to understand why this is happening in the first place before actual development is appropriate, but I'd welcome ideas from the community, or more experimental results.
[source](https://github.com/creativecommons/cccatalog/issues/326#issuecomment-598729510)

*mathemancer* commented on Fri Mar 13 2020:

>Further info:  If you re-run the job with one of the splits above, you get the *exact same* result, so at least it's deterministic, if not that easy to figure out.
[...]

PR #809, opened 2022-10-18
Refactor Flickr to use ProviderDataIngester

### LINKED ISSUE #585: Refactor Flickr to use ProviderDataIngester
## Suggested Improvement

Refactor the Flickr provider script to use the new ProviderDataIngester base class

## Benefit

More details in #229

## Implementation

stacimc:
## Fixes

Fixes #585 by @stacimc 

## Description

Refactors Flickr to use the ProviderDataIngester base class! Also hooks up the Flickr reingestion workflow with the new ingester.

There is one significant change made to the provider script in this refactor, which is the temporary removal of filesize information. Previously, we were making an additional request for each record to obtain this info. When I tested with this in place, it took ~2 minutes to get 100 records, versus ~4seconds for 100 records without it. This is not tenable for a provider that ingests so much data, especially for the reingestion workflow. I've opened #811 to track finding a better way to do this.

The Flickr provider script also has a few pre-existing issues, which are not resolved in this simple refactor. This is tracked in #26. Briefly: we're getting enormous numbers (100ks) of duplicate records from Flickr and possibly missing some unique ones, due to strange behavior from the Flickr API. I did some investigation but ultimately tabled it to work on as part of the Stability milestone.

Unfortunately this means that I cannot recommend turning on either Flickr or its reingestion workflow until after at least that issue is addressed.

## Testing Instructions

`just test`

Try running the workflow and reingestion workflow locally. I strongly recommend setting an `ingestion_limit` Airflow variable!

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida:
@stacimc the frontend does infer the image width,height, and even filetype from the image file itself, after loading the image on the frontend. IT tries to read the API and then falls back to the values read in 'realtime' from the image.

Here's the relevant code: 

https://github.com/WordPress/openverse-frontend/blob/6a70b968fe3679db2898270cb0161574b255b0c6/src/pages/image/_id.vue#L186-L187

stacimc:
I tested the full reingestion workflow, with an `ingestion_limit` of 100 records set. It worked well and took **13min 43 sec** with that ingestion limit. However I can't recommend turning on the reingestion workflow until we resolve #26.

#26 tracks a weird bug with the Flickr API where the API appears to be giving us a huge number of duplicates -- and strangely, this is exacerbated by requesting data for a larger period of time. (Meaning the problem is worse if you request photos for an entire day, versus splitting the data into hour chunks and getting the photos for each hour).

Currently (not an addition in this PR), the Flickr script splits the ingestion day into 48 half-hour chunks. Unfortunately I still saw this buggy behavior locally when I ran the provider script. I tried running it for about 25 minutes and observed:
* The script made it through 15 of these half-hour chunks in that time (7.5 hours worth of data)
* The first 14 half-hour chunks processed very quickly, and the bulk of processing time was spent on the last chunk.
* 649,342 records were ingested
* **634,295** of these were duplicates. Only ~15k unique records were ingested

I'm going to try increasing the number of chunks. It's currently unclear to me whether we can turn Flickr on until this has been resolved 👀

obulat (commented):
I was trying to run it by running the script separately, without Airflow (`python .../flickr.py`), and I think there's something wrong with passing of the arguments to the base class `__init__`:
We call `main` with the only argument,`date`, as a string. It becomes a part of the `args` and is passed as such to the base class. However, the base class expects the first parameter to be a `conf` dictionary:
https://github.com/WordPress/openverse-catalog/blob/065b22b9d194def4bf517a29c75a8b06d90dc171/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py#L84

I've seen the same thing in some other scripts as well (Europeana refactoring and others).

openverse_catalog/dags/providers/provider_api_scripts/flickr.py
```diff
@@ -59,345 +36,258 @@
...
+        # Build a list of start/end timestamps demarcating portions of the day
+        # for which to ingest data.
+        timestamp_pairs = self._derive_timestamp_pair_list(self.date)
+
+        for start_ts, end_ts in timestamp_pairs:
+            logger.info(f"Ingesting data for start: {start_ts}, end: {end_ts}")
+            super().ingest_records(start_timestamp=start_ts, end_timestamp=end_ts)
+
+    def get_next_query_params(self, prev_query_params, **kwargs):
+        start_timestamp = kwargs.get("start_timestamp")
```
obulat, line 100: It seems like these two lines can be moved inside the `if` clause, to line 104.

openverse_catalog/dags/providers/provider_api_scripts/flickr.py
```diff
@@ -59,345 +36,258 @@
...
-        1 for screenshots
-        3 for other
-    Treating everything different from photos as unknown.
-    """
-    if "content_type" in image_data and image_data["content_type"] == "0":
-        return ImageCategory.PHOTOGRAPH.value
-    return None
+def main(date):
+    ingester = FlickrDataIngester()
+    ingester.ingest_records(date=date)
```
obulat, line 290:
```suggestion
def main(date):
    ingester = FlickrDataIngester(date=date)
    ingester.ingest_records()
```

obulat: I think I've figured this out. We need to pass the date to the ingester itself, not its `ingest_records` method, right? At least, this way it works for me locally when I run the script using `python flickr.py`

openverse_catalog/dags/providers/provider_api_scripts/flickr.py
```diff
@@ -59,345 +36,258 @@
...
 
+    def _url_join(self, *args):
+        return "/".join([s.strip("/") for s in args])
+
+    @staticmethod
+    def _get_largest_image_size(image_data):
+        """
+        Returns the key for the largest image size available.
+        """
+        for size in ["l", "m", "s"]:
```
AetherUnbound, line 202: We're also including `url_t` in the query params, should `t` (I'm assuming this means "tiny") be another option here as well?

openverse_catalog/dags/providers/provider_api_scripts/flickr.py
```diff
@@ -59,345 +36,258 @@
...
+        license_, license_version = LICENSE_INFO.get(license_id)
+        return get_license_info(license_=license_, license_version=license_version)
+
+    def _get_file_properties(self, image_url):
+        """
+        Get the size of the image in bytes and its filetype.
+        """
+        filesize, filetype = None, None
+        if image_url:
+            filetype = image_url.split(".")[-1]
```
AetherUnbound, line 225: If this is just coming from the image URL, should we just remove/comment out this whole function for now since the media store will get the filetype as well?

openverse_catalog/dags/providers/provider_workflows.py
```diff
@@ -136,10 +137,16 @@ def __post_init__(self):
     ),
     ProviderWorkflow(
         provider_script="flickr",
+        ingestion_callable=FlickrDataIngester,
         start_date=datetime(2004, 2, 1),
         schedule_string="@daily",
         dated=True,
     ),
+    ProviderWorkflow(
```
AetherUnbound, line 145: Oop, thanks for alphabetizing these 😄

AetherUnbound (commented):
This is looking good! Though it's disappointing we won't be able to turn it on until we get the duplicates issue resolved 😞

openverse_catalog/dags/providers/provider_api_scripts/flickr.py
stacimc: Oh wow, good catch! And makes me realize I haven't actually been testing running the scripts via the CLI. Thank you for being more thorough!

openverse_catalog/dags/providers/provider_api_scripts/flickr.py
stacimc: I'm not positive why we wouldn't, but we did not accept "t" in the original implementation and we also had a test that asserted that we get no `image_url` when only `url_t` is provided (using this resource: [image_data_no_image_url.json](https://github.com/WordPress/openverse-catalog/blob/main/tests/dags/providers/provider_api_scripts/resources/flickr/image_data_no_image_url.json)), so I'm at least cautious about changing it. Maybe in a separate issue?

openverse_catalog/dags/providers/provider_api_scripts/flickr.py
stacimc: Good point! Yeah, let's just get rid of this entirely.

openverse_catalog/dags/providers/provider_api_scripts/flickr.py
```diff
@@ -59,345 +36,258 @@
...
+                "api_key": self.api_key,
+                "license": self.license_param,
+                "per_page": self.batch_limit,
+                "method": "flickr.photos.search",
+                "media": "photos",
+                "safe_search": 1,  # Restrict to 'safe'
+                "extras": (
+                    "description,license,date_upload,date_taken,owner_name,tags,o_dims,"
+                    "url_t,url_s,url_m,url_l,views,content_type"
+                ),
```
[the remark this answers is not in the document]
stacimc: _Much_ better 👍

openverse_catalog/dags/providers/provider_api_scripts/flickr.py
AetherUnbound: That works! And good find 🧐

AetherUnbound (approved):
Looks great! I tested this locally and I was able to get it to ingest a few days worth of data.

Interestingly, we have thumbnails for some Flickr records in production and it also looks like we _could_ collect thumbnail URLs going forward (I replaced the `_b` suffix in a few URLs with `_m` and got a much smaller image, I'm assuming those are "big" and "medium" respectively, especially because `_s` gives a very small image!). That said, we haven't decided yet on how to collect those so I don't think it's worth trying to add that as part of this PR 🙂

obulat (approved):
I tested the refactored version locally, and it ingests data well. The requests for image dimensions were taking so much time! The dimensions are very important for the frontend, but I think we can handle them in a separate process, together with other similar providers.

stacimc:
Rebased to fix merge conflicts

PR #818, opened 2022-10-21
Respect ingestion limit in process_batch

stacimc:
## Description

I think this should fix the final big edge case with ingestion limit :)

Before this PR, when an INGESTION_LIMIT was set in a provider script, we would cap the `batch_limit` to that value, and then check if the limit was exceeded _after each call to `process_batch`_. This leaves an edge case when a provider doesn't pass `batch_limit` to their API, and may get back huge batches that are larger than our limit.

This PR adds one extra check for the `ingestion_limit`, inside `process_batch` itself. Now we will halt processing even within a batch when the limit is reached.

## Testing Instructions

Metropolitan is a good example of the buggy behavior. You can observe the bug in action by setting an `INGESTION_LIMIT` Airflow variable to `100` locally and running `metropolitan_workflow`. You'll see that it appears to ignore the limit and continues processing after 100.

Then checkout this branch and try again. You could also try verifying that the limit is still respected on other providers like Cleveland.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
@@ -351,7 +351,7 @@ class ProviderDataIngester(ABC):
         Returns the total count of records ingested up to this point, for all
         media types.
         """
-        record_count = 0
+        processed_count = 0
 
         for data in media_batch:
             record_data = self.get_record_data(data)
@@ -375,9 +375,13 @@ class ProviderDataIngester(ABC):
                 # Add the record to the correct store
                 store = self.media_stores[media_type]
                 store.add_item(**record)
-                record_count += 1
+                processed_count += 1
 
-        return record_count
+                if self.limit and (self.record_count + processed_count) >= self.limit:
+                    logger.info("Ingestion limit has been reached. Halting processing.")
+                    return processed_count
+
+        return processed_count
 
     @abstractmethod
     def get_media_type(self, record: dict) -> str:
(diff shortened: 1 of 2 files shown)


AetherUnbound (approved):
Nice catch! Do you think we'd be able to remove any of the higher level checks if we have this one at the lower level? 😮

stacimc:
> Do you think we'd be able to remove any of the higher level checks if we have this one at the lower level? 😮

Weirdly, I don't think so! Or rather, we _could_, but while it would stop the ingester from _committing_ any new records past the limit, it wouldn't stop it from trying to fetch additional batches (it would keep fetching and return early each time). The current checks are all needed to make sure ingestion totally halts immediately.

There's two other checks:
* [[1]](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py#L158) At the beginning of `ingest_records`. We need this one to stop the ingester from attempting to start ingestion again in providers that override `ingest_records` to be called multiple times (like Finnish).
* [[2]](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py#L206) At the end of `ingest_records` -- I thought we might be able to get rid of this one, but we need it to set `should_continue` to False and stop the loop.

It is surprisingly complicated to implement in a way that suits all our providers!

krysal (approved):
Works great! Also, thanks for the context, I was feeling a dejavú like, some work in this area has been done before.

stacimc:
>  I was feeling a dejavú like, some work in this area has been done before.

You and me both! 😄

PR #790, opened 2022-10-10
Add docs and template ProviderDataIngester

stacimc:
## Fixes

Fixes #757 by @stacimc

## Description

* Adds a `create_provider_ingester` script that can be used to generate a templated provider api script and test file. The generated files include a lot of documentation and TODOs aimed at helping a new contributor flesh out an ingestion class. Credit to an [earlier template script in this repository](https://github.com/WordPress/openverse-catalog/tree/1c9fd4cf46f7c90f8e9db6eaacb31d6ad031fcf5/openverse_catalog/templates), from which I lifted quite a bit of code!
* Adds the following docs (would love better title suggestions 😅):
  * `adding_a_new_provider.md`: Details a very brief overview of what it even means to add a provider to Openverse, explains how to use the script, and also a brief explanation of how to add a `ProviderWorkflow` to generate the actual DAG
  * `provider_data_ingester_faq.md`: Describes "advanced options" for implementing some common non-standard use cases in a ProviderDataIngester, framed as an FAQ
  * `data_models.md`: Extremely temporary documentation of our columns, meant to be replaced by #783.
* Expands some of the documentation in the actual `ProviderDataIngester` class

The intention is to avoid too much duplication of documentation from the _code_. If deemed necessary, we can look into auto-generating some from the doc strings, but my strategy was:
* Script makes it easy to generate a stubbed out new ingester class
* Comments and TODOs within the generated class lead the developer through implementing a "straightforward" provider
* "Advanced" options are documented separately with examples

There's definitely a balancing act of where to document things in the code vs the docfile vs the template. My goal is mostly for this to be a good starting point!

## Testing Instructions

* Read the docs , make sure links work, etc :)
* Test the new script
  * `just test`
    * Verify that the test files (`foobar_industries.py` & `test_foobar_industries.py`) were deleted after the tests finished
  * Run it with some sample data, ie: `just add-provider "Foo Museum" "https://foo-museum.org/api/v1/" image audio`
    * Verify the `foo_museum.py` provider script and `test_foo_museum.py` files were generated
    * Inspect the files to make sure they make sense and that the name/endpoint/media types were templated in properly
    * Try again with different media types, including a single media type
    

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [ ] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [ ] My commit messages follow [best practices][best_practices].
- [ ] My code follows the established code style of the repository.
[...]

Issue #702, opened 2022-08-26
Add configuration to skip specific ingestion errors


stacimc:
## Problem

#650 added dagrun conf options to make a provider script skip ingestion errors. This is currently all-or-nothing; when `skip_ingestion_errors` is `true`, _all_ errors (except for AirflowExceptions) are caught, allowing ingestion to continue, and then re-raised at the end of ingestion.

In reality, it's likely that we'll want to skip _specific_ errors, but we may still want ingestion to be halted if an unexpected error is thrown.

## Description

We can update the conf option to be similar to the `silenced_slack_alerts` configuration (#654), which allows matching an error predicate. Instead of a boolean value, `skip_ingestion_errors` could be a list of strings to match on.

Like `silenced_slack_alerts`, we should make sure this allows us to match specific error message text and error types (ie adding 'KeyError' to the list should allow us to skip all KeyErrors). It may be nice to make sure there's still a way to easily skip _all_ errors if that's truly what we want.

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

stacimc:
Resolved in #654

stacimc:
I mistakenly closed this thinking it was referring to skipping _Slack notifications_ for specific errors. This issue is for skipping the _errors themselves during ingestion_.

PR #821, opened 2022-10-24
Refactor Europeana to use `ProviderDataIngester`

### LINKED ISSUE #584: Refactor Europeana to use ProviderDataIngester
## Suggested Improvement

Refactor the Europeana provider script to use the new ProviderDataIngester base class

## Benefit

More details in #229

## Implementation

sarayourfriend:
## Fixes

Fixes #584 by @stacimc 

## Description

1. Refactors Europeana's provider script to use the ProviderDataIngester base class.
2. Adds a simple default implementation for `get_media_type` to the base class to remove boilerplate for providers that only have a single media type. This seemed like an easy improvement to me but there may be a reason why this isn't a good default implementation.

## Testing Instructions

Checkout the unit tests. Many of them had to be removed and new ones added because they were testing functionality that doesn't exist anymore (is abstracted away) or had to test new functionality. The existing tests that pertained to the data transformation I tried to keep as close as possible to the original tests. I caught some bugs in my refactor this way, which was very helpful! I did not add any new data transformation tests as the existing ones seemed comprehensive enough to me.

Finally, run the DAG. Follow the instructions in the README to run Airflow locally and then run the DAG. You can run it without an API key for testing. Currently this doesn't work though, at least not for me locally, due to an issue with the filename generation task. cc @stacimc for help debugging this. Here's the log output from the task:

```
[2022-10-24, 04:38:11 UTC] {taskinstance.py:1851} ERROR - Task failed with exception
  File "/usr/local/airflow/.local/lib/python3.10/site-packages/airflow/operators/python.py", line 175, in execute
    return_value = self.execute_callable()
  File "/usr/local/airflow/.local/lib/python3.10/site-packages/airflow/operators/python.py", line 193, in execute_callable
    return self.python_callable(*self.op_args, **self.op_kwargs)
  File "/usr/local/airflow/openverse_catalog/dags/providers/factory_utils.py", line 66, in generate_tsv_filenames
    stores = _load_provider_script_stores(ingestion_callable, media_types)
  File "/usr/local/airflow/openverse_catalog/dags/providers/factory_utils.py", line 37, in _load_provider_script_stores
    raise ValueError(
ValueError: Expected stores in providers.provider_api_scripts.europeana were missing: ['image']
```

I haven't dug into this yet but it seems like probably a configuration issue on my end.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


openverse_catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -8,266 +8,172 @@
...
-from common.storage.image import ImageStore
-from requests.exceptions import JSONDecodeError
+from providers.provider_api_scripts.provider_data_ingester import ProviderDataIngester
 
 
-logging.basicConfig(
-    format="%(asctime)s - %(name)s - %(levelname)s:  %(message)s", level=logging.INFO
-)
 logger = logging.getLogger(__name__)
+logging.getLogger(common.urls.__name__).setLevel(logging.WARNING)
```
obulat, line 23: Should we set the `urls` logging to WARNING in all provider scripts?

stacimc:
> Currently this doesn't work though, at least not for me locally, due to an issue with the filename generation task. cc @stacimc for help debugging this.

I think you need to update the `ProviderWorkflow` configuration for Europeana, [here](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_workflows.py#L125) by adding your new class as the `ingestion_callable`. That should take care of it!

Since Europeana is dated, we'll also need to add it to the `ProviderReingestionWorkflow` configuration [here](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_reingestion_workflows.py#L46). That should be all you need to make the reingestion workflow work as well, but if there's any issue that can always be a separate PR!

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -8,266 +8,172 @@
...
+        return ""
 
-        if next_cursor is None:
-            break
 
-        cursor = next_cursor
+class EuropeanaDataIngester(ProviderDataIngester):
+    providers = {"image": prov.EUROPEANA_DEFAULT_PROVIDER}
+    sub_providers = prov.EUROPEANA_SUB_PROVIDERS
+    batch_limit = 100
```
stacimc, line 106: Nit, `100` is the default batch_limit, so this can be removed.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -371,14 +371,19 @@ def process_batch(self, media_batch):
...
+        to returning the only media type defined in the ``providers`` attribute.
         """
-        pass
+        if len(self.providers) == 1:
+            return list(self.providers.keys())[0]
+
+        raise NotImplementedError(
+            "Provider scripts that support multiple media types "
+            "must provide an override for ``get_media_type``."
+        )
```
AetherUnbound, line 386: Great idea!

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -8,266 +8,172 @@
...
+    and simplify testing a bit.
+    """
+
+    def get_record_data(self, data: dict) -> dict:
+        record = {
+            "foreign_landing_url": self.get_foreign_landing_url(data),
+            "image_url": data.get("edmIsShownBy")[0],
+            "foreign_identifier": data.get("id"),
+            "meta_data": self.get_meta_data_dict(data),
+            "title": data.get("title")[0],
```
AetherUnbound, line 38: For both of these `get(<x>)[0]` calls, it's possible `get` could return `None`, which would cause the indexing into `None` to raise an error. Could we add defaults or extra checking steps for those two fields here?

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
AetherUnbound, line 106: I believe this is the default batch limit, so it shouldn't need to be overwritten here!

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -8,266 +8,172 @@
...
-        raise Exception(
-            f"More than one sub-provider identified for the "
-            f"image with foreign ID {foreign_id}"
-        )
-    source = (
-        eligible_sub_providers.pop() if len(eligible_sub_providers) == 1 else provider
-    )
+    def get_next_query_params(self, prev_query_params) -> dict:
+        if not prev_query_params:
+            return self.base_request_body
```
stacimc, line 151: From looking at the original code, do you not need `"cursor": "*"` in the initial request? It seems like it's working fine as is, though 🤔

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -8,266 +8,172 @@
...
-        license_info=license_info,
-        foreign_identifier=foreign_id,
-        title=title,
-        meta_data=meta_data,
-        source=source,
-    )
+    def get_should_continue(self, response_json):
+        if response_json.get("success") != "True":
+            logger.warning('Request failed with ``success = "False"``')
+            return False
```
AetherUnbound, line 160: I _think_ this logic might be better suited in `get_batch_data`, since `get_should_continue` is typically concerned with the token/cursor iteration rather than error handling. @stacimc what do you think?

AetherUnbound (commented):
This is great, thanks for your additional contributions to the base class & testing mocks! I have a few small comments. I think the big required changes are the ones Staci has mentioned 🙂

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -8,266 +8,172 @@
...
+class EuropeanaRecordBuilder:
+    """
+    A small class to contain the record building functionality
+    and simplify testing a bit.
+    """
+
+    def get_record_data(self, data: dict) -> dict:
+        record = {
+            "foreign_landing_url": self.get_foreign_landing_url(data),
+            "image_url": data.get("edmIsShownBy")[0],
```
stacimc, line 35: Is it possible for this (or `foreign_landing_url`, `foreign_id`, or `license_info`) to be `None`? We typically get the required fields first and return `None` early from `get_record_data` if they cannot be extracted.

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -8,266 +8,172 @@
...
+        return {k: v for k, v in meta_data.items() if v is not None}
+
+    def get_description(self, data: dict) -> str | None:
+        lang_aware_description = data.get("dcDescriptionLangAware")
+        if lang_aware_description:
+            description = lang_aware_description.get(
+                "en"
+            ) or lang_aware_description.get("def")
+        else:
+            description = data.get("dcDescription")
```
stacimc, line 95: This misses the case where `dcDescriptionLangAware` is available but does not have the `en` or `def` keys. In that case we still want to fall back to `dcDescription`.

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
stacimc, line 386: I _love_ this! 😄 Can you update the new `ProviderDataIngester` docs and template?

Out of scope, but should we make an issue to clean up our other provider scripts (and their tests)?

stacimc (changes requested):
This is fantastic 🥳 I like your handling of the cursor and the RecordBuilder. The default implementation of `get_media_type` is 💯 as well!

After updating the `ProviderWorkflow` configuration for Europeana, I was able to test this locally and compare it to the output of the legacy implementation. Everything looked good except for missing `description` in the metadata, which I mentioned in one of the comments. If you want to reproduce that, I was running the DAG for 2022-11-03.

This doesn't necessarily need to change in this PR but a note for testers: the [start date for Europeana](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_workflows.py#L128) is `2011-09-01` which means it starts backfilling from 2011 and does not get any data. (My fault, I set overly cautious start dates awhile back.) I adjusted this locally to a more reasonable `2020-11-01`, which starts the backfill from around the time the script was last ran.

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
sarayourfriend: I considered this when writing it, but if either `title` or `image_url` are missing for something, wouldn't we want it to fail out? Is there a different way to fail it out that works better with Airflow's expectations?

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
sarayourfriend: Are those fields allows to be None? :open_mouth:

openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
sarayourfriend: > Out of scope, but should we make an issue to clean up our other provider scripts (and their tests)?

Yes, in a separate issue, for sure! I can do that.

sarayourfriend: Done: https://github.com/WordPress/openverse-catalog/issues/822

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
sarayourfriend: Ah, good catch!

sarayourfriend: Fixed in https://github.com/WordPress/openverse-catalog/pull/821/commits/dfac506409087cc3a48d689ef49e7503bc65940b

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
sarayourfriend: What would you suggest should change here, specifically? Would `get_batch_data` just return an empty list in this case?

I don't personally see a big difference here... but I like that all the "is this response usable and can we move forward" is encapsulated in a single method rather than spread across two methods. But there may be a call order issue here, if "success" is False then we can't process the batch at all (in addition to not continuing, to maintain parity behaviour). :thinking:

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
AetherUnbound: I think we'd want to skip this record in that case, but not necessarily fail the entire ingestion run. A lot of our providers operate under the assumption that attributes missing from a particular record, even if necessary, would just be ignored and processing would continue. I feel like that might be the best approach here: log that something which was expected to be present was missing, then return `None` from the `get_record_data` function.

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
AetherUnbound: Yes, I would suggest moving L158 and L159 here to `get_batch_data`, and add a `return None` within the condition rather than `return False` (although `False` would work too because the check that follows is [falsy anyways](https://github.com/WordPress/openverse-catalog/blob/b13390fda17e14fc8b92559490a31eae80437d66/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py#L173)). Here are some examples from other providers:

* [Brooklyn Museum](https://github.com/WordPress/openverse-catalog/blob/9be8bcec541f95dcd38d76a1064ce72cd75f8255/openverse_catalog/dags/providers/provider_api_scripts/brooklyn_museum.py#L41-L48)
* [Finnish Museum](https://github.com/WordPress/openverse-catalog/blob/1882d7776a16c6d7ebc9d0a28c502c25777c6a40/openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py#L47-L56)

My interpretation of these methods is that if we try to get the batch and there's an error, there's functionally no batch to process (as you mention) and `get_batch_data` should return `None`. The `get_should_continue` mechanism is for more peculiar cases (like this one, or Wikimedia) where the API _additionally_ provides some token or other mechanism which tells us whether we should continue processing.

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
obulat: In other scripts, if a required value is None, we stop the current item's step early by using either `return None` or `continue`/`break` inside a loop. Here is an example:
https://github.com/WordPress/openverse-catalog/blob/1506d6d9d391d7d12c8f5c515737623218036928/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py#L46-L48

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -8,266 +8,173 @@
...
+            "source": (
+                eligible_sub_providers.pop()
+                if len(eligible_sub_providers) == 1
+                else EuropeanaDataIngester.providers["image"]
+            )
+        }
+
+    def _get_license_url(self, license_field) -> str | None:
+        if len(license_field) > 1:
+            logger.warning("More than one license field found")
```
obulat, line 66: Not sure if it should be done in this PR, but I think this log should be downgraded to `debug`

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
sarayourfriend: Right on, makes sense! I wonder if we'd want some kind of "incomplete record store" or something, but a discussion for the a different day :slightly_smiling_face: 

Do you happen to know the complete list of fields that are required for a particular record to be valid? Maybe this is something the base class could handle instead by verifying that all required fields are truthy (`not None` and `!= ""`, I suppose)?

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
sarayourfriend: > My interpretation of these methods is that if we try to get the batch and there's an error, there's functionally no batch to process (as you mention) and get_batch_data should return None

Interesting! This makes sense, but then my instinct is that `get_batch_data` should raise an error in these cases that is caught and handled differently. Returning `None` seems ambiguous to me (though this is just my first time working on a provider script, so take that with a grain of salt).

I'll make the changes you suggested, thanks, Madison.

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
sarayourfriend: Added back in 1ea1a507

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
sarayourfriend: I'd love a list of the required fields, maybe on the base class, to reference at the _end_ of the `get_record_data` method... or potentially as something that the base class itself checks. These kinds of checks seem like boilerplate and distributed, undocumented knowledge that the base class would be perfect for.

obulat: The list of required fields is in the `ImageStore.add_item` method's docs (the same is available in the AudioStore):

https://github.com/WordPress/openverse-catalog/blob/1506d6d9d391d7d12c8f5c515737623218036928/openverse_catalog/dags/common/storage/image.py#L67-L77

Do you think there's a better place for this list?

The checks are boilerplate, but not boilerplate _enough_ to extract them because all API responses are different, and the required pieces of information can be located in different stages of the script. The idea behind returning `None` is to fail early and skip the unnecessary work of extracting the metadata (which can be quite tedious for some API responses) if the item does not have the required fields and will be discarded anyways.

AetherUnbound: Staci also recently added documentation around this exact thing as part of #790, which can be found here: https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/docs/data_models.md

It's also a goal of the Data Normalization project (I believe) to improve on this understanding and ensure that we have documentation around it.

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
AetherUnbound: @obulat responded well to this in [another comment on this PR](https://github.com/WordPress/openverse-catalog/pull/821#discussion_r1004463445), and I added a note there too about recently added documentation 😄

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
AetherUnbound: For context on this, these scripts have historically operated in a manner that prioritizes gathering available records and preventing errors as much as possible. This was because (previously) an error condition meant that however x many hours of processing that had already occurred had to be done _again_ if we wanted to re-run a given day. We'd still get all the data we had gotten so far, but we hadn't had a mechanism until #650 (and with further improvement coming via #702) to pick up at a specific set of query params.

I think maintaining that resiliency within the scripts can be a good thing! However, particularly as we finally get everything back up and running we can start operating more under the principal of letting things break, because that tells us our assumptions about the API responses were wrong.

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
stacimc: I'll also say here that these fields being None shouldn't cause errors or anything -- I believe the MediaStore should successfully discard them (if I'm wrong and that's not the case I agree we should add this behavior in the Data normalization project). The point of the None checks is mostly to halt processing of a record _early_ if it's just going to be discarded anyway, as a general best practice.

We could update the base class to, instead of having just an abstract `get_record_data`, have abstract `get_required_record_data` and `get_optional_record_data` methods. Or even individual abstract methods for fetching each required field (`get_license_info`, `get_foreign_identifier`...). Then we could make `get_record_data` on the base class do the None checks for us.

I'll make an issue.

stacimc: #834

sarayourfriend:
@stacimc @AetherUnbound @obulat please re-review at your leisure. I've added code to short circuit record building when certain fields are empty.

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -8,266 +8,226 @@
...
+    """
+    A small class to contain the record building functionality
+    and simplify testing a bit.
+    """
 
-    while cursor is not None:
-        image_list, next_cursor, total_number_of_images = _get_image_list(
-            start_timestamp, end_timestamp, cursor
-        )
+    def get_record_data(self, data: dict) -> dict:
```
obulat, line 57:
```suggestion
    def get_record_data(self, data: dict) -> dict | None:
```

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -8,266 +8,226 @@
...
+    @raise_if_empty
+    def _get_image_url(self, data: dict) -> str | None:
+        group = data.get("edmIsShownBy")
+        return group[0] if group else None
+
+    @raise_if_empty
+    def _get_foreign_identifier(self, data: dict) -> str | None:
+        return data.get("id")
+
+    @raise_if_empty
```
obulat, line 102: A title is not required for a media item, some of them have `""` as the title. This might not be needed for Europeana, but is necessary for some providers who have user-provided items (such as Flickr).

obulat (approved):
I'm approving because I wouldn't want to block a working script, but I'm not sure I like the current implementation of short-circuiting.

I don't know if it's just because I'm more used to the old way, but it seems like the checks for None are much shorter than the decorated functions that replace them here. Overall, it seems more difficult to read than the old 'check for None and return early' way.

tests/dags/providers/provider_api_scripts/test_europeana.py
```diff
@@ -1,22 +1,18 @@
...
+from common.licenses import LicenseInfo, get_license_info
+from providers.provider_api_scripts.europeana import (
+    EuropeanaDataIngester,
+    EuropeanaRecordBuilder,
+)
 
 
 RESOURCES = os.path.join(
     os.path.abspath(os.path.dirname(__file__)), "resources/europeana"
 )
```
obulat, line 15:
```suggestion
RESOURCES = Path(__file__).parent / "resources/europeana"

```

tests/dags/providers/provider_api_scripts/test_europeana.py
```diff
@@ -1,22 +1,18 @@
 import json
-import logging
 import os
```
obulat, line 2:
```suggestion
from pathlib import Path
```

tests/dags/providers/provider_api_scripts/test_europeana.py
```diff
@@ -1,22 +1,18 @@
...
 )
 
-logging.basicConfig(
-    format="%(asctime)s - %(name)s - %(levelname)s:  %(message)s",
-    level=logging.DEBUG,
-)
-
 
 def _get_resource_json(json_name):
     with open(os.path.join(RESOURCES, json_name)) as f:
```
obulat, line 18:
```suggestion
    with open(RESOURCES / json_name) as f:
```

sarayourfriend:
> I don't know if it's just because I'm more used to the old way, but it seems like the checks for None are much shorter than the decorated functions that replace them here. Overall, it seems more difficult to read than the old 'check for None and return early' way.

That's fair. I guess I was trying to avoid doing this run around (just a psuedo-code kind of example):

```py
foreign_identifier = data.get("guid")
if not foreign_identifier:
    logger.warning(f"foreign_identifier was empty: {foreign_identifier}")
    return None

# ... etc
# then much later in the function

return {
    "foreign_identifier": foreign_identifier,
    # etc... repeat x3? x4?
}
```

The decorator approach feels like it gets rid of the boilerplate checks a bit and also automatically adds logging around it. It's definitely not as straightforward, so I can appreciate the sense you have about it being more complicated and harder to follow. As a pattern I feel like it might be an easier to manage approach, or at least one that ties into one of the suggestions @stacimc made of having the base class require methods like `get_foreign_identifier` and then allow the base class to check these things...

I could very well be over-engineering this though and am totally open to changing it if others are on the same page as you, Olga.

stacimc:
Regarding the decorators -- I like them personally, but I understand having some reservations especially about consistency among providers and the possibility that it might be more confusing to implement for a newcomer to the project. 

> As a pattern I feel like it might be an easier to manage approach, or at least one that ties into one of the suggestions @stacimc made of having the base class require methods like get_foreign_identifier and then allow the base class to check these things...

This is exactly what I'm thinking -- I would love to use this approach in the _base class_ as part of #834. If the pattern needs to be re-implemented in each individual provider as we're doing currently, then I favor recommending the more straightforward old approach.

I'm personally fine with merging this into Europeana for now, though. How about making a suggestion on #834 to extract the logic from Europeana when we work on that issue? I'll add some notes there.

openverse_catalog/dags/providers/provider_api_scripts/europeana.py
stacimc: Agreed -- I'm not _sure_ if it's absolutely necessary to remove this check for Europeana or if we can assume all records will have a title, but since this will halt ingestion entirely if it raises I'd prefer to remove this check.

stacimc (approved):
This works great! I'd prefer to remove the exception on empty titles but from looking at the API docs I don't believe it should be an issue for Europeana in the same way that it would for more user-generated content, so I don't think anything here should block merge. Really nice work, I love the additions to the base class 😍

qq
```diff
@@ -0,0 +1,1399 @@
+[33m1ea1a507[m[33m ([m[1;36mHEAD -> [m[1;32mrefactor/europeana-provider-base-class[m[33m)[m Add back default cursor from previous implementation
```
AetherUnbound, line 1: Should this file be here?

AetherUnbound (changes requested):
I think I'm with the general consensus here - I do appreciate the decorator approach, but doing so in just this provider could make it more confusing and less approachable for folks (especially since all of the other providers follow a different pattern). That said, I'd also be okay merging this as-is.

My only concern is there seems to be an output file that was erroneously committed!

qq
sarayourfriend: Definitely not! I think I was trying to use vim binding to exit something that wasn't responding to `q` :cry: Somehow I am notorious for committing useless files like this, thanks for catching it. It is removed in https://github.com/WordPress/openverse-catalog/pull/821/commits/f02a4741814ede69b5349b0988241a88263364f1

AetherUnbound (approved):
🚀 Awesome work!!

PR #848, opened 2022-10-28
Fix Europeana success test, activate reingestion workflows

AetherUnbound:
## Description

This PR addresses a number of items with Europeana. I'm happy to split this out into multiple PRs if that would be most warranted, but all of the commits are focused on Europeana so this seemed easiest.

1. Fixed API responses
2. Updated the start date of the daily DAG
3. Updated the reingestion DAG

### API responses

After testing this DAG after the merge of #821, I noticed that we were receiving no data for the run. The logs had `Request failed with success = "False"` in each of the data pull tasks. After some investigation, it appeared that the value for `success` that we were receiving was actually the Python/JSON native `true`, rather than `"True"`. @stacimc mentioned that she was able to test the DAG successfully while the PR was being reviewed. It looks like this was because the initial check for `success` was in `get_should_continue` rather than `get_batch`. Based on the lifecycle of the `ProviderDataIngester`, the result of `get_should_continue` is only used _after at least one batch has been processed_:

https://github.com/WordPress/openverse-catalog/blob/91ab102df69a004224e339c1a86d6ee8f552f8a0/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py#L183-L190

This meant that during testing, data was still being processed even though the `success` check was causing `should_continue` to be `False`. Moving the check into `get_batch` right before merging caused the DAG to no longer process any data.

To rectify this, I've updated the values stored in our response JSONs for testing and updated the check itself to use Python native boolean types.

### Updated start date

Since we have the reingestion workflow, we don't need this DAG to run over the last decade's worth of data. This is in-line with other changes we've made to dated DAGs.

### Updated reingestion workflow

I've also updated the reingestion workflow to use the new provider data ingester base class. I was able able to test this and have it run successfully!

## Testing Instructions

1. Add the Europeana API key to your local environment if you haven't already
2. Run the daily `europeana_workflow`
3. Run the reingestion `europeana_ingestion_workflow`

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[...]

PR #849, opened 2022-10-28
Remove legacy provider DAG logic

### LINKED ISSUE #824: Remove legacy provider logic for DAG wrapper functions
## Description

While working on the provider refactors, we needed to support both the ingestion classes _and_ the module-based legacy providers. Once the provider refactors are complete, we can remove the logic we had around handling these legacy provider scripts:

https://github.com/WordPress/openverse-catalog/blob/b4ef93ce5f825e4ae0e3ba3814340ec42ade7fa2/openverse_catalog/dags/providers/factory_utils.py#L63-L66

We can also modify the `ProviderWorkflow` class to make more assumptions about the workflow based on the ingestion class alone. We should now be able to dynamically determine the provider name and media types rather than having to explicitly define them (for most cases):

https://github.com/WordPress/openverse-catalog/blob/8c26510d7479fbac23803361b4be08c12e43f045/openverse_catalog/dags/providers/provider_workflows.py#L147-L151

We'll need to complete all refactors in order to remove these legacy shims.

## Additional context

For more context on the latter point, see this discussion: https://github.com/WordPress/openverse-catalog/pull/790#discussion_r1001068351

## Implementation

stacimc:
## Fixes

Fixes #824 by @AetherUnbound 
## Description

* Removes now-unused logic branches that handled legacy provider scripts
* Updates the ProviderWorkflow and ProviderReingestionWorkflow dataclasses to infer the provider script, provider name, and media types from the configured ingestion class
  * This means that to configure a new DAG, *all* you need to define in the dataclass is the Ingester class! Everything else is optional
* Updates some docs

**Note: #848 should be merged first and this will need to be rebased to pull in Europeana changes.**

Not included:
It may be time to revisit the `generate_tsv_filenames` logic, which I have not touched in this PR beyond deleting fully-unused legacy branches. Should we open a new issue for this?

## Testing Instructions

```just test```

Test some DAGs! Make sure to try at least one audio-only, one image-only, one audio-and-image, and one reingestion workflow. If you're looking for a thrill set `INGESTION_LIMIT` to 10 and turn them all on 🥳 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_reingestion_workflows.py
+++ b/openverse_catalog/dags/providers/provider_reingestion_workflows.py
@@ -47,13 +47,18 @@ class ProviderReingestionWorkflow(ProviderWorkflow):
 
     def __post_init__(self):
         if not self.dag_id:
-            self.dag_id = f"{self.provider_script}_reingestion_workflow"
+            # Call super() first to initialize the provider_name
+            super().__post_init__()
+            # Override the dag_id
+            self.dag_id = f"{self.provider_name}_reingestion_workflow"
+            return
+
+        super().__post_init__()
 
 
 PROVIDER_REINGESTION_WORKFLOWS = [
     ProviderReingestionWorkflow(
         # 60 total reingestion days
-        provider_script="europeana",
         ingestion_callable=EuropeanaDataIngester,
         max_active_tasks=3,
         pull_timeout=timedelta(hours=12),
@@ -63,7 +68,6 @@ PROVIDER_REINGESTION_WORKFLOWS = [
     ),
     ProviderReingestionWorkflow(
         # 128 total reingestion days
-        provider_script="flickr",
         ingestion_callable=FlickrDataIngester,
         pull_timeout=timedelta(minutes=30),
         daily_list_length=7,
@@ -75,7 +79,6 @@ PROVIDER_REINGESTION_WORKFLOWS = [
     ),
     ProviderReingestionWorkflow(
         # 64 total reingestion days
-        provider_script="metropolitan_museum",
         ingestion_callable=MetMuseumDataIngester,
         max_active_tasks=2,
         pull_timeout=timedelta(hours=12),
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: openverse_catalog/dags/providers/provider_reingestion_workflows.py [hunk 1]
```diff
@@ -46,24 +46,19 @@ class ProviderReingestionWorkflow(ProviderWorkflow):
...
     def __post_init__(self):
         if not self.dag_id:
-            self.dag_id = f"{self.provider_script}_reingestion_workflow"
+            # Call super() first to initialize the provider_name
+            super().__post_init__()
+            # Override the dag_id
+            self.dag_id = f"{self.provider_name}_reingestion_workflow"
+            return
+
+        super().__post_init__()
```
AetherUnbound, line 55: Do we need to call `super().__post_init__()` twice? 😮

AetherUnbound (changes requested):
Yay for deleted code! 🥳 I have one note, and it looks like the `--extended` tests are failing 🤔

openverse_catalog/dags/providers/provider_reingestion_workflows.py
stacimc: Yep, unfortunately. If a `dag_id` wasn't provided in the configuration, we want to override it. We need to call super first in order to set the `provider_name` (which is used in building the dag_id), and _then_ we can override it. We can't do this check _after_ calling super, because `dag_id` will have been set (to the regular `_workflow` name) during the super call. So we must call super within the branch.

But if `dag_id` was provided, we still want to call super(), so we also have to do it outside the branch. The return on L53 prevents it from actually being run twice.

AetherUnbound: Oh, oops! Okay I see, we're not actually calling it twice as you point out. Thanks!

AetherUnbound (approved):
🚀 Woohoo!

zackkrida (approved):
> If you're looking for a thrill set INGESTION_LIMIT to 10 and turn them all on 🥳

Indeed, I am, so I turned on _many_ of the DAGs 😆 awesome work!

PR #851, opened 2022-10-31
Filter out the extra fields with no additional info in Cleveland Museum metadata

satya-vinay:
## Fixes
Fixes https://github.com/WordPress/openverse-catalog/issues/825 by @AetherUnbound

## Description
This PR removes the default empty string assignment to metadata, while extracting metadata fields from input data object.
So that all the None value fields would be filtered out at the second to last line of the _get_metadata function, resulting in only metadata fields which has info.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
@@ -105,7 +105,7 @@ class ClevelandDataIngester(ProviderDataIngester):
             "credit_line": data.get("creditline"),
             "classification": data.get("type"),
             "tombstone": data.get("tombstone"),
-            "culture": ",".join([i for i in data.get("culture") if i is not None]) if data.get("culture") is not None else None,
+            "culture": ",".join([i for i in data.get("culture", []) if i is not None]) or None,
         }
         metadata = {k: v for k, v in metadata.items() if v is not None}
         return metadata
--- a/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
@@ -105,7 +105,8 @@ class ClevelandDataIngester(ProviderDataIngester):
             "credit_line": data.get("creditline"),
             "classification": data.get("type"),
             "tombstone": data.get("tombstone"),
-            "culture": ",".join([i for i in data.get("culture", []) if i is not None]) or None,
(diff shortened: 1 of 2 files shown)


openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
```diff
@@ -99,13 +99,13 @@ def _get_int_value(data: Dict, key: str) -> int | None:
...
-            "classification": data.get("type", ""),
-            "tombstone": data.get("tombstone", ""),
-            "culture": ",".join([i for i in data.get("culture", []) if i is not None]),
+            "accession_number": data.get("accession_number"),
+            "technique": data.get("technique"),
+            "date": data.get("creation_date"),
+            "credit_line": data.get("creditline"),
+            "classification": data.get("type"),
+            "tombstone": data.get("tombstone"),
+            "culture": ",".join([i for i in data.get("culture") if i is not None]) if data.get("culture") is not None else None,
```
AetherUnbound, line 108: I think one way to potentially change this might be:
```suggestion
            "culture": ",".join([i for i in data.get("culture", []) if i is not None]) or None,
```
If `data["culture"]` is indeed missing, then the expression will evaluate to `",".join([])` which then becomes `""`. When used with an `or` expression, if the first term is falsy Python will use the second term which would be `None` in this case:

```python
[ins] In [4]: print(f'{",".join([])=}')
",".join([])=''

[ins] In [5]: print(f'{",".join([]) or None=}')
",".join([]) or None=None
```

AetherUnbound (commented):
This is great, thank you @satya-vinay for adding tests as well! 😄 I have one suggested improvement.

openverse_catalog/dags/providers/provider_api_scripts/cleveland_museum.py
satya-vinay: I have used this approach just to make no computation at LHS when culture is None.😅
But I guess using `or` would be much simpler and readable. Thanks for your detailed explanation😁

AetherUnbound:
Looks like linting is still failing, you can address this by running `just lint` locally and committing the changes.

AetherUnbound (approved):
Fantastic, thanks for the contribution! 🚀

stacimc (approved):
Looks great! Thanks for the contribution, and for adding tests @satya-vinay 🎉

Issue #855, opened 2022-11-02
Upgrade to Python 3.11


AetherUnbound:
## Description

[Python 3.11 is finally out](https://docs.python.org/3/whatsnew/3.11.html), and with it comes a tremendous potential speedup:

> Python 3.11 is between 10-60% faster than Python 3.10. On average, we measured a 1.25x speedup on the standard benchmark suite.

There are also some other exciting changes, including [`TypedDict` optional fields improvements](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep655), [enriched exceptions](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep678), and [finer-grained exceptions](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep657).

We should try to get on this version as soon as possible, but unfortunately we're dependent on Airflow being Python 3.11 compatible. Work is being done upstream to accomplish this, and we should try to monitor it closely!

https://github.com/apache/airflow/issues/27292

https://github.com/apache/airflow/pull/27264

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

dhruvkb:
This issue should go in `WordPress/openverse`. We would definitely want Python 3.11 in the API and ingestion-server as well!

AetherUnbound:
Agreed, we'll want them upgraded too! I'm assuming dependabot will actually cover those two repos next month, but I think a separate issue in this repo is reasonable since we're pretty heavily dependent on Airflow supporting 3.11 before advancing on this.

PR #874, opened 2022-11-21
Use Airflow base Docker image

### LINKED ISSUE #532: Investigate using Airflow's docker image as a base
## Description

We're presently using `python:3.10-slim` as our base image. Airflow provides their own images that we could potentially use as an alternative base: https://airflow.apache.org/docs/docker-stack/index.html.

We should investigate whether this would be advantageous for us (easier to upgrade versions, not have to manage constraints, smaller image size, etc).

## Additional context

See #502 for more context

## Implementation

AetherUnbound:
## Fixes

Fixes #532 by @AetherUnbound

## Description

This PR replaces our current custom Airflow docker image with the official upstream Airflow docker image. With this comes a number of complications and changes. I don't see this as a solid "win". There are lots of pros and cons that are worth weighing before we decide to merge (or not) this change:

**Pros**:
- Running database migrations on startup is baked into the image for us and just requires setting an environment variable.
- Creating a web UI admin user is as easy as setting a few variables, we no longer need to add logic for this.
- We can remove the bespoke `wait_for_db.py` script since [waiting for the database to be ready is baked in](https://airflow.apache.org/docs/docker-stack/entrypoint.html#waits-for-airflow-db-connection).
- The main processes no longer runs as the `root` user, which [is more secure](https://docs.bitnami.com/tutorials/why-non-root-containers-are-important-for-security) and follows the principle of least privilege.
- Reduced maintenance burden - the Airflow folks put in a lot of effort to make sure that the Docker images are production ready.
- We would have separate containers for the webserver & scheduler (rather than having them run on the same container), which helps isolate those services more.

**Cons**:
- Running as a non-root user can cause complications with bind-mounted volumes. Permissions on the host machine will no longer match up with the Docker user, so certain operations have to be adjusted. This can be seen in changes to the `justfile` (specifically the `generate-dag-docs` recipe, which can no longer write out the file without special considerations) or to tests (see the changes to `test_create_provider_ingester.py`). This is not necessarily a bad thing, but it has brought me a bit of frustration and can cause trouble and strange behavior down the line.
- Startup time is slower. Since we're now handing over management of the migrations to Airflow and the upstream entrypoint has [lots of other startup behavior](https://airflow.apache.org/docs/docker-stack/entrypoint.html), there's more that runs at each container start. In production this is not a problem (since the containers shouldn't be starting & stopping regularly), but it is noticeable in development. Many of our `just` recipes take a few seconds longer to run. Nothing insurmountable from my perspective, just something I've noticed.
- Held back on Python releases. We were able to jump on Python 3.10 before it was officially supported by Airflow. This was because Airflow has a large number of provider libraries which it also supports, and it needed to ensure compatibility with all of those libraries before the application as a whole was ready for Python 3.10. We use a very small portion of the available providers, so as it stands we are held back from upgrading to 3.11 until those upstream issues with providers we don't use are resolved (see #855).
- We now need to set up a CSRF secret key for **both development and production environments** (they are currently defined in neither).

**Other changes**:
- Airflow home folder got changed from `/usr/local/airflow` to `/opt airflow`
- I fixed an issue with the create provider ingester tests wherein we were accidentally handing in a string where a list should go and it was producing one provider for each letter of the word `image`.

## Testing Instructions

Run `just build` on this branch. Make sure you're able to run every `just` recipe we have without error. You may need to run `just recreate` if you encounter any errors initially.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


AetherUnbound:
As expected, there are a few issues with permissions now that we're using a non-root docker user. One test and the `generate-dag-docs` just recipe fails. I'll try to think about how to get around this.

docker/airflow/Dockerfile
```diff
@@ -33,21 +34,19 @@ ENV AIRFLOW__LOGGING__REMOTE_LOG_CONN_ID=aws_default
 ENV AIRFLOW__LOGGING__REMOTE_BASE_LOG_FOLDER=s3://openverse-airflow-logs
 
 
-RUN apt-get update && apt-get -yqq upgrade && apt-get -yqq install \
+USER root
+RUN apt-get update && apt-get -yqq install \
```
sarayourfriend, line 38: Why no more `upgrade`?

AetherUnbound: There were some serious version conflicts/difficulties in doing this with the Airflow image. I should have noted exactly what the problem was, I can try and re-add it to note the specific error I received.

docker-compose.override.yml
```diff
@@ -76,26 +94,29 @@ services:
...
-    volumes:
-      - ./openverse_catalog:/usr/local/airflow/openverse_catalog
-    stdin_open: true
-    tty: true
+      - scheduler
 
 volumes:
   postgres:
   minio:
+  ipython:
```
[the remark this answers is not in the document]
AetherUnbound: This is for retaining `ipython` history across docker restarts. Previously we were mounting a local folder to accomplish this (which I'm not even sure this actually worked as expected due to permissions issues) but now that we're running as the `airflow` user within the container it's a LOT more difficult to try and manage mounting that and making it write-accessible to the `airflow` user. It's way easier to simply add a volume mount here instead, since docker manages those entirely and permissions will be preserved. The downside there is that `just recreate` will destroy that history, but I still find it useful to retain the history where possible during development!

zackkrida:
Weighing the pros and cons I think I am still in favor of this work. The main con I'm actually concerned with is "Running as a non-root user can cause complications with bind-mounted volumes." because it admittedly sounds a bit vague. Is there anything we can do in terms of documentation to help mitigate future pain from this one? The others feel like reasonable tradeoffs but this one and it's ability to cause future unforeseen frustration or issues seems unfortunate.

rwidom:
> Weighing the pros and cons I think I am still in favor of this work. The main con I'm actually concerned with is "Running as a non-root user can cause complications with bind-mounted volumes." because it admittedly sounds a bit vague. Is there anything we can do in terms of documentation to help mitigate future pain from this one? The others feel like reasonable tradeoffs but this one and it's ability to cause future unforeseen frustration or issues seems unfortunate.

Yeah, does this mean that code changes will no longer be automatically synced to the docker container? That could mean a lot more `just recreate`s for developers. Or maybe I'm misunderstanding and/or there is a workaround I'm not thinking of... ???

AetherUnbound:
> Weighing the pros and cons I think I am still in favor of this work. The main con I'm actually concerned with is "Running as a non-root user can cause complications with bind-mounted volumes." because it admittedly sounds a bit vague. Is there anything we can do in terms of documentation to help mitigate future pain from this one?

I think a lot of the concern I had in that department was centered around 1) the [issues I had in a personal project](https://github.com/OrcaCollective/techbloc-airflow/pull/8) which required [mounting SSH keys in a peculiar way to ensure permissions were preserved](https://github.com/OrcaCollective/techbloc-airflow/pull/7/files#diff-6f9d41d046756f0ddc2fcee0626bdb50100d12b88f293734eff742818e03efa2R64-R73) and 2) the [workaround we have to do for the DAG doc generation script](https://github.com/WordPress/openverse-catalog/pull/874/files#diff-deb9bb56fb122db0b605aa5b63f95a4665c905b18dd670e1fa6c877576a94ff1R126) and [provider script generation tests](https://github.com/WordPress/openverse-catalog/pull/874/files#diff-e8749752ee7270716e940381f68cfd05deaacfcef29423e7f1e82a0fcf9dcd19R51-R57). Admittedly I was able to find a solution for the former (and most of the trouble I had with it was an unrelated issue with `cp dir/*` not receiving any contents and failing because of _that_, not permissions).

This issue will only ever come up in the following scenarios:
1. We need to add a new bind mount to the container which requires write permissions
2. We add a new script/task/something that needs to write to an _existing_ bind mount (a la DAG doc generation or provider script generation)

Given all that context, where do you think documentation for this scenario might be best suited?

> Yeah, does this mean that code changes will no longer be automatically synced to the docker container? That could mean a lot more just recreates for developers. Or maybe I'm misunderstanding and/or there is a workaround I'm not thinking of... ???

Good clarification question - this should only affect _write permissions_ for bind-mounted volumes. Places where we only need to read files (like for DAG parsing!) should not be affected! Auto-reload and (in production) auto-deployment will both work as they did before 😄

stacimc:
This looks good to me. `just` recipes are running well -- I'm actually not even seeing a perceptible difference in how long it takes to run them locally. Initially had CSRF problems while playing around with local Airflow, but all is well after I remembered to update the `.env` with the secret key 🙂 (including testing auto-reload!)

Looking at pros/cons, I'm happy to move forward with this. Just echoing a (small) concern about documentation given the need for special handling in this scenario:

> We add a new script/task/something that needs to write to an existing bind mount (a la DAG doc generation or provider script generation, more likely but still very uncommon)

Maybe we could throw something into the `docs` folder? Or even just a small note in the main README 🤔

AetherUnbound:
Thanks Staci! I was trying to think where the best place for a piece of info like that to go would be (`justfile` since it's likely where we'd add the script? `docker-compose.*.yml` since that'd be where we might add a bind mount if needed?), README seems like the best and most general place 👍🏼

AetherUnbound:
Just a note for myself: **I will need to update the Airflow version** (and this can supersede #911)

AetherUnbound:
I'm actually going to hold off on upgrading to Airflow 2.5.0 in this PR since the changes are non-trivial, but I've made #924 to track that.

AetherUnbound:
@krysal @stacimc this should be ready for review now!

AetherUnbound:
I've tested this as well by building & starting the image from `main`, then changing my local repo to this PR. DAGs still function and nothing should break 😄

stacimc:
After `just build` on main, switching to this branch, and running `just build` once more, everything looks good in my local Airflow. Old DagRuns are still present, and I was able to run several DAGs as tests (tested with `jamendo`, `metropolitan`, and `audio_data_refresh`).

Locally I'm seeing three test failures consistently with `PermissionError`s, though:

```
         - ../../usr/local/airflow/tests/dags/common/storage/test_audio.py:103 test_AudioStore_add_item_flushes_buffer
         - tests/dags/common/storage/test_media.py:91 test_MediaStore_add_item_flushes_buffer
         - ../../usr/local/airflow/tests/dags/providers/provider_api_scripts/test_cleveland_museum.py:168 test_handle_response
```

AetherUnbound:
@stacimc you may need to run `just recreate` to recreate the volumes those tests write too, since they were likely created with the old `root` user and the new `airflow` user doesn't have access.

stacimc (approved):
Passing after a recreate 😃 LGTM

docker/airflow/Dockerfile
```diff
@@ -1,17 +1,18 @@
 # Cannot use $PYTHON_VERSION because that is overwritten by the base image
 # with the patch version (we only care about major and minor version)
 ARG PROJECT_PY_VERSION
```
krysal, line 3: Minor suggestion for clarity:
```suggestion
ARG PROJECT_PYTHON_VERSION
```

docker/airflow/Dockerfile
```diff
@@ -33,21 +34,19 @@ ENV AIRFLOW__LOGGING__REMOTE_LOG_CONN_ID=aws_default
...
+RUN apt-get update && apt-get -yqq install \
     build-essential \
     libpq-dev \
     libffi-dev \
     && apt-get autoremove -y \
     && rm -rf /var/lib/apt/lists/*
-RUN useradd -m -d ${AIRFLOW_HOME} airflow && \
-    mkdir -p ${OUTPUT_DIR} && \
-    chown airflow:airflow ${OUTPUT_DIR}
+RUN mkdir -p ${OUTPUT_DIR} /home/airflow/.ipython && chown airflow ${OUTPUT_DIR} /home/airflow/.ipython
```
krysal, line 44: New lines helps the understanding of each statement:
```suggestion

RUN mkdir -p ${OUTPUT_DIR} /home/airflow/.ipython && \
    chown airflow ${OUTPUT_DIR} /home/airflow/.ipython
```
and maybe add a comment that is for retaining `ipython` history?

krysal (approved):
I'd like to deep dive more into what the base Airflow image brings but don't want to withhold this any longer. So approving to unblock it ✅

I left some minor suggestions.

docker/airflow/Dockerfile
AetherUnbound: We'd need to change this in a few places (I think in the CI/CD files files as well):

https://github.com/WordPress/openverse-catalog/blob/fba7566540343549b5a01211e69c9237c0ed6863/.github/workflows/ci_cd.yml#L83

If it's okay with you, I'll defer this for now and we can make the change in a separate PR 🙂

docker/airflow/Dockerfile
AetherUnbound: Done in 31988874c2ae6f7d9062d80a47c1d4ee341f5b0f!

Issue #665, opened 2022-08-16
Finnish museum `pull_data` exceeds timeout


stacimc:
## Description

The Finnish Museum DAG has been consistently timing out at the `pull_data` step. Recently its `pull_data` task was updated to have a 5 day execution_timeout, but the most recent DagRun had to be manually failed after the task was still running after 15 days.

We should investigate why the task was able to exceed this limit.

## Reproduction

Observe the logs in Airflow for the run dated 2022-07-01.
You can also inspect the Task Instance Details and see that the `execution_timeout` is set to 5 days as expected.

## Resolution

- [ ] 🙋 I would be interested in resolving this bug.

AetherUnbound:
It is worth taking a look at #621 first to see if that would be one way to tackle the large amounts of data we receive from this provider.

AetherUnbound:
Bumping this to high as the provider DAG is currently disabled.

stacimc:
Investigating #621 now

AetherUnbound:
@stacimc do you think this can be closed by #879 or do we need to run the DAG first to find out?

stacimc:
@AetherUnbound Good question! I think it's probably fine to close, since we won't *really* know until we hit an ingestion day with a lot of data, and we can always re-open. In local testing, I was able to fully ingest data for one such day without hitting the timeout, so we should be good!

PR #879, opened 2022-11-28
Make Finnish DAG dated

### LINKED ISSUE #621: Investigate converting Finnish Museums to a dated DAG
## Problem

The Finnish Museum DAG has consistently run up against the 3 hour execution limit, and generates so many logs that Airflow is unable to render the log page. We have also been having issues loading the data volume that Finnish supplies (see #499 & #392)

## Description

We should investigate whether it would be possible to convert this DAG into a dated DAG. If it would be possible to do so, we should move forward with converting it. This will help reduce the volume pulled on each run and should address the timeout issues.

## Implementation

stacimc:
## Fixes

Fixes #621 and hopefully #665

## Description

Converts the Finnish Museum provider DAG to a dated DAG, using the `last_indexed` filter to get records that were updated since the last DagRun.

In production, the Finnish provider DAG is ingesting huge amounts of data and ultimately timing out. Converting the DAG to a _dated_ DAG reduces the amount of data being consumed at any one time, and we hoped this would resolve the issue on its own.

Unfortunately I discovered that even when converted to a daily DAG, we still encounter large batches of data. Perhaps unsurprisingly for a museum provider, it seems that data is indexed very sparsely and in large batches at once. That means the huge majority of days have no data, while a small number of days have _very large_ amounts. Moreover, these batches are large enough to be buggy on their own. I described my steps to investigate and reproduce this in [this comment](https://github.com/WordPress/openverse-catalog/pull/879#issuecomment-1333078808), but essentially: for large querysets, the API will continue to return fallacious data to us no matter how large a `page number` we request, apparently into perpetuity. As a result ingestion continues until it times out.

I addressed this by copying a strategy from our Flickr DAG: we now break the ingestion day up into 48 half-hour increments and ingest data one time slice at a time. This prevents us from ever ingesting too large a dataset at once.

In summary, this PR:
* converts the Finnish museum to a dated DAG
* further breaks ingestion into half-hour increments to avoid the buggy API behavior
* updates the `start_date` of the DAG to November 2015, which is the earliest data I could find for the API
* adds some extra logging

## Notes
* I chose not to try pulling out the time splitting code into a shared utility, as this logic is currently _not working_ for Flickr. We may revisit it as part of that work
* This DAG would be incredibly noisy if we tried to backfill it all the way to 2015, so we should NOT do so unless we come up with a solution for silencing the Slack notifications.
* Because we're filtering on `last_indexed`, we'll handle updates in the regular ingestion flow (ie, no need for a separate reingestion DAG!) The `catchup` feature will handle recovering skipped days if the DAG is ever turned off for any reason.

I'll do some more investigation to try to understand what the current status of Finnish Museum data is in our catalog. It's not immediately clear to me how much we need to backfill, and we'll want to be pretty targeted about it since this will be a very noisy DAG. If #880 is merged, we'll be able to manually run the DAG for specific dates 😃 

## Testing Instructions

`just test`, run the Finnish DAG locally.

I was able to reproduce the time-out bug locally by running the DAG for November 3rd, 2015, and then confirm that ingestion worked for that day after adding in the time-splitting logic. If you would like to confirm this too, here's how:
* Make sure you don't have an `ingestion_limit` set
* Delete any local DagRuns you may have of the Finnish DAG
* Enable the Finnish DAG. It should automatically start running at `November 1st, 2015`, so you'll see the first two DagRuns pass with zero records. Then it will begin the "buggy" day
* Let it run! I got `147,245` records in a little over 2 hours 🎉 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


stacimc:
Leaving some notes here about my investigation into what’s happening with this provider, since it ended up being a very interesting problem and may be helpful for reviewers/for future reference.

It looks like the Finnish Museums API behaves unexpectedly when querying large amounts of data. You can reproduce this yourself:

This query looks for images associated with the "0/SA-kuva/" building, last indexed on `2015-11-03`:

```
curl -X ‘GET’ \
    ‘https://api.finna.fi/api/v1/search?type=AllFields&filter%5B%5D=format%3A%220%2FImage%2F%22&filter%5B%5D=building%3A%220%2FSA-kuva%2F%22&filter%5B%5D=last_indexed%3A%22%5B2015-11-03T00%3A30%3A00Z%20TO%202015-11-04T00%3A00%3A00Z%5D%22&page=1&limit=100'
    -H ‘accept: application/json’
```

You’ll see a response with `”resultCount”: 147245`. So setting the `limit` to 100 items per page, we _should_ expect to get back 1,473 pages back from the API. However, we’re getting data back _no matter what_ page number we send, even arbitrarily large numbers. Here’s the request for page 10,000:

```
curl -X ‘GET’ \
    ‘https://api.finna.fi/api/v1/search?type=AllFields&filter%5B%5D=format%3A%220%2FImage%2F%22&filter%5B%5D=building%3A%220%2FSA-kuva%2F%22&filter%5B%5D=last_indexed%3A%22%5B2015-11-03T00%3A30%3A00Z%20TO%202015-11-04T00%3A00%3A00Z%5D%22&page=10000&limit=100'
    -H ‘accept: application/json’
```

This strange behavior does _not_ happen with smaller querysets. For example, the following day has only 8 records:

```
# This request results in a response containing all 8 records
curl -X ‘GET’ \
    ‘https://api.finna.fi/api/v1/search?type=AllFields&filter%5B%5D=format%3A%220%2FImage%2F%22&filter%5B%5D=building%3A%220%2FSA-kuva%2F%22&filter%5B%5D=last_indexed%3A%22%5B2015-11-04T00%3A30%3A00Z%20TO%202015-11-05T00%3A00%3A00Z%5D%22&page=1&limit=100'
    -H 'accept: application/json'

# If you update `page` to 2, the response contains no additional records:
# {"resultCount":8,"status":"OK"}
```

So if we try to process a **large** dataset as normal, we’ll keep ingesting data forever and eventually time out, just as we see in production. I let the DAG run locally for a few hours until the ingested record count hit >440k and inspected the resulting tsv with awk out of curiosity. Almost every row in the tsv was triplicated, meaning that we had just about the expected number of unique records. Unfortunately, it doesn’t look like the API is sending us the data repeatedly in order (meaning I don’t think we can just cut off ingestion at some arbitrary point).

Interestingly, I was aware that we had a very similar problem with Flickr with large datasets. With Flickr, we addressed it by splitting the ingestion day into half hour increments and ingesting data for each time slice at a time. Although this strategy does not appear to be working any longer for Flickr, I tested it out here and was able to get a full, successful run of the Finnish DAG for Nov 3rd, 2015 🎉

openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
```diff
@@ -26,16 +43,71 @@ class FinnishMuseumsDataIngester(ProviderDataIngester):
...
+        ]
+        return pair_list
+
     def ingest_records(self, **kwargs):
         for building in BUILDINGS:
             logger.info(f"Obtaining images of building {building}")
-            super().ingest_records(building=building)
+
+            for start_ts, end_ts in self.timestamp_pairs:
+                self.new_iteration = True
```
stacimc, line 92: Because this DAG is so sparse, one optimization that might be nice would be to first request the entire day's worth of data, and then only bother breaking it down into increments if the `resultCount` is above some threshold (or even just non-zero!) 🤔

AetherUnbound: That'd be a good optimization! Maybe we could make an issue for it to include down the line?

AetherUnbound (approved):
Wow, incredible work troubleshooting & diagnosing this one! 🤯 I'm not able to test these changes but code seems solid. I wonder, the behavior you observed might be what's causing #450 as well 😮 At least the first noticed duplication case.

openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
stacimc: I agree! #913

krysal:
I'll I ran out of time today but this is the first thing I'll review tomorrow!

krysal (approved):
> Because we're filtering on last_indexed, we'll handle updates in the regular ingestion flow (ie, no need for a separate reingestion DAG!) The catchup feature will handle recovering skipped days if the DAG is ever turned off for any reason.

Awesome, thanks for all the details explaining what was happening here. Glad that we don't need another DAG for Finnish 🎉 All looks great here to me.

krysal:
I started getting data at the interval `2015-11-03T15:30:00Z` to `2015-11-03T16:00:00Z` ✅

PR #880, opened 2022-11-29
Allow overriding the date from the DagRun conf

stacimc:
## Description

Adds an option to the DagRun conf to let you override the `date` of a run of a dated provider DAG.

When you manually run a dated DAG, it is always passed in today's date. This option allows you to pass in any arbitrary date you'd like. While I don't necessarily expect this to be used frequently in production, it has come in handy many times for local testing.

## Testing Instructions

To trigger a DAG using a date override, select the `Trigger DAG w/ config` option seen here:

Screen Shot 2022-11-28 at 4 15 23 PM

Then pass your chosen date in using a `YYYY-MM-DD` format like this:
```{"date": "2022-04-10"}```

Pick a dated DAG, like Europeana:
* Try running it normally by enabling the DAG locally and letting a scheduled run start. Verify that it correctly uses the scheduled date.
* Try running it **manually** without a DagRun conf. Verify that it uses today's date.
* Try triggering the DAG w/ a config and pass in a different date. Verify the custom date is used. You should see a message in the logs like `Using date <date> from dagrun conf.`

## Notes
* The DAG will error if nonsense data or an incorrectly formatted date is passed in.
* You *can* set a date override on any provider DAG, even a non-dated one. In this case it simply won't do anything.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
@@ -2,6 +2,7 @@ import json
 import logging
 import traceback
 from abc import ABC, abstractmethod
+from datetime import datetime
 
 from airflow.exceptions import AirflowException
 from airflow.models import Variable
@@ -125,6 +126,12 @@ class ProviderDataIngester(ABC):
 
         # dag_run configuration options
         conf = conf or {}
+
+        # Allow overriding the date with a %Y-%m-%d string from the dagrun conf.
+        date_override = conf.get("date")
+        if date_override and datetime.strptime(date_override, "%Y-%m-%d"):
+            logger.info(f"Using date {date_override} from dagrun conf.")
+            self.date = date_override
 
         # Used to skip over errors and continue ingestion. When enabled, errors
         # are not reported until ingestion has completed.
--- a/tests/dags/providers/provider_api_scripts/test_provider_data_ingester.py
+++ b/tests/dags/providers/provider_api_scripts/test_provider_data_ingester.py
@@ -88,6 +88,30 @@ def test_batch_limit_is_capped_to_ingestion_limit():
         assert ingester.limit == 20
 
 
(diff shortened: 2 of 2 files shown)


krysal (approved):
The code looks good to me, though, what is this date used for? It's not totally clear to me. 

Is there a place where we can find these configuration parameters with their descriptions or use cases documented? Yesterday I was looking to limit the number of records pulled, and I wasn't sure if there is a param for it (there is an Airflow Variable but something for an individual DAG run? would be nice to add it if it's not already done).

stacimc:
> The code looks good to me, though, what is this date used for? It's not totally clear to me.

This is the ingestion `date` that's used in dated DAGs. Typically, the logical date of the DagRun is passed in automatically. The goal of this PR is to allow a dev a way to override that to whatever date we want, because otherwise there isn't currently an easy way to run ingestion for an arbitrary day.

(A recent use case for me was testing with Finnish Museums. I wanted to reproduce a bug that only happens on certain days, and found it useful to be able to easily kick off ingestion for specific dates.)

> Is there a place where we can find these configuration parameters with their descriptions or use cases documented? 

There isn't, but there is an issue to add documentation! #736 Couldn't agree more, it's easy to forget what the available options are.

> Yesterday I was looking to limit the number of records pulled, and I wasn't sure if there is a param for it (there is an Airflow Variable but something for an individual DAG run? would be nice to add it if it's not already done).

The `ingestion_limit` Airflow variable is what you're thinking of, and you're right that it applies to _all_ DAGs. Having that available as an option for individual DAG runs sounds like it could be very handy! I made #906 to track it, thanks 😄

AetherUnbound (approved):
Great idea! I can't test this locally at this time, but the implementation & test look 💯

AetherUnbound:
Finally able to test, works great!

Issue #688, opened 2022-08-19
Move API key Variable access to provider ingester init function


AetherUnbound:
## Description

Presently the API keys are pulled from Airflow Variables, typically at the top of a provider script. Here's an example:

https://github.com/WordPress/openverse-catalog/blob/0072114ea479b378a94e1144340935023acbdf3d/openverse_catalog/dags/providers/provider_api_scripts/brooklyn_museum.py#L22

Unfortunately this means that the Variable is retrieved from Airflow (and thus queried from the database) on every DAG parse iteration. It isn't needed until actual DAG runtime, so we should consider deferring its instantiation until the `ProviderDataIngester::__init__` function rather than at the top of the module.

Not all of these scripts are converted into the new provider ingester class, so we'll have to wait to complete this ticket until (or as part of) milestone v1.3.2. 

## Additional context

Below is an example of the current process. I added the following line to Airflow's `variable.py` file within the docker container to track usage.

```bash
$ diff -Naur /home/madison/programs/anaconda3/envs/openverse-catalog/lib/python3.10/site-packages/airflow/models/variable.py variable.py 
--- /home/madison/programs/anaconda3/envs/openverse-catalog/lib/python3.10/site-packages/airflow/models/variable.py	2022-07-13 15:23:29.445279243 -0700
+++ variable.py	2022-08-19 15:38:52.992392655 -0700
@@ -130,6 +130,7 @@
         :param default_var: Default value of the Variable if the Variable doesn't exist
         :param deserialize_json: Deserialize the value to a Python dict
         """
+        print(f"Running get for {key}!")
         var_val = Variable.get_variable_from_secrets(key=key)
         if var_val is None:
             if default_var is not cls.__NO_DEFAULT_SENTINEL:
```

I then ran `airflow dags list` within the container, here was the output:

```bash
airflow@604f79cb5c26:~$ airflow dags list
/usr/local/airflow/.local/lib/python3.10/site-packages/airflow/utils/log/file_task_handler.py:52 DeprecationWarning: Passing filename_template to FileTaskHandler is deprecated and has no effect
Running get for API_KEY_EUROPEANA!
Running get for API_KEY_FLICKR!
Running get for API_KEY_BROOKLYN_MUSEUM!
Running get for API_KEY_FREESOUND!
Running get for API_KEY_JAMENDO!
Running get for API_KEY_NYPL!
Running get for API_KEY_DATA_GOV!
[...]

PR #904, opened 2022-12-02
RDS Snapshot rotation DAG

sarayourfriend:
## Fixes

Fixes https://github.com/WordPress/openverse-infrastructure/issues/94 by @AetherUnbound 

For those who cannot access the issue above, the goal here is to work around RDS's default implementation only allowing daily snapshots while the catalogue only realistically requires a weekly snapshot to be sustainable.

## Description

I am still working on how to test this. I think I'd like to provision a new very inexpensive RDS instance to test it against. It's the only way I can think to actually test this functionality. For the unit tests I'll just be mocking the rds client for the one function that needs testing here (delete previous snapshots).

## Testing Instructions

I'll update this once I have a suggestion for how to test this. Unfortunately this will probably require a live RDS instance safe to test against.

Update your .env file with the new variables.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
@AetherUnbound Can you think of any other reasonable ways to test this other than using a new temporary test-only RDS instance to point the DAG to?

AetherUnbound:
> @AetherUnbound Can you think of any other reasonable ways to test this other than using a new temporary test-only RDS instance to point the DAG to?

Unfortunately not 😅 I would like to consider the upstream operators well-tested, so if we're covering our own logic with the current tests, I think that's sufficient!

sarayourfriend:
> Unless you're hoping to test this manually once to make sure it all runs fine, in which case I say go for it slightly_smiling_face

Yeah, I kind of wanted to test it manually, but it would be easier to do this in the production Airflow, I think, just with a non-critical database :slightly_smiling_face:

openverse_catalog/dags/maintenance/rotate_db_snapshots.py
```diff
@@ -0,0 +1,91 @@
...
+from airflow.providers.amazon.aws.operators.rds import RdsCreateDbSnapshotOperator
+from airflow.providers.amazon.aws.sensors.rds import RdsSnapshotExistenceSensor
+
+
+logger = logging.getLogger(__name__)
+
+DAG_ID = "check_silenced_dags"
+MAX_ACTIVE = 1
+
+AIRFLOW_RDS_ARN = Variable.get("AIRFLOW_RDS_ARN")
```
AetherUnbound, line 31: Just a note, it looks like there are a number of issues related to loading DAGs when this variable is not defined. Do you mind adding a comment to the PR description that reviewers will need to add the new variables to their `.env` file?

openverse_catalog/dags/maintenance/rotate_db_snapshots.py
```diff
@@ -0,0 +1,91 @@
...
+import boto3
+from airflow.decorators import dag, task
+from airflow.models import Variable
+from airflow.providers.amazon.aws.operators.rds import RdsCreateDbSnapshotOperator
+from airflow.providers.amazon.aws.sensors.rds import RdsSnapshotExistenceSensor
+
+
+logger = logging.getLogger(__name__)
+
+DAG_ID = "check_silenced_dags"
```
AetherUnbound, line 28: This is causing duplication errors since this ID is already being used by the `check_silenced_dags` DAG: https://github.com/WordPress/openverse-catalog/blob/06335539a187036435e0cb02fa94ee4324a7c8da/openverse_catalog/dags/maintenance/check_silenced_dags.py#L35

Maybe:
```suggestion
DAG_ID = "rotate_db_snapshots"
```

openverse_catalog/dags/maintenance/rotate_db_snapshots.py
```diff
@@ -0,0 +1,91 @@
...
+MAX_ACTIVE = 1
+
+AIRFLOW_RDS_ARN = Variable.get("AIRFLOW_RDS_ARN")
+AIRFLOW_RDS_SNAPSHOTS_TO_RETAIN = Variable.get(
+    "AIRFLOW_RDS_SNAPSHOTS_TO_RETAIN", default_var="7"
+)
```
AetherUnbound, line 34: FWIW, it's generally a good idea (if we can help it) to refrain from calling `Variable.get` at the module level since that will issue a database call for each `.get` during each DAG parse interval. Since these values aren't needed until runtime, we can use some indirection Airflow provides to prevent having to assign them here. For example, many of the RDS operators have `db_identifier` in their [templated fields](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/operators/rds/index.html#airflow.providers.amazon.aws.operators.rds.RdsCreateDbSnapshotOperator.template_fields). That means that we can use [the template variable reference](https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html#airflow-variables-in-templates) within a string for that value. Below is a diff which shows how this could work:

```diff
diff --git a/openverse_catalog/dags/maintenance/rotate_db_snapshots.py b/openverse_catalog/dags/maintenance/rotate_db_snapshots.py
index b24504c..ea13fc7 100644
--- a/openverse_catalog/dags/maintenance/rotate_db_snapshots.py
+++ b/openverse_catalog/dags/maintenance/rotate_db_snapshots.py
@@ -28,25 +28,18 @@ logger = logging.getLogger(__name__)
 DAG_ID = "rotate_db_snapshots"
 MAX_ACTIVE = 1
 
-AIRFLOW_RDS_ARN = Variable.get("AIRFLOW_RDS_ARN")
-AIRFLOW_RDS_SNAPSHOTS_TO_RETAIN = Variable.get(
-    "AIRFLOW_RDS_SNAPSHOTS_TO_RETAIN", default_var="7"
-)
-
 
 @task()
-def delete_previous_snapshots():
+def delete_previous_snapshots(rds_arn: str, snapshots_to_retain: int):
     rds = boto3.client("rds")
     # Snapshot object documentation:
     # https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DBSnapshot.html
     snapshots = rds.describe_db_snapshots(
-        DBInstanceIdentifier=AIRFLOW_RDS_ARN,
+        DBInstanceIdentifier=rds_arn,
     )
 
-    to_retain = int(AIRFLOW_RDS_SNAPSHOTS_TO_RETAIN)
-
-    if len(snapshots) <= to_retain or not (
-        snapshots_to_delete := snapshots[to_retain:]
+    if len(snapshots) <= snapshots_to_retain or not (
+        snapshots_to_delete := snapshots[snapshots_to_retain:]
     ):
         logger.info("No snapshots to delete.")
         return
@@ -68,13 +61,16 @@ def delete_previous_snapshots():
     catchup=False,
     # Use the docstring at the top of the file as md docs in the UI
     doc_md=__doc__,
+    # Render JSON deserialized values as native objects rather than strings
+    render_template_as_native_obj=True,
 )
 def rotate_db_snapshots():
     snapshot_id = "airflow-{{ ds }}"
+    db_identifier = "{{ var.value.AIRFLOW_RDS_ARN }}"
     create_db_snapshot = RdsCreateDbSnapshotOperator(
         task_id="create_snapshot",
         db_type="instance",
...skipping...
     rds = boto3.client("rds")
     # Snapshot object documentation:
     # https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DBSnapshot.html
     snapshots = rds.describe_db_snapshots(
-        DBInstanceIdentifier=AIRFLOW_RDS_ARN,
+        DBInstanceIdentifier=rds_arn,
     )
 
-    to_retain = int(AIRFLOW_RDS_SNAPSHOTS_TO_RETAIN)
-
-    if len(snapshots) <= to_retain or not (
-        snapshots_to_delete := snapshots[to_retain:]
+    if len(snapshots) <= snapshots_to_retain or not (
+        snapshots_to_delete := snapshots[snapshots_to_retain:]
     ):
         logger.info("No snapshots to delete.")
         return
@@ -68,13 +61,16 @@ def delete_previous_snapshots():
     catchup=False,
     # Use the docstring at the top of the file as md docs in the UI
     doc_md=__doc__,
+    # Render JSON deserialized values as native objects rather than strings
+    render_template_as_native_obj=True,
 )
 def rotate_db_snapshots():
     snapshot_id = "airflow-{{ ds }}"
+    db_identifier = "{{ var.value.AIRFLOW_RDS_ARN }}"
     create_db_snapshot = RdsCreateDbSnapshotOperator(
         task_id="create_snapshot",
         db_type="instance",
-        db_identifier=AIRFLOW_RDS_ARN,
+        db_identifier=db_identifier,
         db_snapshot_identifier=snapshot_id,
     )
     wait_for_snapshot_availability = RdsSnapshotExistenceSensor(
@@ -85,7 +81,10 @@ def rotate_db_snapshots():
         target_statuses=["available"],
     )
 
-    create_db_snapshot >> wait_for_snapshot_availability >> delete_previous_snapshots()
+    create_db_snapshot >> wait_for_snapshot_availability >> delete_previous_snapshots(
+        rds_arn=db_identifier,
+        snapshots_to_retain="{{ var.json.AIRFLOW_RDS_SNAPSHOTS_TO_RETAIN }}",
+    )
 
 
 rotate_db_snapshots()
```

AetherUnbound, line 34: As a side note, we can also use Variable's [`deserialize_json` parameter](https://airflow.apache.org/docs/apache-airflow/stable/concepts/variables.html) to convert this to an `int` implicitly. This would become:

```suggestion
AIRFLOW_RDS_SNAPSHOTS_TO_RETAIN = Variable.get(
    "AIRFLOW_RDS_SNAPSHOTS_TO_RETAIN",
    default_var=7,
    deserialize_json=True,
)
```
In this case if the value in the Variable field the string `"14"`, `AIRFLOW_RDS_SNAPSHOTS_TO_RETAIN` would then become `14`.

openverse_catalog/dags/maintenance/rotate_db_snapshots.py
```diff
@@ -0,0 +1,91 @@
...
+        rds.delete_db_snapshot(DBSnapshotIdentifier=snapshot["DBSnapshotIdentifier"])
+
+
+@dag(
+    dag_id=DAG_ID,
+    # At 00:00 on Saturday, this puts it before the data refresh starts
+    schedule="0 0 * * 6",
+    start_date=datetime(2022, 12, 2),
+    tags=["maintenance"],
+    max_active_tasks=MAX_ACTIVE,
```
AetherUnbound, line 66: Since the task flow is already serial in nature and has its dependencies defined, this line is redundant:
```suggestion
```

tests/dags/maintenance/test_rotate_db_snapshots.py
```diff
@@ -0,0 +1,48 @@
...
+
+    monkeypatch.setattr(boto3, "client", get_client)
+    return rds
+
+
+def _make_snapshot(_id):
+    return {"DBSnapshotIdentifier": _id}
+
+
+@pytest.mark.parametrize(
```
AetherUnbound, line 23: Just a note, if we also allow the snapshots to be configured via function argument as I describe in my comment above, we could try out values other than `7` here to make sure those work in the tests too 😄

AetherUnbound (changes requested):
I'm glad there was already a few operators for us to use! I have a couple of notes and one blocking change.

openverse_catalog/dags/maintenance/rotate_db_snapshots.py
sarayourfriend: Oh dear, I copied this and _definitely_ meant to change the text but just totally forgot what I was doing last week :facepalm: The name is clearly not correct, thank you for the suggested change.

openverse_catalog/dags/maintenance/rotate_db_snapshots.py
AetherUnbound: I would not say this DAG is an edge case and yes we should change those other instances if possible to! We have this issue for tracking the removal of the API key retrieval at the module level: https://github.com/WordPress/openverse-catalog/issues/688

I'll make another for changing the other instances 🙂

sarayourfriend: Great. This seems like something perfect for https://github.com/hchasestevens/bellybutton along with using datetime instead of pendulum.

AetherUnbound (commented):
It looks like tests are failing, apologies for not including the appropriate test changes in my original diff. Here's a potential change based on the update:

```diff
diff --git a/tests/dags/maintenance/test_rotate_db_snapshots.py b/tests/dags/maintenance/test_rotate_db_snapshots.py
index 0fecc61..2f94216 100644
--- a/tests/dags/maintenance/test_rotate_db_snapshots.py
+++ b/tests/dags/maintenance/test_rotate_db_snapshots.py
@@ -21,17 +21,21 @@ def _make_snapshot(_id):
 
 
 @pytest.mark.parametrize(
-    "snapshots",
+    "snapshots, snapshots_to_retain",
     (
-        # Less than 7
-        [_make_snapshot(1)],
+        # Less than the number of snapshots to retain
+        ([_make_snapshot(1)], 2),
+        ([_make_snapshot(1)], 7),
         # Exactly the number we want to keep
-        [_make_snapshot(i) for i in range(7)],
+        ([_make_snapshot(i) for i in range(2)], 2),
+        ([_make_snapshot(i) for i in range(7)], 7),
     ),
 )
-def test_delete_previous_snapshots_no_snapshots_to_delete(snapshots, rds_client):
+def test_delete_previous_snapshots_no_snapshots_to_delete(
+    snapshots, snapshots_to_retain, rds_client
+):
     rds_client.describe_db_snapshots.return_value = snapshots
-    delete_previous_snapshots.function()
+    delete_previous_snapshots.function("fake_arn", snapshots_to_retain)
     rds_client.delete_db_snapshot.assert_not_called()
 
 
@@ -39,7 +43,7 @@ def test_delete_previous_snapshots(rds_client):
     snapshots = [_make_snapshot(i) for i in range(10)]
     snapshots_to_delete = snapshots[7:]
     rds_client.describe_db_snapshots.return_value = snapshots
-    delete_previous_snapshots.function()
+    delete_previous_snapshots.function("fake_arn", 7)
     rds_client.delete_db_snapshot.assert_has_calls(
         [
             mock.call(DBSnapshotIdentifier=snapshot["DBSnapshotIdentifier"])
```

Additionally it looks like `just generate-dag-docs` needs to be re-run as well.

AetherUnbound (approved):
Code looks great, although the DAG documentation check step is still failing. Once that's addressed this is good to merge 👍🏼

openverse_catalog/dags/maintenance/rotate_db_snapshots.py
```diff
@@ -0,0 +1,92 @@
...
+def delete_previous_snapshots(rds_arn: str, snapshots_to_retain: int):
+    rds = boto3.client("rds")
+    # Snapshot object documentation:
+    # https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DBSnapshot.html
+    snapshots = rds.describe_db_snapshots(
+        DBInstanceIdentifier=rds_arn,
+    )
+
+    if len(snapshots) <= snapshots_to_retain or not (
+        snapshots_to_delete := snapshots[snapshots_to_retain:]
```
stacimc, line 41: Can we rely on the snapshots being in order such that the oldest get deleted here? It wasn't clear to me from a quick look at the documentation, but it seems like the Sample Response given [here](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBSnapshots.html) is not sorted, unless I'm looking at it wrong.

sarayourfriend: I don't know why the sample response is not ordered but the actual API response I get when I run `aws rds describe-db-snapshots --db-instance-identifier=upstream-openverse-db` is ordered by date in the way the code expects (most recent first).

stacimc: Very strange. Looking around I see a couple [StackOverflow posts](https://stackoverflow.com/questions/69558838/how-can-i-get-the-name-of-the-most-recent-snapshot-for-an-rds-db-instance-using) that suggest sorting the results. I would certainly _expect_ the snapshots to be ordered, but if they aren't we'd risk failing to retain recent backups.

sarayourfriend: I'll add a sort, I'm assuming that's the change you'd recommend, but let me know if there's something else you'd like to see instead.

stacimc: Thanks. Given the documentation shows unsorted responses, and the potential gravity of losing backups, that makes me more comfortable.

stacimc (approved):
LGTM, excited to see this in Airflow 😄

PR #914, opened 2022-12-09
Break load_from_s3 into separate tasks to fix duplicate reporting

### LINKED ISSUE #800: Incorrect duplicate values reported for Smithsonian
## Description

Our most recent Smithsonian run produced some seemingly-incorrect reporting results:

image

For maintainers, the relevant Airflow log link is here: https://airflow.openverse.engineering/log?dag_id=smithsonian_workflow&task_id=ingest_data.load_image_data.load_from_s3&execution_date=2022-10-02T00%3A00%3A00%2B00%3A00

## Resolution

stacimc:
## Fixes

Fixes #800 by @AetherUnbound 

## Description

We've noticed some very strange duplicate reporting from the Smithsonian provider on a few DagRuns:

Screen Shot 2022-12-08 at 5 15 19 PM

The bug is caused when the `load_from_s3` step retries (see investigation described in my comment [here](https://github.com/WordPress/openverse-catalog/issues/800#issuecomment-1343610886)). Essentially, at the moment the `load_from_s3` step does three things:

1. Loads all of the pulled data from s3 into a temporary `load_table`
2. Runs some cleanup steps on the data, which among other things **deletes rows that have duplicated foreign ids**
3. Upserts the remaining data to the catalog DB

The bug happens when the `load_from_s3` step successfully does step 1 & 2, but then errors or times out during the upsert. When the task retries, _all three steps_ are performed again. That means all the pulled data is loaded into the _already populated `load_table`_, and will all be identified as foreign id duplicates in step 2. Consequently misleading numbers of duplicates are reported.

This PR solves the problem by separating the `load_from_s3` task out into three individual tasks that perform each of the above steps (`load_from_s3`, `clean_data`, and `upsert_data`). If an error or timeout occurs during a task, only that specific step is re-run.

Screen Shot 2022-12-09 at 2 24 51 PM

## Notes
* The `load_timeout` configuration option has been renamed to `upsert_timeout` for clarity, and is only applied to the `upsert_data` step. Loading from s3 and cleaning data should both be very quick.
* Since it's clear the Smithsonian upsert step was timing out occasionally (leading to the original bug), I increased the timeout from 4 hours to 6
* Initially I only split the task into two tasks, a `load_from_s3` step and a combined cleaning/upsertion step. The problem with this approach is that a similar bug can happen if the clean/upsert task retries after timing out during upsertion. This time, when the cleaning step runs, it will detect **no** duplicates because they were already deleted on the previous try.

## Testing Instructions

`just test`

Try running a few DAGs locally and ensure that the reporting looks correct. You may want to use the `INGESTION_LIMIT` Airflow variable to speed the process up. I tested with:
* Jamendo (audio provider)
* Smithsonian (image provider)
* Wikimedia (dated provider, image and audio)
* Wikimedia Reingestion (reingestion workflow)

Bonus testing points: I then pulled `main` and re-ran my test DAGs to confirm that the numbers reported were the same :)

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/loader/loader.py
+++ b/openverse_catalog/dags/common/loader/loader.py
@@ -34,37 +34,19 @@ def load_s3_data(
     )
 
 
-def load_from_s3(
-    bucket: str,
-    key: str,
-    postgres_conn_id: str,
-    media_type: str,
-    identifier: str,
-) -> int:
-    """
-    Loads data from S3 into the intermediary loading table.
-
-    Returns the int number of rows that were loaded.
-    """
-    return sql.load_s3_data_to_intermediate_table(
-        postgres_conn_id, bucket, key, identifier, media_type
-    )
-
-
 def upsert_data(
     postgres_conn_id: str,
     media_type: str,
     tsv_version: str,
     identifier: str,
     loaded_count: int,
+    duplicates_count: tuple[int, int],
 ) -> RecordMetrics:
     """
-    Cleans data in the intermediary loading table, and then upserts it into
-    the Catalog DB.
+    Upserts data into the catalog DB from the loading table, and calculates
+    final record metrics.
     """
-    missing_columns, foreign_id_dup = sql.clean_intermediate_table_data(
-        postgres_conn_id, identifier, media_type
-    )
+    missing_columns, foreign_id_dup = duplicates_count
     upserted = sql.upsert_records_to_db_table(
         postgres_conn_id, identifier, media_type=media_type, tsv_version=tsv_version
     )
--- a/openverse_catalog/dags/common/loader/loader.py
+++ b/openverse_catalog/dags/common/loader/loader.py
@@ -35,18 +35,39 @@ def load_s3_data(
 
 
 def load_from_s3(
-    bucket,
-    key,
-    postgres_conn_id,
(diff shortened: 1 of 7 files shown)


openverse_catalog/dags/common/loader/loader.py
```diff
@@ -34,19 +34,22 @@ def load_s3_data(
     )
```
stacimc, line 36: Note for reviewers: in this file I deleted the `load_from_s3` method in favor of calling directly through to the `sql.load_s3_data_to_intermediate_table`, since it no longer has any additional logic.

Then I added the new `upsert_data` method.

krysal (approved):
Love that each task is separated and clearly well-defined now! LGTM 🚀

stacimc:
> Edit: I forgot to mention that I ran the Smithsonian DAG and I got the error of a missing provider the first time:

@krysal Wow, good catch! Looks like a subprovider got added while the DAG was otherwise disabled due to this error 😅  I went ahead and added it in this PR and confirmed it's working for me now locally as well.

PR #916, opened 2022-12-10
Halt ingestion when WordPress Photo Directory reaches last page

### LINKED ISSUE #853: WordPress Photo Directory provider script does not stop
## Description

After the refactor performed in #835, the WordPress Photo Directory provider script no longer stops attempting to process data after hitting the last page. A recent run with `skip_ingestion_errors` (which was run as a result of #852) caught the script in an infinite loop which kept it iterating over pages which did not exist. We ended up forcing the run to fail and had a 16MB log file with the following lines repeated thousands of times:

```
[2022-11-01T16:18:32.116+0000] {requester.py:75} WARNING - Unable to request URL: https://wordpress.org/photos/wp-json/wp/v2/photos?format=json&page=50&per_page=100&_embed=true  Status code: 400
[2022-11-01T16:18:32.116+0000] {requester.py:121} WARNING - Bad response_json:  None
[2022-11-01T16:18:32.116+0000] {requester.py:122} WARNING - Retrying:
_get_response_json(
    https://wordpress.org/photos/wp-json/wp/v2/photos,
    {'format': 'json', 'page': 50, 'per_page': 100, '_embed': 'true'},
    retries=2)
```

The response from that URL is:
```json
{
    "code": "rest_post_invalid_page_number",
    "message": "The page number requested is larger than the number of pages available.",
    "data": {
        "status": 400
    }
}
```

It looks like there was some logic for calculating total available pages which were not carried over from the refactor:

https://github.com/WordPress/openverse-catalog/pull/835/files#diff-b98978a812fb92894554fca75f318f598e0135bfda7f2d2c8397f7ae5c5d8fe8L83

## Reproduction

1. Trigger the DAG with the config: `{"init_query_params": {'format': 'json', 'page': 49, 'per_page': 100, '_embed': 'true'}}`
2. Observe that after one page, the DAG run fails on the above error

## Resolution

stacimc:
## Fixes

Fixes #853 by @AetherUnbound 

## Description

The WordPress Photo Directory DAG was failing to halt ingestion gracefully when it reached the end of its data. Instead, it raises a 400 when attempting to access a page number beyond the total page count. This is especially bad if `skip_ingestion_errors` is enabled, because it will retry infinitely.

To fix this we need to make a HEAD request to get the total page count. This information is only available in the headers and not `response_json`, so we are not able to fetch it in `get_batch_data`. This PR makes this additional request on the first batch, and then halts ingestion during `get_should_continue` when we reach the final page.

## Testing Instructions

At the time this PR was written, this DAG detects 53 pages of results. To test it locally, I recommend editing the initial `current_page` variable in the `__init__` method to `52`, so the DAG will run quickly and hit the last page:
```
def __init__(self, *args, **kwargs):
        <...>

        self.current_page = 52
```

After making the edit, run the DAG and verify that it stops gracefully when the last page has been reached. You should be able to see a log that looks like:

```
The final page of data has been processed. Halting ingestion.
```

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/wordpress.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/wordpress.py
@@ -62,6 +62,7 @@ class WordPressDataIngester(ProviderDataIngester):
                 self.endpoint, params={"per_page": self.batch_limit, "_embed": "true"}
             )
             self.total_pages = int(response.headers.get("X-WP-TotalPages", 0))
+            logger.info(f"{self.total_pages} pages detected.")
 
         return {
             "format": "json",
--- a/openverse_catalog/dags/providers/provider_api_scripts/wordpress.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/wordpress.py
@@ -47,7 +47,7 @@ class WordPressDataIngester(ProviderDataIngester):
 
         # Total pages is determined on the first request
         self.total_pages = None
-        self.current_page = 0
+        self.current_page = 1
 
     def get_media_type(self, record: dict) -> str:
         return constants.IMAGE
@@ -64,8 +64,9 @@ class WordPressDataIngester(ProviderDataIngester):
             self.total_pages = int(response.headers.get("X-WP-TotalPages", 0))
             logger.info(f"{self.total_pages} pages detected.")
 
-        # Increment the page number for the next batch
-        self.current_page += 1
+        # Increment the page number for subsequent batches
+        if prev_query_params:
+            self.current_page = prev_query_params["page"] + 1
 
         return {
             "format": "json",
--- a/openverse_catalog/dags/providers/provider_api_scripts/wordpress.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/wordpress.py
@@ -47,7 +47,7 @@ class WordPressDataIngester(ProviderDataIngester):
 
         # Total pages is determined on the first request
         self.total_pages = None
-        self.current_page = 1
+        self.current_page = 0
 
     def get_media_type(self, record: dict) -> str:
         return constants.IMAGE
@@ -64,6 +64,9 @@ class WordPressDataIngester(ProviderDataIngester):
             self.total_pages = int(response.headers.get("X-WP-TotalPages", 0))
             logger.info(f"{self.total_pages} pages detected.")
 
+        # Increment the page number for the next batch
+        self.current_page += 1
+
         return {
             "format": "json",
             "page": self.current_page,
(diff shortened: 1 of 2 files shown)


sarayourfriend (approved):
This LGTM and tests fine locally. Thank you for the good instructions.

I was wondering if you had considered any alternative solutions to this, perhaps ones that would add the ability to get the headers to the base class? I had one idea for how to do it so that it would always be available as part of `response_json` but didn't want to get into it if you thought it was a YAGNI type situation.

Anyway, LGTM either way and the solution makes sense for this DAG.

openverse_catalog/dags/providers/provider_api_scripts/wordpress.py
```diff
@@ -45,25 +45,48 @@ def __init__(self, *args, **kwargs):
...
+        }
 
     def get_batch_data(self, response_json):
         if isinstance(response_json, list) and len(response_json):
             return response_json
         return None
 
+    def get_should_continue(self, response_json):
+        # Increment the page number for the next batch
+        self.current_page += 1
```
krysal, line 81: Interesting that you decided to increment the page rather than in the `get_next_query_params`method.

krysal (approved):
Nice! I was thinking about the general solution for the response headers as well but that seemed to require much refactorization of other DAGs for just using it here. I like this.

Good code coverage in testing too! It's appreciated.

openverse_catalog/dags/providers/provider_api_scripts/wordpress.py
AetherUnbound: +1, it seems like this should be in `get_next_query_params` (if possible) to match the other patterns we have!

sarayourfriend:
I have at least one idea that wouldn't require any refactors of other DAGs, but would be a little obscure: subclass `dict` and add a `headers` attribute. Then map the JSON response into the dict subclass and add the headers to the headers attribute.

openverse_catalog/dags/providers/provider_api_scripts/wordpress.py
stacimc: Fair enough, done! Excellent call out: I believe we actually _need_ to increment there, so that we can increment based off the previous page (instead of naively adding +1), or else we end up with a nasty edge case when `initial_query_params` are supplied. Thanks 😄

stacimc:
> I was wondering if you had considered any alternative solutions to this, perhaps ones that would add the ability to get the headers to the base class?

We definitely _can_, but I wasn't aware of any other cases where we needed it (although I could be wrong). I'd say I have a slight preference to refactoring more fully if/when another use case arises... although the dict subclassing idea is pretty slick @sarayourfriend 🤩

sarayourfriend:
> I'd say I have a slight preference to refactoring more fully if/when another use case arises

I like this as well :rocket:

Issue #754, opened 2022-09-30
Use `HEAD` request in Freesound ingester for file size request


AetherUnbound:
## Description

Dependent on #753 and #746

We should replace the `requests.head` call with the `DelayedRequester::head` call described in #753:

https://github.com/WordPress/openverse-catalog/blob/c332465db5dfdff41f9f89cd6792c01732a570be/openverse_catalog/dags/providers/provider_api_scripts/freesound.py#L172

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

AetherUnbound:
I've bumped this to high because we're now seeing failures of the DAG occurring at this request. Specifically, if the URL 404s, this raises a `KeyError` for `content-length` ([example Airflow logs](https://airflow.openverse.engineering/log?execution_date=2022-11-01T00%3A00%3A00%2B00%3A00&task_id=ingest_data.pull_audio_data&dag_id=freesound_workflow&map_index=-1)). Using the machinery of the `DelayedRequester::head` method might help us address this.

AetherUnbound:
Here's the failing sound in question: https://freesound.org/people/dl-jones/sounds/653078/

It looks like missing preview for a sound is indeed an issue with the Freesound data that we'll just need to account for. In these cases would we just want to skip the record entirely, since we can't preview it on our end? CC @WordPress/openverse-catalog

zackkrida:
> In these cases would we just want to skip the record entirely, since we can't preview it on our end?

This is an interesting question, and I think it's okay to skip these records. The audio player is broken in FreeSound's frontend as well, which is a terrible user experience.

AetherUnbound:
It's odd though, I just logged in and you _can_ download the audio file itself 🤔

AetherUnbound:
I'm setting this priority to low since #920 is the issue that's actually affecting the DAG's ability to run successfully.

PR #928, opened 2022-12-20
Handle Freesound 404s when fetching audio sets

### LINKED ISSUE #659: Freesound fails fetching audio_set
## Description

We've seen some recent Freesound failures of the nature:
```
[2022-08-01, 10:21:46 UTC] {requester.py:59} WARNING - Unable to request URL: https://freesound.org/apiv2/packs/26413/?token=***.  Status code: 404
[2022-08-01, 10:21:46 UTC] {requester.py:99} WARNING - Bad response_json:  None
[2022-08-01, 10:21:46 UTC] {requester.py:100} WARNING - Retrying:
_get_response_json(
    https://freesound.org/apiv2/packs/26413/,
    {'token': '***'},
    retries=-1)
[2022-08-01, 10:21:46 UTC] {requester.py:85} ERROR - No retries remaining.  Failure.
```

The error appears to happen while fetching an [audio_set](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_api_scripts/freesound.py#L201). The request already has 3 retries. We should investigate the root cause and add handling.

## Additional context

This may be related to/causing #658.

## Resolution

stacimc:
## Fixes

Fixes #659

## Background and Investigation

Sometimes Freesound returns a 404 when trying to fetch audio_set information. In a production run with `skip_ingestion_errors` enabled, we encountered this problem only for a single audio set (https://freesound.org/apiv2/packs/35596/). However since the task timed out (and some batches are skipped), it's possible the error would have been encountered additional times in a truly full run.

## Description

This PR:

* moves the token from query params into the headers
* skips errors when trying to fetch an audio_set

When an error happens fetching audio set info, I opted to **return `None` for the set id & name and just continue**. Another option would be to discard the record.

## Notes

* We already have three retries when fetching audio_set information. Since we've had trouble with flakiness, I've left the retries in and only swallow the error when retries are exhausted. I do not think this will lead to many useless retries as the 404 is only happening for a single audio set in testing.
* `DelayedRequester.get_response_json` does not make it easy to add detection/handling for the 404 specifically. We _could_ use `self.delayed_requester.get` directly, but would then lose all of the other handling already implemented in `get_response_json`. 

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/DAGs.md
+++ b/DAGs.md
@@ -261,9 +261,8 @@ ETL Process: Use the API to identify all CC-licensed images.
 
 Output: TSV file containing the image, the respective meta-data.
 
-Notes: https://freesound.org/docs/api/ Rate limit: 60 requests per minute, 2000
-requests per day. This script can be run either to ingest the full dataset or as
-a dated DAG.
+Notes: https://freesound.org/docs/api/ Rate limit: No limit for our API key.
+This script can be run either to ingest the full dataset or as a dated DAG.
 
 ## `image_data_refresh`
 
--- a/DAGs.md
+++ b/DAGs.md
@@ -261,8 +261,9 @@ ETL Process: Use the API to identify all CC-licensed images.
 
 Output: TSV file containing the image, the respective meta-data.
 
-Notes: https://freesound.org/apiv2/search/text' No rate limit specified. This
-script can be run either to ingest the full dataset or as a dated DAG.
+Notes: https://freesound.org/docs/api/ Rate limit: 60 requests per minute, 2000
+requests per day. This script can be run either to ingest the full dataset or as
+a dated DAG.
 
 ## `image_data_refresh`
 
--- a/openverse_catalog/dags/providers/provider_api_scripts/freesound.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/freesound.py
@@ -7,7 +7,7 @@ Output:                 TSV file containing the image, the respective
                         meta-data.
 
 Notes:                  https://freesound.org/docs/api/
-                        Rate limit: 60 requests per minute, 2000 requests per day.
+                        Rate limit: No limit for our API key.
                         This script can be run either to ingest the full dataset or
                         as a dated DAG.
 """
--- a/openverse_catalog/dags/providers/provider_api_scripts/freesound.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/freesound.py
@@ -42,13 +42,13 @@ class FreesoundDataIngester(ProviderDataIngester):
         "preview-lq-ogg": 80000,
     }
 
-    headers = {
-        "Accept": "application/json",
-    }
-
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.api_key = Variable.get("API_KEY_FREESOUND")
+        self.headers = {
+            "Accept": "application/json",
+            "Authorization": f"Token {self.api_key}",
+        }
 
     def get_media_type(self, record: dict) -> str:
         return constants.AUDIO
@@ -64,7 +64,6 @@ class FreesoundDataIngester(ProviderDataIngester):
                 )
             return {
                 "format": "json",
-                "token": self.api_key,
                 "query": "",
                 "page_size": self.batch_limit,
                 "fields": ",".join(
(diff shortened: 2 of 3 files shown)


AetherUnbound (approved):
Looks good to me! Is there _any_ information we can pull out of the caught `RetriesExceeded` exception that we could use to check that it's a 404 error? Not blocking for this PR by any means, but it would be cool if we could restrict the error cases we're handling.

stacimc:
> Is there any information we can pull out of the caught RetriesExceeded exception that we could use to check that it's a 404 error? Not blocking for this PR by any means, but it would be cool if we could restrict the error cases we're handling.

Totally agree 😞 Unfortunately there's currently [no information being passed back](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/requester.py#L135) at all in the RetriesExceeded exception. We could maybe update the DelayedRequester to include some more information in the error message but since there's a lot of handling there I'm not immediately sure what to include. `status_code` and any `error` in the json if present? What if it's a JSONDecodeError?

AetherUnbound:
> Totally agree disappointed Unfortunately there's currently [no information being passed back](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/requester.py#L135) at all in the RetriesExceeded exception. We could maybe update the DelayedRequester to include some more information in the error message but since there's a lot of handling there I'm not immediately sure what to include. `status_code` and any `error` in the json if present? What if it's a JSONDecodeError?

Those are great questions! I think this ties a lot into https://github.com/WordPress/openverse-catalog/issues/856, I wonder how we could leverage [exception chaining](https://peps.python.org/pep-0344/) to help provide us with more information _when there is an exception_, and then add some other details in cases the request failed 🤔

DAGs.md
```diff
@@ -261,8 +261,9 @@ ETL Process: Use the API to identify all CC-licensed images.
 
 Output: TSV file containing the image, the respective meta-data.
 
-Notes: https://freesound.org/apiv2/search/text' No rate limit specified. This
-script can be run either to ingest the full dataset or as a dated DAG.
+Notes: https://freesound.org/docs/api/ Rate limit: 60 requests per minute, 2000
+requests per day. This script can be run either to ingest the full dataset or as
+a dated DAG.
```
krysal, line 266: This should not apply to us as Frederic Font from Freesound granted us unlimited permissions on our API key. Do these refer to a technical limitation? Was this discussed with them?

stacimc: No, this was just taken directly from the linked API docs as I did not see it documented anywhere that we had been granted additional permissions. Thanks for noting that, I will update the docstring here! That's good to know, as I was concerned we might be hitting the limit 😅

openverse_catalog/dags/providers/provider_api_scripts/freesound.py
```diff
@@ -42,13 +43,13 @@ class FreesoundDataIngester(ProviderDataIngester):
...
-        "Accept": "application/json",
-    }
-
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.api_key = Variable.get("API_KEY_FREESOUND")
+        self.headers = {
+            "Accept": "application/json",
+            "Authorization": f"Token {self.api_key}",
+        }
```
krysal, line 52: Doesn't the call to the parent constructor has to be after setting the headers for `DelayedRequester` to be configured correctly?

stacimc: 😮 Good point! This currently works as is because `get_response_json` also explicitly passes `self.headers` in [here](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py#L353). But it would certainly cause a problem if we ever used `self.delayed_requester` directly in the child class, which we actually have plans to do in https://github.com/WordPress/openverse-catalog/issues/754! And it's good practice, besides.

Done, thanks 😄

krysal: Perfect!

krysal (approved):
🚀

PR #934, opened 2022-12-23
Finnish DAG: Dynamically generate timeslices depending on the amount of records

### LINKED ISSUE #913: Conditionally break down Finnish ingestion into increments
## Problem

#879 makes the Finnish provider DAG break ingestion up into half-hour increments for the ingestion date in order to circumvent buggy behavior with large batches.

Because the Finnish provider script is very sparse, most ingestion days will return no data and it's unnecessary and inefficient to split up the requests like this. For small batches it also leads to unnecessarily cluttered logs.

## Description

When ingestion first starts, request the records for the **entire day** first and check the `resultCount`. Only split ingestion up into half-hour increments if the `resultCount > 0`.

## Alternatives

A small number of ingestion days may have a nonzero but still small amount of data. We could consider identifying a safe threshold below which we don't split up ingestion, instead of `0`. (For example I have personally observed a day with ~500 records passed ingestion without encountering any bugs). This isn't strictly necessary, though.

## Additional context

See comment [here](https://github.com/WordPress/openverse-catalog/pull/879#discussion_r1036648573)

## Implementation

stacimc:
## Fixes

Fixes #913 

#913 is low priority because it just tracks optimizing the DAG for days with small amounts of data. Since the DAG is now also raising errors for days with very large amounts (see 12-16-22), I'm increasing the PR to high priority.

## Description

In #879 I split the ingestion of data for the Finnish DAG into 48 half-hour increments over the ingestion day. This attempts to avoid some buggy behavior we're seeing with the Finnish API where, if the query is sufficiently large, it will keep returning pages of duplicate data indefinitely and never halt ingestion.

This has a few problems:
* Most days, the DAG returns 0 records but we still make 48 requests _per building_.
* Even on days with large amounts of data, the data is rarely distributed evenly across the day. In my tests I saw that the vast majority of the data was usually confined to one or two hours. So, much of the day is split into unnecessarily small intervals.
* On days with very large amounts of data, it looks like my half-hour intervals were not small enough, and we still see the bug. (See Dec 30 2022 run in production).

This PR:
* handles all of the above problems by taking the record count into account when generating timeslices (described in the section below)
* since it's possible my new slices are still not small enough, it also adds in detection for the bug (when the fetched count exceeds the reported result count). It will fail the task early.

### How the timeslices are generated

The code itself is heavily documented, but to reiterate for convenience:

* If there are no records for the day (true for the majority of runs), it returns immediately and does not ingest at all
* If there are fewer than 10,000 records, it does not split the day up at all and runs ingestion in one large chunk
* If there are more than 10,000 records for the day, it gets the recordCount for _each hour_ and only adds timeslices for hours that contain data.
  * hours with < 10k records are processed in a single slice
  * hours with < 100k records are processed in 12 slices (5 min intervals)
  * hours with > 100k records are processed in 20 slices (3 min intervals)
 
The result is **one extra request** for the majority of ingestion days, when there is only a small amount of data, and **twenty five extra requests (to get the record counts)** for days with very large amounts of data.

At the very most this could result in 20*24 = 480 time slices for a particular building, but this would only happen for an ingestion day with over 100k records _every single hour_. Based on my observations I expect to see far, far fewer.

A quick comparison for requests/slices from the recent failed 12-30-22 run:
* On main, we did 48 slices * 4 buildings = **192 total "slices"**
* On this branch, 1 building was processed in a single slice, 2 had no data and halted early, and 1 had >100k records in a single hour. That means all four buildings will processed in just **21 slices**, with only 28 extra requests made to generate those slices effectively (1 for each of the 3 with little/no data, 25 for the large one).

## Potential Future Work

I'm going to be looking at whether some of this logic can be pulled out of Finnish and into either a utility class or the base class itself, as we have similar logic needed in multiple providers. In particular I'm going to test this slice generation against Flickr 🤔

## Testing Instructions

```just test```, and read through the new tests to make sure they make sense.

Also try running the Finnish DAG locally and make sure it works.

If you want to reproduce the bug, 12-30-2022 is an example of a DagRun that failed in production even with the 48-slice approach on main. Locally, I verified that it now passes on this branch by triggering a DagRun with the config:

```
{"date": "2022-12-30"}
```

It passed in 3 hrs 39 minutes. 

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
@@ -19,7 +19,6 @@ from datetime import datetime, timedelta, timezone
 from itertools import chain
 
 from airflow.exceptions import AirflowException
-from common import slack
 from common.licenses import get_license_info
 from common.loader import provider_details as prov
 from providers.provider_api_scripts.provider_data_ingester import ProviderDataIngester
@@ -240,12 +239,10 @@ class FinnishMuseumsDataIngester(ProviderDataIngester):
         # Detect a bug when the API continues serving pages of data in excess of
         # the stated resultCount.
         if self.fetched_count > total_count:
-            slack.send_alert(
+            raise AirflowException(
                 f"Expected {total_count} records, but {self.fetched_count} have"
-                " been fetched. Halting ingestion. Consider reducing the ingestion"
-                " interval."
+                " been fetched. Consider reducing the ingestion interval."
             )
-            raise AirflowException
 
         return response_json["records"]
 
--- a/openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
@@ -18,6 +18,7 @@ import logging
 from datetime import datetime, timedelta, timezone
 from itertools import chain
 
+from airflow.exceptions import AirflowException
 from common import slack
 from common.licenses import get_license_info
 from common.loader import provider_details as prov
@@ -244,9 +245,7 @@ class FinnishMuseumsDataIngester(ProviderDataIngester):
                 " been fetched. Halting ingestion. Consider reducing the ingestion"
                 " interval."
             )
-            # Returning None will halt ingestion for this iteration, and the DAG will
-            # continue on to the next building or time interval.
-            return None
+            raise AirflowException
 
         return response_json["records"]
 
--- a/openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
@@ -94,8 +94,12 @@ class FinnishMuseumsDataIngester(ProviderDataIngester):
         """
         seconds_in_time_slice = (end_date - start_date).total_seconds()
         portion = int(seconds_in_time_slice / number_of_divisions)
-        # TODO: should there be an error if the division ^ does not result in even
-        # portions? Definitely if we make this configurable by dag conf.
+        # Double check that the division resulted in even portions
+        if portion != seconds_in_time_slice / number_of_divisions:
+            raise ValueError(
+                f"The time slice from {start_date} to {end_date} cannot be divided "
+                f"evenly into {number_of_divisions} !"
+            )
 
         # Generate the start/end timestamps for each 'slice' of the interval
         return [
@@ -233,7 +237,9 @@ class FinnishMuseumsDataIngester(ProviderDataIngester):
             return None
 
         total_count = response_json.get("resultCount")
-        # Update the number of records we have pulled for this iteration
+        # Update the number of records we have pulled for this iteration.
+        # Note that this tracks the number of records pulled from the API, not
+        # the number actually written to TSV.
         self.ingested_count += len(response_json.get("records"))
 
         if self.new_iteration:
--- a/openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
@@ -62,9 +62,9 @@ class FinnishMuseumsDataIngester(ProviderDataIngester):
         # A flag that is True only when we are processing the first batch of data in
         # a new iteration.
         self.new_iteration = True
-        # Use to keep track of the number of records we've ingested so far, specifically
-        # in this iteration.
-        self.ingested_count = 0
+        # Use to keep track of the number of records we've fetched from the API so far,
+        # specifically in this iteration.
+        self.fetched_count = 0
 
         super().__init__(*args, **kwargs)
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py [hunk 1]
```diff
@@ -42,54 +48,155 @@ class FinnishMuseumsDataIngester(ProviderDataIngester):
...
+                logger.info(f"No data detected for {start_hour}. Continuing.")
+                continue
+            if record_count < 10_000:
+                # This hour doesn't have much data, ingest it in one chunk
+                pairs_list.append((start_hour, end_hour))
+                continue
+            # If we got this far, this hour has a lot of data. It is split into 12 5-min
+            # intervals if it has fewer than 100k, or 20 3-min intervals if more.
+            num_divisions = 12 if record_count < 100_000 else 20
+            min_slices = self._get_timestamp_query_params_list(
```
stacimc, line 180: reminder to myself to change the name `min_slices`, currently looks like 'minimum'

tests/dags/providers/provider_api_scripts/test_finnish_museums.py
```diff
@@ -30,73 +36,230 @@ def _get_resource_json(json_name):
...
+            # Now try to ingest records
+            fm.ingest_records()
+
+            # get_mock should have been called 4 times
+            assert get_mock.call_count == 4
+            # process_mock should only have been called twice. The second 'batch' for the first time slice
+            # should have returned early
+            assert process_mock.call_count == 2
+            # Slack should have been alerted to the issue
+            assert slack_alert_mock.called is True
```
stacimc, line 248: note to self:  test that get_response_json was called specifically twice for the first timeslice and twice for the second, by checking the query params it was called with.

tests/dags/providers/provider_api_scripts/test_finnish_museums.py
```diff
@@ -29,74 +36,207 @@ def _get_resource_json(json_name):
...
+        # Mock the calls to _get_record_count in order
+        mock_count.side_effect = [
+            150_000,  # Getting total count for the entire day
+            0,  # Get count for first hour, count == 0
+            10,  # Get count for second hour, count < 10_000
+            101_000,  # Get count for third hour, count > 100_000
+            49_090,  # Get count for fourth hour, 10_000 < count < 100_000
+        ] + list(
+            repeat(0, 20)
+        )  # Fill list with count == 0 for the remaining hours
```
AetherUnbound, line 133: Nice! This rocks 🚀

AetherUnbound (approved):
This is fantastic! Even though there is a lot of heuristics built in here, the solution is simpler than I though. I'm really glad the API supplies us with a result count 😌 I was able to run the tests successfully locally! I have a few nits but nothing to block a merge.

openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
```diff
@@ -42,54 +42,153 @@ class FinnishMuseumsDataIngester(ProviderDataIngester):
...
-        seconds_in_a_day = 86400
-        number_of_divisions = 48  # Half-hour increments
-        portion = int(seconds_in_a_day / number_of_divisions)
-
-        def _format_timestamp(d):
-            return d.isoformat().replace("+00:00", "Z")
+        seconds_in_time_slice = (end_date - start_date).total_seconds()
+        portion = int(seconds_in_time_slice / number_of_divisions)
+        # Double check that the division resulted in even portions
+        if portion != seconds_in_time_slice / number_of_divisions:
```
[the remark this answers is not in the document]
stacimc: For sure!

krysal (approved):
Awesome solution. Thank you for including so many details in comments. Looks great to me!

openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
```diff
@@ -42,54 +42,153 @@ class FinnishMuseumsDataIngester(ProviderDataIngester):
...
+
+        For logging and alerting purposes, we maintain several instance variables
+        that help track each iteration.
+        """
+        # A flag that is True only when we are processing the first batch of data in
+        # a new iteration.
         self.new_iteration = True
+        # Use to keep track of the number of records we've fetched from the API so far,
+        # specifically in this iteration.
+        self.fetched_count = 0
```
krysal, line 68: Pycharm suggests adding `timestamp_pairs` here in the `__init__()`, given it's used later in the `ingest_records` method.
```suggestion
        self.fetched_count = 0
	self.timestamp_pairs = None
	
```

krysal (commented):
Aside from #975, I also noticed the images are watermarked with what seems to be the name of the building they belong to, maybe this is the reason for the`watermarked` column's existence? Interesting finding!

openverse_catalog/dags/providers/provider_api_scripts/finnish_museums.py
stacimc: Thank you! Now you point it out, `timestamp_pairs` can actually just be a local variable to `ingest_records`.

PR #973, opened 2023-01-25
Discard audio file when preview 404s fetching filesize

### LINKED ISSUE #754: Use `HEAD` request in Freesound ingester for file size request

### LINKED ISSUE #920: Freesound preview for valid file could 404
## Description

It appears that the preview URLs which Freesound provides may 404 even for valid URLs. See the full context in #754.

We'll want to handle the missing preview files gracefully. Since lacking a preview means the sound can't be played in the frontend (or even on the provider's website), we've decided that we should just **skip these records entirely** if the preview link 404s.

stacimc:
## Fixes

Fixes #920 by @AetherUnbound 
Fixes #754 by @AetherUnbound 

## Description

* Updates Freesound to use the new `DelayedRequester::head` call
* Adds handling for an edge case when we get a 404 on a preview URL, which we use for getting filesize information. Since getting a 404 on the preview means the record is not playable on our frontend, we return `None` early and skip ingestion for these records.

## Testing Instructions

Check out the new unit tests.

Run the Freesound DAG locally and verify that it still works.

One way to force a 404 for testing is to update the `_get_audio_file_size` method to append some nonsense to the end of the URL before making the request:
```
@retry(flaky_exceptions, tries=3, delay=1, backoff=2)
    def _get_audio_file_size(self, url):
        # Force a 404 by breaking the url 
        url = url + 'foo'

        response = self.delayed_requester.head(url)
        ...
```
Then re-run the DAG. You should see a lot of logs that look like this:
```
{requester.py:78} WARNING - Unable to request URL: https://cdn.freesound.org/previews/664/664719_14565529-hq.mp3foo  Status code: 404
```
Manually `Mark Success` the `pull_data` step once you have verified these logs, or else it will try and fail for all records. Then verify in `report_load_completion` that no records were actually ingested. 

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/freesound.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/freesound.py
@@ -15,7 +15,6 @@ import functools
 import logging
 from datetime import datetime
 
-import requests
 from airflow.models import Variable
 from common import constants
 from common.licenses.licenses import get_license_info
@@ -179,7 +178,11 @@ class FreesoundDataIngester(ProviderDataIngester):
 
         TODO(obulat): move filesize detection to the polite crawler
         """
-        return int(requests.head(url).headers["content-length"])
+        response = self.delayed_requester.head(url)
+
+        if response:
+            return response.headers.get("content-length")
+        return None
 
     def _get_audio_files(
         self, media_data
@@ -193,11 +196,16 @@ class FreesoundDataIngester(ProviderDataIngester):
         if not (preview_url := previews.get(self.preferred_preview)):
             return None, None
 
+        # If unable to get filesize from the preview, skip this audio
+        # This may happen if the preview 404s
+        if (filesize := self._get_audio_file_size(preview_url)) is None:
+            return None, None
+
         main_file = {
             "audio_url": preview_url,
             "filetype": self.preferred_preview.split("-")[-1],
             "bit_rate": FreesoundDataIngester.preview_bitrates[self.preferred_preview],
-            "filesize": self._get_audio_file_size(preview_url),
+            "filesize": int(filesize),
         }
         # These are the original files, needs auth for downloading.
         # bit_rate in kilobytes, converted to bytes
--- a/openverse_catalog/dags/providers/provider_api_scripts/freesound.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/freesound.py
@@ -175,8 +175,6 @@ class FreesoundDataIngester(ProviderDataIngester):
         Both of these seem transient and may be the result of some odd behavior on the
         Freesound API end. We have an API key that's supposed to be maxed out, so
         I can't imagine it's throttling (aetherunbound).
-
-        TODO(obulat): move filesize detection to the polite crawler
         """
         response = self.delayed_requester.head(url)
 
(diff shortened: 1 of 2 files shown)


openverse_catalog/dags/providers/provider_api_scripts/freesound.py
```diff
@@ -179,7 +178,11 @@ def _get_audio_file_size(self, url):
 
         TODO(obulat): move filesize detection to the polite crawler
         """
-        return int(requests.head(url).headers["content-length"])
```
krysal, line 182: It seems like the TODO task is done 😄 and we can remove the comment 👏
Aside from that, LGTM!

krysal (approved):

openverse_catalog/dags/providers/provider_api_scripts/freesound.py
AetherUnbound: Well not quite 😅 Since we don't have the polite crawler running at all. But I agree, we can remove the comment 🙂

AetherUnbound (approved):
This looks good! One important thing this changes - we stop processing if we're unable to find the filesize, _regardless of whether or not its because the preview 404'd_ or if it was a different reason that prevented us from doing so. I think that's fine, but I wanted to make it explicit that this is our new approach for this DAG since we're not differentiating on a 404 specifically (and I think it'd be particularly difficult to do with the delayed requester 😅).

krysal:
@AetherUnbound Oh, so I see the other cases you mentioned in code comments:

> Freesound can be finicky, so we want to retry it a few times on
        these conditions:
          * SSLError - 'EOF occurred in violation of protocol (_ssl.c:1129)'
          * ConnectionError - '[Errno 113] No route to host'

But what kind of handling was you expecting here to happen? I think we already do the retry with the use of the delayed requester (that I confused with the crawler thing, my bad!) and don't see what else can be done in the different situations, other than discarding the track.

AetherUnbound:
That's a good point of clarification @krysal, there's quite a few things happening with this particular request! In the cases you point out, the `SSLError` and `ConnectionError`, those are _transient errors that occurred when making the request to Freesound_ (added in #330 & #413 respectively). We [do have retries on the requester](https://github.com/WordPress/openverse-catalog/blob/6c35cabf96b1eab2426020b47cad41a408cacf4d/openverse_catalog/dags/common/requester.py#L137-L160), but those are only for cases where the connection was able to be made successfully (albeit returning a non 200 response or an error in the response JSON). Since the errors we were seeing with Freesound were raising exceptions even within that block, we added the outer `@retry` wrappers to retry those exceptions. Freesound is the only provider this happens with, though if we begin to see it happening consistently with other providers, perhaps it would make sense to retry _all_ errors initially within the requester rather than only successful connection requests.

The point with my comment was merely that it'd be possible for Freesound to return a 500 (or some other non-200 error code) consistently for all retries, but still potentially reference a valid record. In this case we'd discard the track entirely since the preview was not available (at the time). Previously this would raise an exception and potentially allow us to retry, now in that case the ingester will just move on. I think it's a very small edge case that we're not likely to encounter, but I wanted to point it out nonetheless 🙂

Issue #953, opened 2023-01-09
Add a task ID specification to the silenced slack notifications information


AetherUnbound:
## Problem

We have some DAGs that may fail at different tasks for different reasons, but with the same predicate. For instance, the Finnish DAG is failing with `AirflowTaskTimeout`s for both the **pull data** task ([reason](https://github.com/WordPress/openverse-catalog/issues/913)) and the **clean data** task ([reason](https://github.com/WordPress/openverse-catalog/issues/450)). We need some way to distinguish these alerts so they can be handled & silenced separately.

## Description

I propose adding a `task_id_pattern` field to the [`SilencedSlackNotification`](https://github.com/WordPress/openverse-catalog/blob/06335539a187036435e0cb02fa94ee4324a7c8da/openverse_catalog/dags/common/slack.py#L69) typed dict which accepts a regex pattern to match against a given task ID (this can't be a direct task ID equality match because the task IDs within task groups _within reingestion workflows_ become quite unique). This would allow us to specify multiple GitHub issues for the same exception across different tasks within a DAG.

PR #976, opened 2023-01-28
Add Airflow variable used to configure overrides for task timeouts

### LINKED ISSUE #724: Allow execution timeouts to be overridden by Variables
## Description

Now that we have catchup enabled, we've recently encountered a scenario with Wikimedia where we're performing a backfill over many days (which also spawned #644). In some of these runs, the execution timed out because of the number of records being processed.

It would be ideal in these cases if we could _temporarily alter_ the execution timeout for any tasks which may explicitly set it (e.g. `pull_media_data`, `load_to_s3`, etc). We could do this by having a Variable which can be configured per DAG ID (see #644, #654, #643, #650) which would allow the execution timeouts to be overridden. This could happen within the `__post_init__` function of the `ProviderWorkflows` dataclass:

https://github.com/WordPress/openverse-catalog/blob/8e7e989ec3e6922681ea18dfb893402f326550d9/openverse_catalog/dags/providers/provider_workflows.py#L91

**Note**: Airflow will not let you modify the execution timeout of an _already running_ task. This setup would be ideal for a few niche scenarios:

- The backfill scenario described above; we know that a few runs went over time, we can bump up the limit, clear those runs, then remove the override after they're complete.
- Determining a reasonable execution timeout for particularly long-running tasks. An example here would be iNaturalist; we still aren't sure how long that would take, and being able to iterate on the appropriate timeout values in production without having to submit multiple PRs to change DAG config would be useful.

## Implementation

stacimc:
## Fixes

Fixes #724 by @AetherUnbound 

## Description

Adds an Airflow Variable that can be used optionally to temporarily alter the execution timeouts for *any task in a provider DAG*, using a `task_id_pattern`. The timeout overrides are configured in the format `<days>d:<hours>h:<minutes>m:<seconds>s`. Example using the `CONFIGURATION_OVERRIDES` Airflow variable:

```
{
  # Example showing overriding timeouts for multiple tasks within a DAG
  "brooklyn_museum_workflow": [
    {
      "task_id_pattern": "pull_image_data",
      "timeout": "10h"  # 10 hrs
    },
    {
      "task_id_pattern": "report_load_completion",
      "timeout": "4s"  # 4 seconds
    }
  ],
  # Example showing task_id_pattern matching reingestion tasks (that have `day_shift_x` suffixes)
  "metropolitan_museum_reingestion_workflow": [
    {
      "task_id_pattern": "upsert_data",
      "timeout": "6d:11h"  # 6 days, 11 hours
    }
  ],
  "inaturalist_workflow": [
    {
      "task_id_pattern": "load_taxa",
      "timeout": "6d:10h:9s"  # 6 days, 10 hours, 9 seconds
    },
    # Example of a mapped task
    {
      "task_id_pattern": "load_transformed_data",
      "timeout": "10m"  # 10 minutes
    }
  ]
}
```

This is intentionally implemented in such a way that it can be extended very quickly to override other properties of tasks. I had `retries` in mind but did not include it in this PR for simplicity.

## Testing Instructions

You can use the example `CONFIGURATION_OVERRIDES` provided above, and play around with adding other configurations. For each timeout you configure, run the DAG and check the `Task Instance Details` page for the task to verify that the `execution_timeout` is set to the one you configured. You can try setting really short timeouts on long-running tasks like `pull_data` to verify that the timeouts will actually be hit.

Make sure to test:
* tasks on a regular ingestion workflow, such as `pull_image_data`
* tasks on a reingestion workflow with day_shift applied. Example, try `upsert_data` and verify that the timeout gets applied to all `upsert_data_day_shift_x` tasks within the workflow
* tasks on iNaturalist (our only example of a 'custom' workflow)
  * specifically, iNaturalist's `load_transformed_data` task, which is a Mapped task. The timeout should be applied to each mapped task

Notes:
* Improperly formatted overrides are simply ignored, rather than raising an error which will break all provider DAGs.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_dag_factory.py
+++ b/openverse_catalog/dags/providers/provider_dag_factory.py
@@ -64,6 +64,7 @@ https://github.com/creativecommons/cccatalog/issues/334)
 """
 import logging
 import os
+import re
 from string import Template
 
 from airflow import DAG
@@ -80,7 +81,11 @@ from providers.factory_utils import (
     pull_media_wrapper,
 )
 from providers.provider_reingestion_workflows import ProviderReingestionWorkflow
-from providers.provider_workflows import ProviderWorkflow
+from providers.provider_workflows import (
+    ProviderWorkflow,
+    TaskOverride,
+    get_time_override,
+)
 
 
 logger = logging.getLogger(__name__)
@@ -95,6 +100,39 @@ DATE_RANGE_ARG_TEMPLATE = "{{{{" + _DATE_RANGE_INNER_TEMPLATE + "}}}}"
 DATE_PARTITION_ARG_TEMPLATE = Template(
     "$media_type/$provider_name/{{ date_partition_for_prefix(dag.schedule_interval, dag_run.logical_date, $reingestion_date ) }}"  # noqa
 )
+
+
+def _apply_configuration_overrides(dag: DAG, overrides: list[TaskOverride]):
+    """
+    For each task in the given DAG, check if any overrides are configured and,
+    if so, apply them.
+    """
+    if not overrides:
+        return
+
+    for task in dag.tasks:
+        if (task_overrides := _get_overrides_for_task(task.task_id, overrides)) is None:
+            continue
+
+        # Format timeout override and apply
+        timeout_override = get_time_override(task_overrides.get("timeout"))
+        if timeout_override:
+            task.execution_timeout = timeout_override
+
+
+def _get_overrides_for_task(
+    task_id: str, overrides: list[TaskOverride]
+) -> TaskOverride | None:
+    """
+    Return any configured TaskOverride whose task_id_pattern matches the
+    id of the given task.
+    """
+    for override in overrides:
+        if re.search(override["task_id_pattern"], task_id):
+            return override
+
+    # No overrides were found for this task.
+    return None
 
 
 def create_ingestion_workflow(
@@ -348,6 +386,9 @@ def create_provider_api_workflow_dag(conf: ProviderWorkflow):
 
         ingest_data >> report_load_completion
 
+    # Apply any overrides from the DAG configuration
+    _apply_configuration_overrides(dag, conf.overrides)
+
     return dag
 
 
@@ -499,5 +540,8 @@ def create_day_partitioned_reingestion_dag(
         # report_load_completion is downstream of all the ingestion TaskGroups in the
         # final list.
         partitioned_ingest_workflows[-1] >> report_load_completion
+
+    # Apply any overrides from the DAG configuration
+    _apply_configuration_overrides(dag, conf.overrides)
 
     return dag
--- a/openverse_catalog/dags/providers/provider_dag_factory.py
+++ b/openverse_catalog/dags/providers/provider_dag_factory.py
@@ -68,6 +68,7 @@ import re
 from string import Template
 
 from airflow import DAG
+from airflow.models.mappedoperator import MappedOperator
 from airflow.models.param import Param
 from airflow.operators.empty import EmptyOperator
 from airflow.operators.python import PythonOperator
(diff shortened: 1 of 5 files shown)


AetherUnbound:
Dang, this is really tough! The flexibility of defining agnostic, per-task timeouts is neat. But I think about what I had to do for #953 - because of the reingestion workflows, we'd probably also need to make this a task ID _pattern_ for matching timeouts 😕 How much additional complexity would that add do you think? 

I think it's fair to try and target the typical `ProviderDataIngester` classes with this, and maybe try to fit iNaturalist _into_ that mold as you're suggesting. Even that utility alone would be beneficial!

stacimc:
@AetherUnbound It ended up being easier than I thought! Updated version of the PR now allows arbitrarily overriding timeout on any task in a provider DAG, including iNaturalist.

openverse_catalog/dags/providers/provider_workflows.py
```diff
@@ -26,6 +27,47 @@
...
+        return None
+
+    ts = time_str.split(":")
+    if len(ts) != 4 or not all(t.isdigit() for t in ts):
+        # Incorrectly formatted time string. Do not apply. We don't raise
+        # an error in order to avoid breaking DAGs.
+        return None
+
+    [days, hours, minutes, seconds] = [int(t) for t in ts]
+    return timedelta(days=days, hours=hours, minutes=minutes, seconds=seconds)
```
AetherUnbound, line 70: I know that this will add complexity to the parsing but...do you think we could do `<days>d:<hours>h:<minutes>m:<seconds>s`? E.g. `4d:5h:0m:10s`? This might make the timeouts a little bit clearer, I was confused initially as to what the values represented.

Airflow uses [pendulum](https://pendulum.eustace.io/docs/) under the hood, so I think we could simplify this too by using something like `from_format` or similar? It doesn't look like there's a method for getting from strings to durations but we might be able to go from a datetime to a timedelta that way. What do you think?

openverse_catalog/dags/providers/provider_reingestion_workflows.py
```diff
@@ -47,11 +47,8 @@ class ProviderReingestionWorkflow(ProviderWorkflow):
 
     def __post_init__(self):
         if not self.dag_id:
-            # Call super() first to initialize the provider_name
-            super().__post_init__()
-            # Override the dag_id
-            self.dag_id = f"{self.provider_name}_reingestion_workflow"
-            return
+            _, provider_name = self._get_module_info()
```
AetherUnbound, line 50: This is a _much_ better approach! 😄

openverse_catalog/dags/providers/provider_dag_factory.py
```diff
@@ -97,6 +103,44 @@
...
+
+        # Format timeout override and apply
+        timeout_override = get_time_override(task_overrides.get("timeout"))
+        if timeout_override:
+            # If the task is a MappedOperator, apply the timeout to partial
+            # kwargs to ensure it is set on each mapped task.
+            if isinstance(task, MappedOperator):
+                task.partial_kwargs["execution_timeout"] = timeout_override
+            else:
+                task.execution_timeout = timeout_override
```
AetherUnbound, line 126: Wow, what a find! 😱

AetherUnbound (commented):
Woohoo! What an excellent feature this will unlock for us. I have one request regarding the timeout format that I'd like your thoughts on 🕐

openverse_catalog/dags/providers/provider_workflows.py
```diff
@@ -51,22 +56,41 @@ class TaskOverride(TypedDict):
...
+        count, unit = match.groups()
+        match unit:
+            case "d":
+                days = int(count)
+            case "h":
+                hours = int(count)
+            case "m":
+                minutes = int(count)
+            case "s":
+                seconds = int(count)
```
AetherUnbound, line 92: A case statement!! Wow, my first ever! 😄

openverse_catalog/dags/providers/provider_workflows.py
AetherUnbound: Oh sorry, I didn't mean to imply that we should have something more complex than that! I think your implementation looks great ✨

AetherUnbound (approved):
This is an excellent change! I know it can be hard to test, but I love the ability to _just_ have the time range we need specified (e.g. `10h`). Very excited for this!

openverse_catalog/dags/providers/provider_workflows.py
stacimc: Oh totally, I'm stopping _myself_ from getting carried away adding more 😄 I was initially trying to make a single regex for matching but updated it so you don't have to include all four parts 😆 I think that ought to cover realistic use cases.

krysal (approved):
It's so flexible! Awesome, it worked like a charm! 💯

Issue #511, opened 2022-05-17
Ensure that all media have `license_url` in `meta_data` field


obulat:
## Problem

The API expects `license_name` and `license_version` to exist in the database. For the `license_url`, it checks if it exists in the `meta_data`, and if not, calculates the URL based on the license name and version.

In #66 we made sure that all the items add a valid `license_url` parameter to the `meta_data` field.   We can be confident that all data that was added after that PR contains `license_url`, but we need to check the older data.

Currently, there are 366,974 images that have a `NULL` value in `meta_data`, based on the following SQL query:
```
deploy@localhost:openledger> SELECT
 COUNT(*) as total,
 COUNT(*) FILTER (WHERE meta_data IS NULL) as meta_data,
 FROM image;
-[ RECORD 1 ]-------------------------
total                   | 563667181
meta_data               | 366974
SELECT 1
```
from the [public Slack chat message](https://wordpress.slack.com/archives/C02012JB00N/p1652848362174809?thread_ts=1652802243.895939&cid=C02012JB00N)

## Description

By checking the database for `license_url` and backfilling the values, if necessary, we can clean up the downstream API and frontend code that will expect the `license_url` to be present as a required field.

## Alternatives

We could continue computing the license URL as we do now, which would mean that we have duplicated code across the stack.

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

obulat:
Some more details about this:

- `meta_data` object has two items: `raw_license_url` and the `license_url`. They have the same value if the provider gives the canonical license url: starting with `https`, with valid combination of license name, version, the jurisdiction (for ported licenses) and a trailing slash.
Prior to running any backfillling queries, we should do quick test for objects that there is no items with the following weird state: having the `raw_license_url` but no `license_url` in the `meta_data`. There probably are no such items, but it would be nice to check to be sure.

- It would be nice to see if we have any value for license version besides "N/A" (for publicdomain), "2.1" (for ported licenses only?), and "1.0", "2.0", "2.5", "3.0", "4.0".

My understanding is that if we have the license name and version, but no URL, in the database, then the license is not ported (i.e., does not have the jurisdiction at the end of the path, like `/jp`, `/au`, `/es`, because it's not recoverable from the two datapoints. _Or_ we have something like "2.1/jp" for the version value...

AetherUnbound:
I'm trying one other query but in case I forget:
```sql
[...]

PR #1005, opened 2023-02-17
Add a DAG for backfilling license_url when meta_data is null

obulat:
## Fixes

Related to #511 by @obulat

## Description

This PR adds `license_url` to the `meta_data` jsonb field where the `meta_data` field is `NULL`. 
To make the query faster, we split it into several steps:

1. Count the number of rows where `meta_data` is `NULL`.
2. Update the `meta_data`:
  a. Select all `identifier`s of the items where `meta_data` is `NULL`. Having `identifier`s will enable us not to do `sequential scan` through all of the database (i.e., not to test each row individually), and only work with the 366 974items identified in the related issue.
  b. For each `license` and `license_version` pair, we run the `UPDATE` query setting the `meta_data` to `{"license_url": <URL>}`. Note: I'm using `Json` here although the column is `jsonb`. I'm not sure if it's correct, but it works locally.
3. The final report checks for rows where `meta_data` is `NULL`. Normally, the count should be 0, but due to the possible lack of `license_version` in some of the rows (see Possible problems below), it might not be.

### Possible problems
There is an [issue in the CC repository](https://github.com/cc-archive/cccatalog/issues/244) about missing `license_version` in some items. If there are still items without `license_version` in the database, their count will be reported in the last step.

### Limitations
This PR only solves a part of the problem for items where `meta_data` is `NULL`. 

I have confirmed that there are some items with a non-null `meta_data` that don't have `license_url` field in `meta_data` (e.g., item with identifier `d876532b-01da-4e14-aa73-237c0a160d1f`).  For this, the possible solution is to add the `license_url` check in the ingestion server, adding it if necessary, and then propagate that data up to the catalog. This will be much faster than doing this on the catalog side because here, we would need to "sequentially scan" hundreds of millions of records.

I'm also not sure about the reporting step, and open to suggestions about it. Although, I'm not sure how much we need to report for a one-off maintenance DAG.

## Testing Instructions

Run some provider API DAGs if you have no data in your local database.
Set `meta_data`  to `NULL`:
`UPDATE image SET meta_data = null`
You can also set some of the `license_version` values to `NULL`:
```
WITH rows AS (SELECT identifier FROM image LIMIT 10)
UPDATE image SET license_version=null
WHERE identifier IN (SELECT identifier FROM image LIMIT 10);
```
Run the `add_license_url` DAG and check that it adds a json object to the `meta_data` field with the license URL in it. Also, if you set the license_version to `NULL`, you should see that reported in the last step.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.
- [x] I ran the DAG documentation generator (if applicable).

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/maintenance/add_license_url.py
+++ b/openverse_catalog/dags/maintenance/add_license_url.py
@@ -1,10 +1,10 @@
 """
 # Add license URL
 
-Add license_url to all rows that have NULL in their meta_data fields.
-The license_url is constructed from the license and license_version fields.
+Add `license_url` to all rows that have `NULL` in their `meta_data` fields.
+The `license_url` is constructed from the `license` and `license_version` fields.
 
-This is a maintenance DAG that should run once.
+This is a maintenance DAG that should be run once.
 """
 import logging
 import os
@@ -125,7 +125,7 @@ Now, there are {null_meta_data_records} records with NULL meta_data.
     send_message(
         message,
         username="Airflow DAG Data Normalization - license_url",
-        dag_id="data_normalization",
+        dag_id=DAG_ID,
     )
 
     logger.info(message)
--- a/openverse_catalog/dags/maintenance/add_license_url.py
+++ b/openverse_catalog/dags/maintenance/add_license_url.py
@@ -1,0 +1,174 @@
+"""
+# DAG to add license_url to NULL meta_data fields
+A maintenance one-off workflow that adds meta_data.license_url to all images.
+"""
+import logging
+import os
+from datetime import datetime, timedelta
+from textwrap import dedent
+from typing import Literal
+
+import jinja2
+from airflow.models import DAG
+from airflow.operators.python import BranchPythonOperator, PythonOperator
+from airflow.providers.postgres.hooks.postgres import PostgresHook
+from common import slack
+from common.constants import XCOM_PULL_TEMPLATE
+from common.licenses.constants import get_reverse_license_path_map
+from common.slack import send_message
+from psycopg2._json import Json
+
+
+DAG_ID = "add_license_url"
+DB_CONN_ID = os.getenv("OPENLEDGER_CONN_ID", "postgres_openledger_testing")
+
+ALERT_EMAIL_ADDRESSES = ""
+RETURN_ROW_COUNT = lambda c: c.rowcount  # noqa: E731
+
+logger = logging.getLogger(__name__)
+
+base_url = "https://creativecommons.org/"
+
+license_map = get_reverse_license_path_map()
+
+
+def get_statistics(
+    postgres_conn_id: str,
+) -> Literal["update_license_url", "final_report"]:
+    logger.info("Getting image records without license_url in meta_data.")
+    postgres = PostgresHook(postgres_conn_id=postgres_conn_id)
+    null_meta_data_count = postgres.get_first(
+        dedent(
+            """
+        SELECT COUNT(*) from image
+        WHERE meta_data IS NULL;
+        """
+        )
+    )[0]
+
+    logger.info(f"There are {null_meta_data_count} records with NULL in meta_data.")
+    if null_meta_data_count > 0:
+        return "update_license_url"
+    else:
+        return "final_report"
+
+
+def update_license_url(postgres_conn_id: str) -> dict[str, int]:
+    """Add license_url to meta_data batching all records with the same license.
+    :param postgres_conn_id: Postgres connection id
+    """
+
+    logger.info("Getting image records without license_url.")
+    postgres = PostgresHook(postgres_conn_id=postgres_conn_id)
+
+    total_count = 0
+    total_counts = {}
+    for license_items, path in license_map.items():
+        license_name, license_version = license_items
+        logger.info(f"Processing {license_name} {license_version}, {license_items}.")
+        license_url = f"{base_url}{path}/"
+
+        select_query = dedent(
+            f"""
+            SELECT identifier FROM image
+            WHERE (
+            meta_data is NULL AND license = '{license_name}'
+            AND license_version = '{license_version}')
+            """
+        )
+        result = postgres.get_records(select_query)
+
+        if not result:
+            logger.info(f"No records to update with {license_url}.")
+            continue
+        logger.info(f"{len(result)} records to update with {license_url}.")
+        license_url_col = {"license_url": license_url}
+        update_license_url_query = dedent(
+            f"""
+            UPDATE image
+            SET meta_data = {Json(license_url_col)}
+            WHERE identifier IN ({','.join([f"'{r[0]}'" for r in result])});
+            """
+        )
+
+        updated_count = postgres.run(
+            update_license_url_query, autocommit=True, handler=RETURN_ROW_COUNT
+        )
+        logger.info(f"{updated_count} records updated with {license_url}.")
+        total_counts[license_url] = updated_count
+        total_count += updated_count
+
+    logger.info(f"{total_count} image records with missing license_url updated.")
+    logger.info(f"Total updated counts: {total_counts}")
+    return total_counts
+
+
+def final_report(postgres_conn_id: str, item_count):
+    logger.info(
+        "Added license_url to all items. Checking for any records "
+        "that still have NULL metadata."
+    )
+    postgres = PostgresHook(postgres_conn_id=postgres_conn_id)
+    null_meta_data_records = postgres.get_first(
+        dedent("SELECT COUNT(*) from image WHERE meta_data IS NULL;")
+    )[0]
+    logger.info(f"There are {null_meta_data_records} records with NULL meta_data.")
+
+    message = f"""
+Added license_url to *{item_count}* items`
+Now, there are {null_meta_data_records} records with NULL meta_data.
+"""
+    send_message(
+        message,
+        username="Airflow DAG Data Normalization - license_url",
+        dag_id="data_normalization",
+    )
+
+    logger.info(message)
+
+
+DAG_DEFAULT_ARGS = {
+    "owner": "data-eng-admin",
+    "depends_on_past": False,
+    "start_date": datetime(2020, 6, 15),
+    "template_undefined": jinja2.Undefined,
+    "email_on_retry": False,
+    "retries": 0,
+    "retry_delay": timedelta(minutes=1),
+    "on_failure_callback": slack.on_failure_callback,
+}
+
+
+dag = DAG(
+    dag_id=DAG_ID,
+    default_args=DAG_DEFAULT_ARGS,
+    catchup=False,
+    # Use the docstring at the top of the file as md docs in the UI
+    doc_md=__doc__,
+    tags=["data_normalization"],
+)
+
+with dag:
+    get_statistics = BranchPythonOperator(
+        task_id="get_stats",
+        python_callable=get_statistics,
+        op_kwargs={"postgres_conn_id": DB_CONN_ID},
+    )
+    update_license_url = PythonOperator(
+        task_id="update_license_url",
+        python_callable=update_license_url,
+        op_kwargs={"postgres_conn_id": DB_CONN_ID},
+    )
+    final_report = PythonOperator(
+        task_id="final_report",
+        python_callable=final_report,
+        op_kwargs={
+            "postgres_conn_id": DB_CONN_ID,
+            "item_count": XCOM_PULL_TEMPLATE.format(
+                update_license_url.task_id, "return_value"
+            ),
+        },
+    )
+
+    get_statistics >> [update_license_url, final_report]
+    update_license_url >> final_report
--- a/openverse_catalog/dags/maintenance/add_license_url.py
+++ b/openverse_catalog/dags/maintenance/add_license_url.py
@@ -24,6 +24,8 @@ from psycopg2._json import Json
 
 
 DAG_ID = "add_license_url"
+UPDATE_LICENSE_URL = "update_license_url"
+FINAL_REPORT = "final_report"
 
 ALERT_EMAIL_ADDRESSES = ""
 
@@ -31,12 +33,10 @@ logger = logging.getLogger(__name__)
 
 base_url = "https://creativecommons.org/"
 
-license_map = get_reverse_license_path_map()
-
 
 def get_statistics(
     postgres_conn_id: str,
-) -> Literal["update_license_url", "final_report"]:
+) -> Literal[UPDATE_LICENSE_URL, FINAL_REPORT]:
     logger.info("Getting image records without license_url in meta_data.")
     postgres = PostgresHook(postgres_conn_id=postgres_conn_id)
     null_meta_data_count = postgres.get_first(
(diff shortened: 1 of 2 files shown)


openverse_catalog/dags/maintenance/add_license_url.py
```diff
@@ -0,0 +1,179 @@
+"""
+# Add license URL
+
+Add license_url to all rows that have NULL in their meta_data fields.
+The license_url is constructed from the license and license_version fields.
+
+This is a maintenance DAG that should run once.
```
AetherUnbound, line 7: Nit: a few tweaks to improve markdown rendering

```suggestion
Add `license_url` to all rows that have `NULL` in their `meta_data` fields.
The `license_url` is constructed from the `license` and `license_version` fields.

This is a maintenance DAG that should be run once.
```

openverse_catalog/dags/maintenance/add_license_url.py
```diff
@@ -0,0 +1,179 @@
...
+from airflow.providers.postgres.hooks.postgres import PostgresHook
+from common import slack
+from common.constants import XCOM_PULL_TEMPLATE
+from common.licenses.constants import get_reverse_license_path_map
+from common.slack import send_message
+from psycopg2._json import Json
+
+
+DAG_ID = "add_license_url"
+DB_CONN_ID = os.getenv("OPENLEDGER_CONN_ID", "postgres_openledger_testing")
```
AetherUnbound, line 27: We can use our constant for this!
https://github.com/WordPress/openverse-catalog/blob/895e907b6f9594b4d43db6a07eae0b3cb4a0e7d0/openverse_catalog/dags/common/constants.py#L32

openverse_catalog/dags/maintenance/add_license_url.py
```diff
@@ -0,0 +1,179 @@
...
+
+ALERT_EMAIL_ADDRESSES = ""
+RETURN_ROW_COUNT = lambda c: c.rowcount  # noqa: E731
```
AetherUnbound, line 30: Could we re-use the existing lambda we have for this?
https://github.com/WordPress/openverse-catalog/blob/af205795b412da9f3d0d7f6d7e066b7d039b1763/openverse_catalog/dags/common/loader/sql.py#L50

openverse_catalog/dags/maintenance/add_license_url.py
```diff
@@ -0,0 +1,179 @@
...
+
+logger = logging.getLogger(__name__)
+
+base_url = "https://creativecommons.org/"
+
+license_map = get_reverse_license_path_map()
```
AetherUnbound, line 36: Since this is set up as a module-level variable, it'll be executed on each DAG parse cycle. It seems it's only used in the `update_license_url` function, perhaps we can just define it and call it there?

openverse_catalog/dags/maintenance/add_license_url.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def get_statistics(
+    postgres_conn_id: str,
+) -> Literal["update_license_url", "final_report"]:
+    logger.info("Getting image records without license_url in meta_data.")
+    postgres = PostgresHook(postgres_conn_id=postgres_conn_id)
```
AetherUnbound, line 43: Just noting that if this gets in after #717 we'll want to add timeouts to this!

openverse_catalog/dags/maintenance/add_license_url.py
```diff
@@ -0,0 +1,179 @@
...
+        WHERE meta_data IS NULL;
+        """
+        )
+    )[0]
+
+    logger.info(f"There are {null_meta_data_count} records with NULL in meta_data.")
+    if null_meta_data_count > 0:
+        return "update_license_url"
+    else:
+        return "final_report"
```
AetherUnbound, line 57: We could potentially define these task ID names at the top (like we do with `DAG_ID`) and just reuse them throughout, just so we don't have strings floating around in several places 🙂

openverse_catalog/dags/maintenance/add_license_url.py
```diff
@@ -0,0 +1,179 @@
...
+def final_report(postgres_conn_id: str, item_count):
+    logger.info(
+        "Added license_url to all items. Checking for any records "
+        "that still have NULL metadata."
+    )
+    postgres = PostgresHook(postgres_conn_id=postgres_conn_id)
+    null_meta_data_records = postgres.get_first(
+        dedent("SELECT COUNT(*) from image WHERE meta_data IS NULL;")
+    )[0]
+    logger.info(f"There are {null_meta_data_records} records with NULL meta_data.")
```
AetherUnbound, line 119: This is the same logic as the function above I think, can we combine it into a utility function that both tasks call?

openverse_catalog/dags/maintenance/add_license_url.py
```diff
@@ -0,0 +1,179 @@
...
+
+    message = f"""
+Added license_url to *{item_count}* items`
+Now, there are {null_meta_data_records} records with NULL meta_data.
+"""
+    send_message(
+        message,
+        username="Airflow DAG Data Normalization - license_url",
+        dag_id="data_normalization",
```
AetherUnbound, line 128: Shouldn't this be?
```suggestion
        dag_id=DAG_ID,
```

openverse_catalog/dags/maintenance/add_license_url.py
```diff
@@ -0,0 +1,179 @@
...
+DAG_DEFAULT_ARGS = {
+    "owner": "data-eng-admin",
+    "depends_on_past": False,
+    "start_date": datetime(2020, 6, 15),
+    "template_undefined": jinja2.Undefined,
+    "email_on_retry": False,
+    "schedule_interval": None,
+    "retries": 0,
+    "on_failure_callback": slack.on_failure_callback,
+}
```
AetherUnbound, line 143: Should this be here? I think we can inherit from here and extend where necessary:
https://github.com/WordPress/openverse-catalog/blob/895e907b6f9594b4d43db6a07eae0b3cb4a0e7d0/openverse_catalog/dags/common/constants.py#L16-L29

AetherUnbound (commented):
This is super straightforward! I really like your approach of 1) select identifiers to update 2) determine update 3) issue update query. I wonder if it might be a good idea to create a more generic interface for that set of actions, since it's one I can image us doing many times in the future 😄 I don't think we need to do it as part of this PR, but it's something to consider as we make more of these down the line.

I have a few suggestions for ways we can simplify this!

openverse_catalog/dags/maintenance/add_license_url.py
obulat: Thank you for this suggestion and the explanation, I was counting on you all to improve this DAG :)

openverse_catalog/dags/maintenance/add_license_url.py
obulat: We can add it in the `DAG` call, right?

openverse_catalog/dags/maintenance/add_license_url.py
obulat: I've used the constant, and have a question about the "schedule_interval": should it be in the `default_args` or the DAG args? Currently, I have both, don't know why.

openverse_catalog/dags/maintenance/add_license_url.py
obulat: Great idea, thank you!

obulat:
> This is super straightforward! I really like your approach of 1) select identifiers to update 2) determine update 3) issue update query. I wonder if it might be a good idea to create a more generic interface for that set of actions, since it's one I can image us doing many times in the future 😄 I don't think we need to do it as part of this PR, but it's something to consider as we make more of these down the line.

I love the idea of creating a generic interface for the actions to update the catalog. I think that in the future we will have several updates that will not need the first step, the selection of identifiers, in the catalog. They will probably be provided by the API (from the cleanup step, as in https://github.com/WordPress/openverse-api/pull/1126) or the Polite crawler, as TSV or some other format. Is it easy to deploy such one-off DAGs in Airflow? Maybe we could even use something else (like Kafka?) in the future for updating the upstream database?

openverse_catalog/dags/maintenance/add_license_url.py
AetherUnbound: We don't need to make any changes to the DAG, the `task` variable will just need to be handed into the `PostgresHook` like some of the examples modified in that PR.

openverse_catalog/dags/maintenance/add_license_url.py
AetherUnbound: The `default_args` are what get passed to each task within the DAG, so the schedule interval should be provided to the `Dag` itself directly!

AetherUnbound:
> I love the idea of creating a generic interface for the actions to update the catalog. I think that in the future we will have several updates that will not need the first step, the selection of identifiers, in the catalog. They will probably be provided by the API (from the cleanup step, as in [WordPress/openverse-api#1126](https://github.com/WordPress/openverse-api/pull/1126)) or the Polite crawler, as TSV or some other format. Is it easy to deploy such one-off DAGs in Airflow? Maybe we could even use something else (like Kafka?) in the future for updating the upstream database?

That's such a good point! It'd definitely be a case to consider as part of this. For something dynamic like you're describing, having to deploy a new DAG could be cumbersome. But it does also provide us a "paper trail" of sorts for migrations, so that has its own advantages. I could see a case where we use Variables or DAG run config to entirely dictate the set of operations, which would remove the need to add a new DAG each time. Something to consider for the RFC there!

AetherUnbound (approved):
Ran this locally and it worked exactly as expected! I have one note but otherwise this is good to go IMO ⭐

openverse_catalog/dags/maintenance/add_license_url.py
```diff
@@ -0,0 +1,162 @@
...
+        op_kwargs={
+            "postgres_conn_id": POSTGRES_CONN_ID,
+            "item_count": XCOM_PULL_TEMPLATE.format(
+                update_license_url.task_id, "return_value"
+            ),
+        },
+    )
+
+    get_statistics >> [update_license_url, final_report]
+    update_license_url >> final_report
```
stacimc, line 162: I think this could be written more simply as `get_statistics >> update_license_url >> final_report`

openverse_catalog/dags/maintenance/add_license_url.py
```diff
@@ -0,0 +1,162 @@
...
+        updated_count = postgres.run(
+            update_license_url_query, autocommit=True, handler=RETURN_ROW_COUNT
+        )
+        logger.info(f"{updated_count} records updated with {license_url}.")
+        total_counts[license_url] = updated_count
+        total_count += updated_count
+
+    logger.info(
+        f"{total_count} image records with missing license_url updated.\n"
+        f"Total updated counts: {total_counts}"
```
stacimc, line 104: Not necessary, but this could be easier to read if the `total_counts` are separated by line breaks.

openverse_catalog/dags/maintenance/add_license_url.py
stacimc: For the `PostgresHook` directly we pass a `default_statement_timeout` instead of just the `task`, like:

```
postgres = PostgresHook(
        postgres_conn_id=postgres_conn_id,
        default_statement_timeout=PostgresHook.get_execution_timeout(task),
    )
```

If not, it'll fall back to the 1-hr default timeout from `DAG_DEFAULT_ARGS`.

stacimc (approved):
This worked great for me! Thank you for the testing instructions, and the refresher on the context behind this. I really like the approach and think it'll be a really interesting template for a generic interface sometime down the road :)

No blocking feedback, this looks good to me!

openverse_catalog/dags/maintenance/add_license_url.py
obulat: If we simplify it, would it skip the `update_license_url` step if the `get_statistics` says that we don't need to update anything? The intention was to skip the update step if we ever run this on a fully cleaned dataset, but I'm not very good with branching operators :)

openverse_catalog/dags/maintenance/add_license_url.py
obulat: I wonder if this 1-hr default timeout is OK here, @stacimc?

openverse_catalog/dags/maintenance/add_license_url.py
AetherUnbound: Ahh, that's right! Do you mind adding a comment to that effect Olga? You're totally right, the current setup is required for the branching operator, but I could see making that change in the future and accidentally breaking something 😅

AetherUnbound:
@obulat this could be merged if you suspect the default timeout (1h) will be sufficient for each posgtres operation made by the hook, but otherwise we'll need to add the task timeout to the hook (or something more appropriate) per [Staci's comment](https://github.com/WordPress/openverse-catalog/pull/1005#discussion_r1120938689).

obulat:
@stacimc, thank you for your instructions for setting the timeout. We have 366,974 images that should be affected. I have no idea what timeout value we should use. 5 hours?.. What do you think?

openverse_catalog/dags/maintenance/add_license_url.py
stacimc: You're totally correct, I missed that there was a branching operator! 😅

stacimc:
The new timeout looks good to me 👍 I think this can be merged and enabled in production. The DAG will be off by default so you'll need to enable it through the Airflow UI.

obulat:
I triggered this query at 05.53 UTC today (2023-03-13).  Here are some notes on the run:

The first step reported a different number of records with `NULL` in metadata, only 28 824, compared with the query results from the last time we tried (more than 300 000 records):

`[2023-03-13, 06:19:11 UTC] {add_license_url.py:60} INFO - There are 28824 records with NULL in meta_data.`

The second step takes around 30 minutes for each license pair (license name and license version), so it will probably time out. If it does, to improve the performance next time, instead of running a select query for every license pair, we could select the records' `identifier`, `license`, `license_url` from rows where `meta_data` is NULL.
Then, in Python, we group them into a dictionary with the license and license version pair as a key, and identifiers as a value:

```py
license_dict = {}
select_query = "SELECT identifier, license, license_url FROM image WHERE meta_data IS NULL;"
results = postgres.get_records(select_query)

for result in results:
    identifier, license_, version = result
    license_pair = f"{license_},{version}"
    if not license_pair in licenses:
        licenses[license_pair] = [identifier]
    else:
        licenses[license_pair].append(identifier)

for license_pair in license_dict:
    license_, license_version = license_pair.split(",")
    license_url =  get_license_info_from_license_pair(license_, license_version)
    update_query = dedent(
            f"""
            UPDATE image
            SET meta_data = {Json(license_url)}
            WHERE identifier IN ({','.join([f"'{r}'" for r in license_dict[license_pair]])});
            """
     updated_count = postgres.run(
            update_license_url_query, autocommit=True, handler=RETURN_ROW_COUNT
        )
```

obulat:
I just realized that each Postgres statement has a timeout of 5 hours. Since each separate license pair query runs about 30 minutes, the whole DAG _should_ complete successfully. But in the future, it's better not to make many `SELECT` queries if its possible to select items in Python.

PR #1008, opened 2023-02-22
Wikimedia: re-attempt large batches with reduced parameter selection

### LINKED ISSUE #725: Wikimedia DAG times out due to very large batches
## Description

Some recent DagRuns of the `wikimedia_commons_workflow` have timed out before completion in the `pull_data` step. It looks like this is happening due to encountering extremely large batches.

These show up in the logs as very long sections where the reported `gaicontinue` token is identical, and the `gucontinue` token is almost identical except for a changing integer suffix. Here's a very small snippet from a recent run:

```
[2022-09-07, 15:52:30 UTC] {wikimedia_commons.py:106} INFO - New continue token: {'gucontinue': 'Arrows-orphan.svg|arwiki|3120418', 'gaicontinue': '20220802024848|Lohjanjärvi_in_September.jpg', 'continue': 'gaicontinue||imageinfo'}
[2022-09-07, 15:52:30 UTC] {wikimedia_commons.py:130} INFO - Got 250 pages
[2022-09-07, 15:52:30 UTC] {wikimedia_commons.py:130} INFO - Got 250 pages
[2022-09-07, 15:52:30 UTC] {wikimedia_commons.py:130} INFO - Got 250 pages
[2022-09-07, 15:52:31 UTC] {wikimedia_commons.py:106} INFO - New continue token: {'gucontinue': 'Arrows-orphan.svg|arwiki|3125414', 'gaicontinue': '20220802024848|Lohjanjärvi_in_September.jpg', 'continue': 'gaicontinue||imageinfo'}
[2022-09-07, 15:52:31 UTC] {wikimedia_commons.py:130} INFO - Got 250 pages
[2022-09-07, 15:52:31 UTC] {wikimedia_commons.py:130} INFO - Got 250 pages
[2022-09-07, 15:52:31 UTC] {wikimedia_commons.py:130} INFO - Got 250 pages
[2022-09-07, 15:52:32 UTC] {wikimedia_commons.py:106} INFO - New continue token: {'gucontinue': 'Arrows-orphan.svg|arwiki|3127601', 'gaicontinue': '20220802024848|Lohjanjärvi_in_September.jpg', 'continue': 'gaicontinue||imageinfo'}
```

In the DagRun that example is taken from, that batch took almost 11.5 hours to process (from the first log containing the `Arrows-orphan` continue token, to the final `Found batchcomplete` log), but only 208 records were ultimately processed from that batch.

## Reproduction

Go to the Wikimedia DAG in production and observe logs for any of the recent failed `pull_data` tasks. The example here comes from processing for August 02, 2022. You should also be able to manually run the DAG with that `date` to reproduce the issue.

## Possible resolution

Some initial thoughts for ways to resolve this:

1. **Increase the timeout for Wikimedia's `pull_data` task across the board.** I'm concerned about doing this because Wikimedia _generally_ should complete very quickly. This could cause problems in reingestion, which tries to run many Wikimedia dagruns.
2. **Add some handling for long-running batches.** We could try to add a custom timeout to Wikimedia's batch processing, to exit early if a batch like this is encountered. If we do so, we risk losing some data.

As part of this ticket it would be a good idea to look through the logs of a few more of these failures to try to confirm how much data we're getting from these large batches.

## Resolution

AetherUnbound:
## Fixes

Fixes #725 by @stacimc

## Description

**Author's note**: This...is admittedly a really massive and complicated PR :sweat_smile: _(at least it seems that way to me!)_ Wikimedia's API is incredibly complicated, their documentation is hard to navigate, and there's a lot of "magic" that happens with it that can make it hard to introspect & grok. I've tried to summarize this all as best as I could in the DAG documentation, hopefully it's not too loquacious.

This PR changes the behavior of the Wikimedia ingestion to detect when a particular batch has been iterated over more than a certain number of times, and adjust the query parameters for the next request to try and reduce the amount of data returned. It does this by _excluding_ the fields which were returning massive amounts of data. As of today, none of those fields are strictly _necessary_, merely useful. Since the items that would cause the issue in question would rarely complete during the task's iteration anyway, having them with _partial_ data seems much better than not having them at all.

Since we're storing more state on the ingester instance as well, I've modified the tests so each test gets a fresh instantiation. I also confirmed that modifying values _on an instance_ doesn't affect the values that are given for new instances of the class (e.g. changing the max iteration cutoff for an instance doesn't change it at the class level).

## Testing Instructions

**Author's note**: this is hairy and complicated and I'm sorry!!

1. `just recreate`
2. Change the following (just for testing purposes, example diff shown below)
    1. Set `max_page_iteration_before_give_up` to 10
    2. Change the start date of Wikimedia to `2023-02-01`
    3. Change the DAG factory to set all dated DAGs' `catchup` to `False`

```diff
diff --git a/openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py b/openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
index 08faa95..91cc462 100644
--- a/openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
@@ -132,7 +132,7 @@ class WikimediaCommonsDataIngester(ProviderDataIngester):
     # These can be very large, so we need to limit the number of attempts otherwise
     # the DAG will hit the timeout when it comes across these items.
     # See: https://github.com/WordPress/openverse-catalog/issues/725
-    max_page_iteration_before_give_up = 100
+    max_page_iteration_before_give_up = 10
 
     image_mediatypes = {"BITMAP", "DRAWING"}
     audio_mediatypes = {"AUDIO"}
diff --git a/openverse_catalog/dags/providers/provider_dag_factory.py b/openverse_catalog/dags/providers/provider_dag_factory.py
index c460223..e00fa87 100644
--- a/openverse_catalog/dags/providers/provider_dag_factory.py
+++ b/openverse_catalog/dags/providers/provider_dag_factory.py
@@ -305,7 +305,7 @@ def create_provider_api_workflow_dag(conf: ProviderWorkflow):
         max_active_runs=conf.max_active_runs,
         start_date=conf.start_date,
         schedule=conf.schedule_string,
-        catchup=conf.dated,  # catchup is turned on for dated DAGs to allow backfilling
+        catchup=False,  # conf.dated,  # catchup is turned on for dated DAGs to allow backfilling
         doc_md=conf.doc_md,
         tags=[
             "provider",
diff --git a/openverse_catalog/dags/providers/provider_workflows.py b/openverse_catalog/dags/providers/provider_workflows.py
index c0d699f..9556c8a 100644
--- a/openverse_catalog/dags/providers/provider_workflows.py
+++ b/openverse_catalog/dags/providers/provider_workflows.py
@@ -202,7 +202,7 @@ PROVIDER_WORKFLOWS = [
     ),
     ProviderWorkflow(
         ingester_class=WikimediaCommonsDataIngester,
-        start_date=datetime(2020, 11, 1),
+        start_date=datetime(2023, 2, 1),
         schedule_string="@daily",
         dated=True,
         pull_timeout=timedelta(hours=12),
```

4. Trigger Wikimedia DAG with config `{"date": "2023-02-10"}`
5. This kicks off a scheduled DAG because the DAG hasn't run yet, go ahead and mark the scheduled run as failed

image

6. Wait for the run to complete (~40 minutes, if it _does_ complete that's evidence that this works because this date hit the timeout!)
7. Run `just db-shell` then execute `select * from image where provider='wikimedia' and md5(foreign_identifier)=md5('4298234');`, this will return the result for the [wikinews commentary icon](https://commons.wikimedia.org/w/index.php?curid=4298234) which was returning thousands of global usage results.
8. Confirm that `meta_data` field for the result above has a `global_usage_count` that is about 2k
9. Confirm in the logs for the task that it says "Hit iteration count limit for <some token>"

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.
- [x] I ran the DAG documentation generator (if applicable).

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/DAGs.md
+++ b/DAGs.md
@@ -59,23 +59,23 @@ The following are DAGs grouped by their primary tag:
 | --------------------------------------------------------------- | ----------------- | ------- | ------------- |
 | `brooklyn_museum_workflow`                                      | `@monthly`        | `False` | image         |
 | `cleveland_museum_workflow`                                     | `@monthly`        | `False` | image         |
-| [`europeana_workflow`](#europeana_workflow)                     | `@daily`          | `True`  | image         |
-| [`finnish_museums_workflow`](#finnish_museums_workflow)         | `@daily`          | `True`  | image         |
-| [`flickr_workflow`](#flickr_workflow)                           | `@daily`          | `True`  | image         |
+| [`europeana_workflow`](#europeana_workflow)                     | `@daily`          | `False` | image         |
+| [`finnish_museums_workflow`](#finnish_museums_workflow)         | `@daily`          | `False` | image         |
+| [`flickr_workflow`](#flickr_workflow)                           | `@daily`          | `False` | image         |
 | [`freesound_workflow`](#freesound_workflow)                     | `@monthly`        | `False` | audio         |
 | [`inaturalist_workflow`](#inaturalist_workflow)                 | `@monthly`        | `False` | image         |
 | [`jamendo_workflow`](#jamendo_workflow)                         | `@monthly`        | `False` | audio         |
-| [`metropolitan_museum_workflow`](#metropolitan_museum_workflow) | `@daily`          | `True`  | image         |
+| [`metropolitan_museum_workflow`](#metropolitan_museum_workflow) | `@daily`          | `False` | image         |
 | `museum_victoria_workflow`                                      | `@monthly`        | `False` | image         |
 | [`nappy_workflow`](#nappy_workflow)                             | `@monthly`        | `False` | image         |
 | `nypl_workflow`                                                 | `@monthly`        | `False` | image         |
-| [`phylopic_workflow`](#phylopic_workflow)                       | `@daily`          | `True`  | image         |
+| [`phylopic_workflow`](#phylopic_workflow)                       | `@daily`          | `False` | image         |
 | [`rawpixel_workflow`](#rawpixel_workflow)                       | `@monthly`        | `False` | image         |
 | [`science_museum_workflow`](#science_museum_workflow)           | `@monthly`        | `False` | image         |
 | [`smithsonian_workflow`](#smithsonian_workflow)                 | `@weekly`         | `False` | image         |
 | [`smk_workflow`](#smk_workflow)                                 | `@monthly`        | `False` | image         |
 | [`stocksnap_workflow`](#stocksnap_workflow)                     | `@monthly`        | `False` | image         |
-| [`wikimedia_commons_workflow`](#wikimedia_commons_workflow)     | `@daily`          | `True`  | image, audio  |
+| [`wikimedia_commons_workflow`](#wikimedia_commons_workflow)     | `@daily`          | `False` | image, audio  |
 | [`wordpress_workflow`](#wordpress_workflow)                     | `@monthly`        | `False` | image         |
 
 ## Provider Reingestion
@@ -747,7 +747,7 @@ In this case, we're iterating over the "global all images" generator
 as the secondary continue iterator. The "globalusage" property would be the next
 property to iterate over. It's also possible for multiple sub-properties to be
 iterated over simultaneously, in which case the "continue" token would not have
-a secondary value (e.g. `gaicontinue||`). Occasionally, the ingester will come
+a secondary value (e.g. `gaicontinue||`). ` Occasionally, the ingester will come
 across a piece of media that has many results for the property it's iterating
 over. An example of this can include an item being on many pages, this it would
 have many "global usage" results. In order to process the entire batch, we have
--- a/DAGs.md
+++ b/DAGs.md
@@ -657,7 +657,7 @@ order to process the entire batch, we have to iterate over _all_ of the returned
 results; Wikimedia does not provide a mechanism to "skip to the end" of a batch.
 On numerous occasions, this iteration has been so extensive that the pull media
 task has hit the task's timeout. To avoid this, we limit the number of
-iterations we make for parsing through a sub-property's data.If we hit the
+iterations we make for parsing through a sub-property's data. If we hit the
 limit, we re-issue the original query _without_ requesting properties that
 returned large amounts of data. Unfortunately, this means that we will **not**
 have that property's data for these items the second time around (e.g.
@@ -776,7 +776,7 @@ order to process the entire batch, we have to iterate over _all_ of the returned
 results; Wikimedia does not provide a mechanism to "skip to the end" of a batch.
 On numerous occasions, this iteration has been so extensive that the pull media
 task has hit the task's timeout. To avoid this, we limit the number of
-iterations we make for parsing through a sub-property's data.If we hit the
+iterations we make for parsing through a sub-property's data. If we hit the
 limit, we re-issue the original query _without_ requesting properties that
 returned large amounts of data. Unfortunately, this means that we will **not**
 have that property's data for these items the second time around (e.g.
--- a/DAGs.md
+++ b/DAGs.md
@@ -747,7 +747,7 @@ In this case, we're iterating over the "global all images" generator
 as the secondary continue iterator. The "globalusage" property would be the next
 property to iterate over. It's also possible for multiple sub-properties to be
 iterated over simultaneously, in which case the "continue" token would not have
-a secondary value (e.g. `gaicontinue||`). ` Occasionally, the ingester will come
+a secondary value (e.g. `gaicontinue||`). Occasionally, the ingester will come
 across a piece of media that has many results for the property it's iterating
 over. An example of this can include an item being on many pages, this it would
 have many "global usage" results. In order to process the entire batch, we have
--- a/DAGs.md
+++ b/DAGs.md
@@ -639,22 +639,33 @@ In this case, we're iterating over the "global all images" generator
 as the secondary continue iterator. The "globalusage" property would be the next
 property to iterate over. It's also possible for multiple sub-properties to be
 iterated over simultaneously, in which case the "continue" token would not have
-a secondary value (e.g. `gaicontinue||`). ` Occasionally, the ingester will come
-across a piece of media that has many results for the property it's iterating
-over. An example of this can include an item being on many pages, this it would
-have many "global usage" results. In order to process the entire batch, we have
-to iterate over _all_ of the returned results; Wikimedia does not provide a
-mechanism to "skip to the end" of a batch. On numerous occasions, this iteration
-has been so extensive that the pull media task has hit the task's timeout. To
-avoid this, we limit the number of iterations we make for parsing through a
-sub-property's data.If we hit the limit, we re-issue the original query
-_without_ requesting properties that returned large amounts of data.
-Unfortunately, this means that we will **not** have that property's data for
-these items the second time around (e.g. popularity data if we needed to skip
-global usage). In the case of popularity, especially since the problem with
-these images is that they're so popular, we want to preserve that information
-where possible! So we cache the popularity data from previous iterations and use
-it in subsequent ones if we come across the same item again.
+a secondary value (e.g. `gaicontinue||`).
+
+In most runs, the "continue" key will be `gaicontinue||` after the first
+request, which means that we have more than one batch to iterate over for the
+primary iterator. Some days will have fewer images than the batch limit but
+still have multiple batches of content on the secondary iterator, which means
+the "continue" key may not have a primary iteration component (e.g.
+`||globalusage`). This token can also be seen when the first request has more
+data in the secondary iterator, before we've processed any data on the primary
+iterator.
+
+Occasionally, the ingester will come across a piece of media that has many
+results for the property it's iterating over. An example of this can include an
+item being on many pages, this it would have many "global usage" results. In
+order to process the entire batch, we have to iterate over _all_ of the returned
+results; Wikimedia does not provide a mechanism to "skip to the end" of a batch.
+On numerous occasions, this iteration has been so extensive that the pull media
+task has hit the task's timeout. To avoid this, we limit the number of
+iterations we make for parsing through a sub-property's data.If we hit the
+limit, we re-issue the original query _without_ requesting properties that
+returned large amounts of data. Unfortunately, this means that we will **not**
+have that property's data for these items the second time around (e.g.
+popularity data if we needed to skip global usage). In the case of popularity,
+especially since the problem with these images is that they're so popular, we
+want to preserve that information where possible! So we cache the popularity
+data from previous iterations and use it in subsequent ones if we come across
+the same item again.
 
 Below are some specific references to various properties, with examples for
 cases where they might exceed the limit. Technically, it's feasible for almost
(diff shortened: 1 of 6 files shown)


stacimc (commented):
@AetherUnbound this is incredible 😮 🔥  The work you’ve done here to synthesize external docs across so many sources and clarify this DAG is _fantastic_ and enormously helpful. I’m so excited that you were able to figure out a clear cause of this bug — and although the context is indeed quite hairy, the solution proposed here is elegant and explained beautifully 🙂  Getting the partial data, especially with cached usage data when possible,  strikes the right balance IMO and would be a huge win, especially for reingestion purposes!

The described tests all worked like a charm for me! Ran in ~35 minutes with a global usage count of 1856 for the icon 🎉 

I think I may have found _one_ other case that isn’t covered. Before following the testing instructions to test for `2023-02-01`, I happened to start a regular manual run for today’s date (2023-02-24). I encountered what seems to me like the large batch problem, but _without a `gaicontinue` token_. So, my initial continue token for the bad batch looked like this:

```
{'gucontinue': 'Flag_of_Yemen.svg|arwiki|3736', 'continue': '||imageinfo'}
```

Subsequent continue tokens just change the final portion of the `gucontinue`:

```
{'gucontinue': 'Flag_of_Yemen.svg|arwiki|42380', 'continue': '||imageinfo'}
{'gucontinue': 'Flag_of_Yemen.svg|arwiki|180004', 'continue': '||imageinfo'}
# … and so on
 ```

Because `gaicontinue` is always `None`, the `iteration_counter` never gets incremented and it continues running. I’m not sure how long it would go if I left it, but I added some logs locally and it definitely got much past 100.

I tried adding a check to also increment the counter when `gaicontinue is None and current_gaicontinue is None`. This effectively breaks out of the batch, but then because `adjust_parameters_for_next_iteration` sets the next continue token to the empty `gaicontinue`, in this case that doesn’t “reset” the token to the beginning of the bad batch. So the batch is not re-processed, and if I understand correctly a totally different batch (which may not have the issue) is processed, with the properties unnecessarily reduced.

To be honest I am not sure I understand the circumstances that cause `gaicontinue` to be absent from the continue token here so I’m not sure how best to handle it. What do you think? Apologies if I explained that very poorly -- you should be able to run the DAG for `2023-02-24` to reproduce it, the problem occurs in the first batch.

**Comment by AetherUnbound** [comment]
Ah jeez, thanks for finding that particular case @stacimc! There are _so_ many ways for the Wikimedia API to behave erratically 😞 I'll look into that case and see what can be done, I wasn't expecting a case where the `gaicontinue` token would be absent entirely...

File: DAGs.md [hunk 1]
```diff
@@ -574,23 +574,219 @@ is undocumented.
...
+parameter to avoid this issue on subsequent iterations.
+
+#### `globalusage`
+
+[Docs](https://commons.wikimedia.org/w/api.php?action=help&modules=query%2Bglobalusage)
+
+[Example where an image is used on almost every wiki](https://commons.wikimedia.org/w/index.php?curid=4298234).
+
+For these requests, we can remove the `globalusage` property from the `prop`
+parameter entirely and eschew the popularity data for these items.
```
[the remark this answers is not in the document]
AetherUnbound: Thanks for taking the time to read it all! It was such a maze to try and navigate 😅 😵‍💫

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -23,94 +112,161 @@
...
 
-        for _ in range(MEAN_GLOBAL_USAGE_LIMIT):
+        for _ in range(self.mean_global_usage_limit):
             response_json = super().get_response_json(
                 query_params,
                 timeout=60,
             )
 
             if response_json is None:
+                logger.warning("Response JSON is None")
```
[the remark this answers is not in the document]
AetherUnbound: Oh shoot, that's a really good point. I don't know when this would occur, but it was a case that was handled previously 🤷🏼‍♀️ Maybe we should raise an exception instead so we can see cases where this might occur?

Honestly looking at this more, I don't think this value should _ever_ be `None` when returned from the DelayedRequester, so I think I'll just remove the condition! Perhaps it was vestigial code from before we unified the requester interface 🙂

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -23,94 +112,161 @@
...
+
+                current_gaicontinue = self.continue_token.get("gaicontinue", None)
+                logger.debug(f"{current_gaicontinue=}")
+                if (
+                    current_gaicontinue is not None
+                    and current_gaicontinue == gaicontinue
+                ):
+                    iteration_count += 1
+                else:
+                    gaicontinue = current_gaicontinue
```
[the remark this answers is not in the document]
AetherUnbound: That's a great point! I think so, I'll add that in.

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -23,94 +112,161 @@
...
+
+                if iteration_count >= self.max_page_iteration_before_give_up:
+                    logger.warning(
+                        f"Hit iteration count limit for '{gaicontinue}', "
+                        "re-attempting with a bare token"
+                    )
+                    self.adjust_parameters_for_next_iteration(gaicontinue)
+                    break
```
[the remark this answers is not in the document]
AetherUnbound: That's exactly correct! And based on this flow, the first batch which was halted actually gets written to the TSV, so we end up with two rows in the TSV for each of the results that were part of both attempts. That's part of the necessity around the popularity data cache, the latest row in the TSV has precedence when loading into the table so we want the last processed result to have all the info we want when loaded.

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -136,7 +292,7 @@ def get_media_pages(self, response_json):
 
     def get_record_data(self, record):
         foreign_id = record.get("pageid")
-        logger.debug(f"Processing page ID: {foreign_id}")
+        # logger.debug(f"Processing page ID: {foreign_id}")
```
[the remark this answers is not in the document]
AetherUnbound:
```suggestion
```

tests/dags/providers/provider_api_scripts/test_wikimedia_commons.py
```diff
@@ -126,7 +151,31 @@ def test_get_response_json_returns_correctly_without_continue(monkeypatch):
...
+    wmc.continue_token = {}
+    print("Attempting!")
+    actual = wmc.get_response_json(wmc.get_next_query_params({}))
+    expected = response.copy()
+    expected.pop("continue")
+    assert actual == expected
+    # This should call exactly (max iterations + 1) times
+    assert get_response_mock.call_count == wmc.max_page_iteration_before_give_up + 1
+    # The props should NOT be the default at this point
+    assert wmc.current_props != wmc.default_props
```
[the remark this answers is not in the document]
AetherUnbound: Good call!

AetherUnbound:
_Okay_, I think I've adjusted the ingester to handle the case you found @stacimc! If you run the same steps again, it should now pass that first iteration failure 🙂

stacimc (commented):
🎉 This looks great! I'm still a _little_ confused about how we reset the parameters in the case where `gaicontinue` is None to make sure that the 'bad' batch gets reprocessed, but it seems to work. I think it makes sense if the idea is that this would only happen for the _first_ batch? Is that right?

Everything is now passing locally! I think other than updating the `max_page_iteration`, this looks good to go 💯

AetherUnbound:
> I'm still a _little_ confused about how we reset the parameters in the case where `gaicontinue` is None to make sure that the 'bad' batch gets reprocessed, but it seems to work. I think it makes sense if the idea is that this would only happen for the _first_ batch? Is that right?

That's exactly right! I think that case where there's no `gaicontinue` could only happen on the first batch, in the following cases:
1. There are fewer results than `batch_limit` to process OR
2. Within the first batch, there's a dataset that would require iterating on any of the subprops

AetherUnbound:
>  I feel so much better about our understanding of the DAG.

Even if this doesn't end up being the best solution or we have to change things up down the line, I think _that_ is the biggest benefit I got from all this. I understand this ding-dang API a lot more! 😄

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -23,94 +120,158 @@
...
+
+        # All media info we care about
+        media_all = "url|user|dimensions|extmetadata|mediatype|size|metadata"
+        # Everything minus the metadata, which is only necessary for audio and can
+        # balloon for PDFs which are considered images by Wikimedia
+        media_no_metadata = "url|user|dimensions|extmetadata|mediatype|size"
+
+    # MappingProxy effectively provides a frozen dictionary, since this is defined
+    # at the class level and technically mutable.
+    default_props = MappingProxyType(
```
krysal, line 171: TIL about `MappingProxyType`!

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -1,19 +1,116 @@
...
+request has more data in the secondary iterator, before we've processed any data on
+the primary iterator.
+
+Occasionally, the ingester will come across a piece of media that has many results for
+the property it's iterating over. An example of this can include an item being on many
+pages, this it would have many "global usage" results. In order to process the entire
+batch, we have to iterate over *all* of the returned results; Wikimedia does not provide
+a mechanism to "skip to the end" of a batch. On numerous occasions, this iteration has
+been so extensive that the pull media task has hit the task's timeout. To avoid this,
+we limit the number of iterations we make for parsing through a sub-property's data.If
```
krysal, line 77:
```suggestion
we limit the number of iterations we make for parsing through a sub-property's data. If
```

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -382,16 +569,35 @@ def extract_geo_data(media_data):
...
+    def extract_global_usage(self, media_data) -> int:
+        """
+        Extract the global usage count from the media data.
+        The global usage count serves as a proxy for popularity data, since it
+        represents how many pages across all available wikis a media item is used on.
+        This uses a cache to record previous values for a given foreign ID, because
+        it is possible that we might encounter a media item several times based on the
+        querying method used (see "props" above). Not all results will have the same
+        (if any) usage data, so we want to record and cache the highest value. Most
+        items have no global usage data, so we only cache items that have a value.
```
krysal, line 581: > it is possible that we might encounter a media item several times based on...

Maybe I'm missing something else but does this mean that we get duplicates from the API? 🤔 How are they handled? Also, given this cache is not using Redis or something similar, although it's just for results with global usage data, it seems still a potentially high numbers of registries to save in memory, per the size of Wikimedia images, our second biggest provider (or perhaps now it's the third one after iNatualist?) Anyway, it's worth giving it a try!

krysal (approved):
The warning was not exagerating (❗️) Wikimedia API is quite intricate and elaborated, wow. Impressive work here. Thanks a million for documenting everything 💯

I left some minor non-blocking comments out of curiosity.

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
AetherUnbound: It does not, we filter out duplicates as part of the TSV loading step: 

https://github.com/WordPress/openverse-catalog/blob/3ab5e572b061180250bfee9741d95c813ce8f7f7/openverse_catalog/dags/common/loader/sql.py#L219-L231

This cache is only per-ingestion run, and it looks like our average run is ingesting about ~15k items. So even if every item had popularity data, the cache would only have 15k records, which I think is okay! Definitely within normal running parameters for Python IMO 😄

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
AetherUnbound: Same! I would have used [frozendict](https://pypi.org/project/frozendict/) but I didn't want to introduce a new library; I was glad to find that this was already available 🙂

openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
krysal: Good to know about the limited number of rows cached! However, for what I can tell from that statement, it's not guaranteed that it will keep the row with the highest "global data usage" value. The records are assumed to be identical and take the first one found. So it's not something critical but I wanted to make sure we're all aware of the situation.

AetherUnbound: Ahh you're totally right, it currently just grabs the highest value from whatever's available:

https://github.com/WordPress/openverse-catalog/blob/c476f923c1ace2ab5f0ef40ff552f9516dba49d3/openverse_catalog/dags/providers/provider_api_scripts/wikimedia_commons.py#L586

So there could be a case potentially where one run gathers a _different but smaller_ set of pages which would actually make the total count higher than the maximum value between that and the cache. But already we're throwing out some popularity data for these records since we don't finish processing them 😅 good thing to be aware of for sure!

PR #1009, opened 2023-02-22
Ensure uniqueness of load table names

### LINKED ISSUE #919: Some Metropolitan reingestion days fail due to DuplicateTable
## Description

On each run of the `metropolitan_reingestion_workflow`, a handful of reingestion days fail during the `create_loading_table` step due to the error:

```
psycopg2.errors.DuplicateTable: relation "provider_data_image_metropolitan_museum_reingestion_20221204t00" already exists
```

In each case I checked, this is the first error in the `create_loading_table` task (ie it's not only throwing this on retries).

## Reproduction

See day_shift_2 or day_shift_1 for the 2022-12-04 run in production.

stacimc:
## Fixes

Fixes #919 by @stacimc 

## Description

Some ingestion days for the `metropolitan_museum_reingestion_workflow` have been failing due to `DuplicateTable` errors. This is because the load table names are generated with the pattern: `provider_data_{media_type}_{provider_name}_{timestamp}_{day_shift}` -- but the very long provider name for this DAG causes the table name to be truncated in Postgres, such that the day_shift and part of the timestamp (which are required for uniqueness) are cut off.

This PR changes the order of the component parts of the table name such that `provider_name` is on the _end_, and is therefore the part that will get truncated if it is too long. This guarantees that `media_type`, `timestamp`, `day_shift`, and the first part of the provider name will fit.

Functionally, this means that part of 'reingestion` gets cut off:

```
    # The generated table name on `main`: note that part of the timestamp and the entire dayshift is cut off,
    # meaning that collisions can easily happen
    provider_data_image_metropolitan_museum_reingestion_20230222T20

    # The new table name: note that only the end of the provider name is cut off, and the table name is still
    # guaranteed unique across ingestion days
    provider_data_image_96_20230222T204839_metropolitan_museum_rein
```

Uniqueness is only an issue if we create two DAGs that have very long provider names, which only differ in the final few characters.

### Alternatives considered

1. Give just this particular DAG a shorter, custom DAG id (like `met_museum_reingestion_workflow` or `metropolitan_reingestion_workflow`. The problem could still arise in the future.
2. Removing the `_reingestion` suffix from the load table name. This would work because the 'normal' ingestion flow does not append `day_shift`, so the two DAGs would still be distinct -- but it is less human readable, and we could still encounter the problem with long provider names in the future.

## Testing Instructions

Run the Metropolitan reingestion DAG -- make sure to set a very low INGESTION_LIMIT before you start. Check the logs to make sure the load table names are being generated as expected.

I also ran a few other DAGs, including `wikimedia_commons_workflow` and `wikimedia_reingestion_workflow`.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/loader/sql.py
+++ b/openverse_catalog/dags/common/loader/sql.py
@@ -19,7 +19,7 @@ from psycopg2.errors import InvalidTextRepresentation
 
 logger = logging.getLogger(__name__)
 
-LOAD_TABLE_NAME_STUB = "provider_data_"
+LOAD_TABLE_NAME_STUB = "load_"
 TABLE_NAMES = {AUDIO: AUDIO, IMAGE: IMAGE}
 DB_USER_NAME = "deploy"
 NOW = "NOW()"
--- a/openverse_catalog/dags/providers/provider_dag_factory.py
+++ b/openverse_catalog/dags/providers/provider_dag_factory.py
@@ -120,7 +120,11 @@ def create_ingestion_workflow(
     with TaskGroup(group_id=append_day_shift("ingest_data")) as ingest_data:
         media_type_name = "mixed" if len(conf.media_types) > 1 else conf.media_types[0]
         provider_name = conf.dag_id.replace("_workflow", "")
-        identifier = f"{provider_name}_{{{{ ts_nodash }}}}_{day_shift}"
+
+        # Unique identifier used to generate the load_table name
+        identifier = f"{{{{ ts_nodash }}}}_{provider_name}"
+        if is_reingestion:
+            identifier = f"{day_shift}_{identifier}"
 
         ingestion_kwargs = {
             "ingester_class": conf.ingester_class,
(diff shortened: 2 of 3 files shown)


krysal (approved):
Makes sense. LGTM!

AetherUnbound (approved):
Clever thinking! 

Do you think it might make sense to move the `provider_data_` prefix entirely and favor it for something shorter? I know that'd make this a bigger change, but certainly that feels like a few characters we could easily throw out 😄 If we replaced it with `load_`, that might give us a few more characters before truncation 🙂 Not necessary though, this is good as-is IMO!

stacimc:
@AetherUnbound It's funny that you say that -- my original implementation was literally to change `provider_data_` to `load_` and remove the `_reingestion` suffix 😄 I changed it to the current approach to keep the readability of the suffix and prevent problems down the line with future long DAG names, but we could definitely keep the shortened prefix as well.

AetherUnbound:
> As it happens I've already tested that and it's a tiny change 100

Ha, incredible, and you're right! What a nice, small change 💃🏼

PR #1028, opened 2023-03-01
Improve license URL validation

### LINKED ISSUE #1027: `_get_valid_cc_url` makes a network request even for known valid license urls
## Problem

`_get_valid_cc_url` makes a network request for all license URLs. If, for whatever reason, the CC request fails, this throws an error.

## Description

Add a pre-check for URLs: convert the URL to lower case, add a trailing slash, and check if the URL is in the existing set of valid license URLs. Only make a network request if the URL is not in that set.

## Alternatives

Keep the current solution and hope for all requests to return correct results.

## Additional context

Does declaring the license URL set at the top of the file make sense? Will it be called on every DAG run?

obulat:
## Fixes
Fixes #1027 

## Description
This PR adds a set of valid license URLs to check the URLs before sending a network request to validate it.
It also lower-cases the URL (for `CC0` urls, for example) and adds a trailing slash.

This PR caught an error in the licenses in the Jamendo tests. In the expected license URL, we did not have a trailing slash. 
```
url="https://creativecommons.org/licenses/by-nc/2.0",
raw_url="http://creativecommons.org/licenses/by-nc/2.0/",
```

## Testing instructions
The tests should pass?

### DIFF
--- a/openverse_catalog/dags/common/licenses/licenses.py
+++ b/openverse_catalog/dags/common/licenses/licenses.py
@@ -15,7 +15,8 @@ logger = logging.getLogger(__name__)
 
 LICENSE_PATH_MAP = constants.get_license_path_map()
 REVERSE_LICENSE_PATH_MAP = constants.get_reverse_license_path_map()
-LICENSE_URLS = set(LICENSE_PATH_MAP.keys())
+CC_BASE_URL = "https://creativecommons.org/"
+LICENSE_URLS = {f"{CC_BASE_URL}{l}/" for l in LICENSE_PATH_MAP.keys()}  # noqa: E741
 
 
 class InvalidLicenseURLException(Exception):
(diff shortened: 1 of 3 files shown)


openverse_catalog/dags/common/licenses/licenses.py
```diff
@@ -143,21 +145,29 @@ def _get_valid_cc_url(license_url) -> str | None:
...
     logger.debug(f"Checking license URL {license_url}")
     if type(license_url) != str:
         logger.debug(f"License URL is not a string. Type is {type(license_url)}")
         return
 
     https_url = urls.add_url_scheme(license_url.lower(), "https")
+    if not https_url.endswith("/"):
+        https_url += "/"
+    if https_url in LICENSE_URLS:
+        return https_url
```
AetherUnbound, line 169: Is there an additional test we can add to `test_licenses.py` for this?

AetherUnbound: I was thinking more along the lines of having a test that ensures the `https_url` gets a `/` appended to the end 😄

AetherUnbound (approved):
Awesome!

openverse_catalog/dags/common/licenses/licenses.py
obulat: I was under impression that the licenses code is heavily tested, @AetherUnbound, and did not know that we didn't test for the trailing slash! 😆

dhruvkb (approved):
Looks good, nice to know that it already caught some problems!

Issue #1024, opened 2023-03-01
Improve iNaturalist date check logging


AetherUnbound:
## Description

We should add some logging to the iNaturalist date check step so we can see both the last successful date, the `LastModified` date, and the comparison made in the logs:

https://github.com/WordPress/openverse-catalog/blob/af205795b412da9f3d0d7f6d7e066b7d039b1763/openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py#L182-L201

rwidom:
I'm happy to take this one on, and FWIW my sense is that since August, inaturalist is on a pretty regular monthly schedule as well, with updated files on the 27th, so that there will always be new data on the 1st of the month. For example, the individual latest files we use:

```
% aws s3 ls s3://inaturalist-open-data
                           PRE metadata/
                           PRE photos/
2023-02-27 03:19:19 3851726115 observations.csv.gz
2023-02-27 03:35:20    7070765 observers.csv.gz
2023-02-27 03:33:34 6921196752 photos.csv.gz
2023-02-27 03:35:17   32362994 taxa.csv.gz
```

And then the archived files:

```
% aws s3 ls s3://inaturalist-open-data/metadata/
2021-03-05 02:38:54          0 
2021-04-15 18:16:54 4831598102 inaturalist-open-data-20210407.tar.gz
2021-05-20 03:14:10 5295325724 inaturalist-open-data-20210520.tar.gz
2021-07-13 10:58:49 5843976816 inaturalist-open-data-20210713.tar.gz
2021-08-30 20:35:39 6326004812 inaturalist-open-data-20210830.tar.gz
2021-09-21 10:18:24 6520166168 inaturalist-open-data-20210921.tar.gz
2021-10-21 05:53:32 6771547711 inaturalist-open-data-20211021.tar.gz
2021-11-30 14:28:57 6981491162 inaturalist-open-data-20211121.tar.gz
2022-01-27 06:58:19 7312475433 inaturalist-open-data-20220127.tar.gz
2022-03-10 17:00:45 7471017683 inaturalist-open-data-20220227.tar.gz
2022-04-04 13:31:34 7615780100 inaturalist-open-data-20220327.tar.gz
2022-04-27 03:31:45 7905359916 inaturalist-open-data-20220427.tar.gz
2022-06-01 23:05:30 8326761961 inaturalist-open-data-20220527.tar.gz
[...]

PR #1035, opened 2023-03-04
Add logging to iNaturalist date check

rwidom:
# Fixes

Fixes https://github.com/WordPress/openverse-catalog/issues/1024 by @AetherUnbound (with some extremely timely and helpful mentorship from @AetherUnbound )

# Description

Adds some logging to the function that checks for new data on s3 at the beginning of the dag, and tests for the function that does the checking.

# Testing Instructions

Do `just test`.

Beyond that, in the local environment, "s3" (actually minio) will always have timestamps from whenever you started the container and loaded in the test data for iNaturalist. So, you can test the "no past successful dag runs" case locally, and the "no new data" case locally. (Run the iNaturalist dag once locally and then run it again to confirm that it skips the rest of the dag and notes why in the logs.) 

In theory, you could go into minio and manually delete the test files and re-upload them and then rerun the dag, but I'm pretty sure that doesn't work with the existing Docker config which has a really helpful setup for every other situation. :)

# Template?

For some reason the template isn't coming up for me on this PR, so I'm going to dig around and try to find formatted checklist and certificate of origin, but I do think it's in good shape.

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/inaturalist.py
@@ -189,6 +189,7 @@ class INaturalistDataIngester(ProviderDataIngester):
     ):
         # if it was never run, assume the data is new
         if last_success is None:
+            logger.info("No last success date, assuming iNaturalist data is new.")
             return
         s3 = S3Hook(aws_conn_id=aws_conn_id)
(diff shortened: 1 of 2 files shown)


AetherUnbound (approved):
Looks great, thank you for jumping on this!

stacimc (approved):
Fantastic 🙂

PR #1045, opened 2023-03-13
Use Python to group items by license to speed up the query

### LINKED ISSUE #1044: `add_license_url` DAG is inefficient and fails due to timeout
## Problem

The DAG runs a `SELECT` query for each (license, license version) pair. Each such query lasts at least 30 minutes, while the update query runs for about 2-3 minutes.

## Description

Instead of running one `SELECT` query per license pair, it's better to select all items (identifier, license and license_version) with `NULL` in `meta_data`, and then group them by license in Python and run the corresponding `UPDATE` queries.

## Alternatives

We could also in the first step select all items with `meta_data` is `NULL` (identifier, license and license_version). Then, add the `license_url` to each row using Python, and then in the next step run the update queries for each row individually:
```sql
UPDATE image
SET meta_data=Json({license_url: get_license_info_from_license_pair(license, license_version)})
WHERE identifier=identifier::UUID
```
I'm not sure if running so many individual queries would be faster than the current batch approach.

## Additional context
Details on the run of the previous version of the DAG are here: https://github.com/WordPress/openverse-catalog/pull/1005#issuecomment-1465883307

obulat:
## Fixes
Fixes #1044 

## Description
This PR updates the second step in the DAG to replace many `SELECT` queries with `SELECT`ing all items with `NULL` in meta_data, then grouping of the items to update using Python function, and then running one `UPDATE` query per license pair (if necessary). Hopefully, this should be much faster than the previous run.

This PR also sets the trigger rule for the last step, because otherwise it was skipped if the second step was skipped.

## Testing instruction
Same as #1005.

### DIFF
--- a/openverse_catalog/dags/maintenance/add_license_url.py
+++ b/openverse_catalog/dags/maintenance/add_license_url.py
@@ -99,7 +99,7 @@ def update_license_url(postgres_conn_id: str, task: AbstractOperator) -> str | N
 
     for license_pair, identifiers in records_to_update.items():
         license_, license_version = license_pair.split(",")
-        license_url = get_license_info_from_license_pair(license_, license_version)[-1]
+        *_, license_url = get_license_info_from_license_pair(license_, license_version)
         if license_url is None:
             logger.info(
                 f"No license pair {license_pair} in the license reverse path map."
--- a/openverse_catalog/dags/maintenance/add_license_url.py
+++ b/openverse_catalog/dags/maintenance/add_license_url.py
@@ -9,6 +9,7 @@ This is a maintenance DAG that should be run once. If all the null values in
 the `meta_data` column are updated, the DAG will only run the first and the
 last step, logging the statistics.
 """
+import json
 import logging
 from datetime import timedelta
 from textwrap import dedent
@@ -17,8 +18,9 @@ from typing import Literal
 from airflow.models import DAG
 from airflow.models.abstractoperator import AbstractOperator
 from airflow.operators.python import BranchPythonOperator, PythonOperator
+from airflow.utils.trigger_rule import TriggerRule
 from common.constants import DAG_DEFAULT_ARGS, POSTGRES_CONN_ID, XCOM_PULL_TEMPLATE
-from common.licenses.constants import get_reverse_license_path_map
+from common.licenses import get_license_info_from_license_pair
 from common.loader.sql import RETURN_ROW_COUNT
 from common.slack import send_message
 from common.sql import PostgresHook
@@ -65,7 +67,7 @@ def get_statistics(
     return next_task
 
 
-def update_license_url(postgres_conn_id: str, task: AbstractOperator) -> dict[str, int]:
+def update_license_url(postgres_conn_id: str, task: AbstractOperator) -> str | None:
     """Add license_url to meta_data batching all records with the same license.
     :param task: automatically passed by Airflow, used to set the execution timeout
     :param postgres_conn_id: Postgres connection id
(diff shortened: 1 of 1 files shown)


openverse_catalog/dags/maintenance/add_license_url.py
```diff
@@ -76,60 +78,85 @@ def update_license_url(postgres_conn_id: str, task: AbstractOperator) -> dict[st
...
+    # Dictionary with license pair as key and list of identifiers as value
+    records_to_update = {}
+
+    for result in records_with_null_in_metadata:
+        identifier, license_, version = result
+        license_pair = f"{license_},{version}"
+        if license_pair not in records_to_update:
+            records_to_update[license_pair] = [identifier]
+        else:
+            records_to_update[license_pair].append(identifier)
```
AetherUnbound, line 98: This could be replaced with a [`defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict)!

```suggestion
    records_to_update = defaultdict(list)

    for result in records_with_null_in_metadata:
        identifier, license_, version = result
        license_pair = f"{license_},{version}"
        records_to_update[license_pair].append(identifier)
```

openverse_catalog/dags/maintenance/add_license_url.py
```diff
@@ -76,60 +78,85 @@ def update_license_url(postgres_conn_id: str, task: AbstractOperator) -> dict[st
...
+
+    total_updated = 0
+    updated_by_license = {}
+
+    for license_pair, identifiers in records_to_update.items():
+        license_, license_version = license_pair.split(",")
```
AetherUnbound, line 104: If we're saving this as a string only to split it out again, we could probably use the license & license version pair in tuple form as the key itself, e.g. `records_to_update[license_, license_version]`. Then we can unpack it directly here too, e.g. `for (license_, license_version), identifiers in records_to_update.items()`.

openverse_catalog/dags/maintenance/add_license_url.py
```diff
@@ -76,60 +78,85 @@ def update_license_url(postgres_conn_id: str, task: AbstractOperator) -> dict[st
...
-
-    logger.info(f"{total_count} image records with missing license_url updated.")
-    for license_url, count in total_counts.items():
-        logger.info(f"{count} records with {license_url}.")
-    return total_counts
+        if updated_count:
+            updated_by_license[license_url] = updated_count
+        total_updated += updated_count
+    logger.info(f"Updated {total_updated} rows")
+    return json.dumps(updated_by_license)
```
AetherUnbound, line 128: You should be able to return this directly without dumping it as a JSON string before hand! You'll just need to set `render_template_as_native_obj=True,` in the DAG settings ([here's an example](https://github.com/WordPress/openverse-catalog/blob/af205795b412da9f3d0d7f6d7e066b7d039b1763/openverse_catalog/dags/maintenance/pr_review_reminders/pr_review_reminders_dag.py#L52)). Then you can do a standard `None` check/use the dictionary directly in the `final_report` task below 🙂

AetherUnbound (commented):
I've got a few comments for format improvements, but I was able to test this locally and it ran great!

openverse_catalog/dags/maintenance/add_license_url.py
obulat: Nice, thank you!

openverse_catalog/dags/maintenance/add_license_url.py
obulat: Thank you for the suggestion, that is so much better :)

AetherUnbound (approved):
Looks and works great!

stacimc (approved):
Apologies for the late review -- this looks excellent!

I have been doing Flickr testing locally and happened to have ~1.55 million records locally 😅 I set the `meta_data` to null for all of them and the DAG ran successfully in just over six minutes :) 

Screen Shot 2023-03-15 at 11 49 43 AM

AetherUnbound:
Heh, what a great test set to have! 😄

obulat:
> Apologies for the late review -- this looks excellent!
> 
> I have been doing Flickr testing locally and happened to have ~1.55 million records locally 😅 I set the `meta_data` to null for all of them and the DAG ran successfully in just over six minutes :)
> 
> Screen Shot 2023-03-15 at 11 49 43 AM

Thank you for your testing, it's more than what I could hope for, @stacimc  😆 I would not have guessed that the most popular license would be by-nd-nc/2.0/jp

obulat:
I ran the DAG in prod, and this is the log output:
```
[2023-03-16, 03:45:26 UTC] {slack.py:321} INFO - 
`add_license_url` DAG run completed.
Update statistics:
https://creativecommons.org/licenses/by-nc-sa/3.0/: 6318 rows
https://creativecommons.org/licenses/by-nd/3.0/: 201 rows
https://creativecommons.org/licenses/by-sa/3.0/: 1545 rows
https://creativecommons.org/licenses/by-nc/3.0/: 2607 rows
https://creativecommons.org/publicdomain/zero/1.0/: 2919 rows
https://creativecommons.org/publicdomain/mark/1.0/: 374 rows
https://creativecommons.org/licenses/by-nc-sa/1.0/: 166 rows
https://creativecommons.org/licenses/by-nc-sa/2.0/: 2165 rows
https://creativecommons.org/licenses/by-nc-sa/2.5/: 14 rows
https://creativecommons.org/licenses/by-sa/2.5/: 25 rows
https://creativecommons.org/licenses/by-sa/2.0/: 213 rows
https://creativecommons.org/licenses/by-nc/4.0/: 26 rows
https://creativecommons.org/licenses/by-sa/4.0/: 7 rows
https://creativecommons.org/licenses/by-nd-nc/2.0/jp/: 98 rows
https://creativecommons.org/licenses/by-nd/2.0/: 16 rows
https://creativecommons.org/licenses/by-sa/1.0/: 1 rows

Now, there are 9382 records with NULL meta_data left.
```
I suspect that the records that still have NULL in meta_data have NULL in the license_version field.

Issue #1007, opened 2023-02-18
Adjust Flickr reingestion schedule


stacimc:
## Problem

Per the discussion in #995, we will soon be enabling the Flickr DAG but expect that the current implementation will result in some unique results being skipped during ingestion. To combat this, we want to increase the frequency of reingestion so that each day gets reingested more frequently and we can expect greater coverage.

## Description

At minimum, we should update the Flickr reingestion workflow to run on a `@daily` schedule (it is currently `@weekly`). **We should wait until the current implementation has been run successfully a few times, and we have data that shows it can complete in under 24 hours.**

Depending on how slow or fast the reingestion DAG takes, we may also be able to increase the number of reingestion days.

## Alternatives

Alongside this effort, we are also planning on exploring other approaches to modifying the Flickr DAG to avoid the problems with duplicates and missing records. This work is meant to improve the DAG in the meantime.

stacimc:
Reingestion has been enabled in production. It is on a `@weekly` schedule at the moment. I'll adjust parameters after collecting some data about how this initial run goes. At minimum I think the schedule will be updated and I suspect `max_active_tasks` can be updated, once I get a sense of how close we are to the rate limit.

stacimc:
This is briefly blocked (and reingestion is disabled) while we ensure that reingestion does not cause rate limiting issues.

PR #1047, opened 2023-03-14
Update Flickr large batch handling

stacimc:
## Fixes

Fast follow for Flickr backfill.
Related to #1007.

## Description

This PR attempts to handle situations where the Flickr API returns excessively large batches of data. The logic should be documented pretty thoroughly in the code, but here's a refresher:

The Flickr API will only return 4,000 unique records for any given set of query params; after that, it will just infinitely return duplicates. Consequently, we have to try to query the API in such a way that we get batches with less than 4,000 records each. Up until now, we have been doing this using the `TimeDelineatedProviderDataIngester` to break the day into small time intervals for ingestion.

There is a limit to the granularity of data by time interval, though -- once you reduce the interval size to about 5 minutes, the number of records stays the same. For example, given a 5-min interval with > 4k records, if you search _any **5 second** interval within this range, you will still be returned >4k records_. So reducing the timespan only works up to a certain point. However, we can still try to reduce the size of the result set by querying for one license type at a time, instead of all 8 license types at once.

This PR detects these large batch intervals during ingestion, adds them to an array for later processing, and skips ingestion for that batch. After 'regular' ingestion completes, each of these large intervals are reprocessed 8 times, once for each license type. It's still possible for a 5-min interval to contain more than 4k records for a single license type, but in this case there's nothing more we can do, so we process the first 4,000 results **and then continue.**

### Notes
* It is impossible to guarantee that we will get all records, but this should dramatically increase what we're able to ingest.
* Notably, this situation is _most likely to arise on ingestion days with a lot of data_, so making this improvement will result in a large data increase.
* After merging this update, I'll re-run the failed days in production.

## Testing Instructions

Try running the flickr DAG locally. In particular, run it for one of the days that failed in production using the DagRun conf options.

I tried a manual run with the conf:
```
{
    "date":"2023-02-26"
}
```

This day failed in production after ingesting **7,253** records. Tested locally against this branch, the run succeeded after about 10 minutes and ingested **56,927** records.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/flickr.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/flickr.py
@@ -15,7 +15,6 @@ import logging
 from datetime import datetime, timedelta
 
 import lxml.html as html
-from airflow.exceptions import AirflowException
 from airflow.models import Variable
 from common import constants
 from common.licenses import get_license_info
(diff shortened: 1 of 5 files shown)


openverse_catalog/dags/providers/provider_api_scripts/time_delineated_provider_data_ingester.py
```diff
@@ -235,9 +250,9 @@ def get_should_continue(self, response_json) -> bool:
                 " been fetched. Consider reducing the ingestion interval."
             )
             if self.should_raise_error:
-                raise AirflowException(error_message)
+                raise Exception(error_message)
```
stacimc, line 253: It's necessary to change this to an `Exception`, because `AirflowException`s are specifically _not handled_ by our ingestion error handling logic (meaning they can never be skipped).

openverse_catalog/dags/providers/provider_api_scripts/flickr.py
```diff
@@ -129,6 +181,32 @@ def get_batch_data(self, response_json):
...
         if response_json is None or response_json.get("stat") != "ok":
             return None
+
+        # Detect if this batch has more than the max_unique_records.
+        detected_count = self.get_record_count_from_response(response_json)
+        if detected_count > self.max_unique_records:
+            logger.error(
+                f"{detected_count} records retrieved, but there is a"
+                f" limit of {self.max_unique_records}."
+            )
```
zackkrida, line 191: This log was so useful when observing this PR in airflow. Before even reading the code it gave a great understanding of what was happening 👍

openverse_catalog/dags/providers/provider_api_scripts/time_delineated_provider_data_ingester.py
zackkrida: @stacimc Is this feedback worth adding to the code?

```
# Must be an `Exception` instead of an `AirflowException` to allow skipping.
```

Or something like that.

zackkrida (approved):
Really cool. On my local run I picked up 46,732 records. As always, phenomenal documentation 😍

obulat (approved):
This is such a clever idea! Hope we can get most of the images now.
Thank you for amazing documentation and tests.

openverse_catalog/dags/providers/provider_api_scripts/time_delineated_provider_data_ingester.py
AetherUnbound: Or potentially even a custom, `Exception`-derived exception subclass?

AetherUnbound (approved):
Incredible documentation, thanks for getting this all down in the various docstrings/comments! I hope this will be useful for other folks who encounter issues with the Flickr API 😄

openverse_catalog/dags/providers/provider_api_scripts/time_delineated_provider_data_ingester.py
stacimc: Our own error handling in the super class will [re-raise this](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py#L250) as an `IngestionError` already.

This isn't a problem with this specific exception so much as a general point about error handling in the ingester classes: we probably shouldn't be raising `AirflowException`s ourselves anyway.

PR #1049, opened 2023-03-16
Handle the upper case licenses in the add_license_dag

obulat:
## Description

After the last run of the `add_license_url` DAG, we still have more than 8000 records with `NULL` in `meta_data`. Most of them are due to the license name being upper case (CC0 instead of cc0).
This PR converts the license name to lower case before using it to create the URL. It also updates both the `meta_data` and the `license`.

To make this DAG faster, I've removed the first step, which basically just repeats the first part of the second task selecting the items where meta_data is NULL.

This DAG also saves the data (identifier, license, license_version) for invalid license pairs that cannot be fixed to S3 for further investigation.

### DIFF
--- a/openverse_catalog/dags/maintenance/add_license_url.py
+++ b/openverse_catalog/dags/maintenance/add_license_url.py
@@ -9,6 +9,7 @@ This is a maintenance DAG that should be run once. If all the null values in
 the `meta_data` column are updated, the DAG will only run the first and the
 last step, logging the statistics.
 """
+import json
 import logging
 from collections import defaultdict
 from datetime import timedelta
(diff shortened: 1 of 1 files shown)


openverse_catalog/dags/maintenance/add_license_url.py
```diff
@@ -118,9 +126,37 @@ def update_license_url(postgres_conn_id: str, task: AbstractOperator) -> dict[st
...
+    s3_key = "invalid_items.tsv"
+    tmp_path = Path("/tmp") / s3_key
+
+    with open(tmp_path, "w+") as f:
+        tsv_writer = csv.DictWriter(
+            f, delimiter="\t", fieldnames=["license", "license_version", "identifier"]
+        )
+        tsv_writer.writeheader()
+        for item in invalid_items:
+            tsv_writer.writerow(item)
```
AetherUnbound, line 153: We could potentially use `tempfile` here to ensure that the TSV is deleted in case anything goes wrong: https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

AetherUnbound (approved):
LGTM! One suggestion but not a blocking one.

stacimc (approved):
LGTM :)

PR #1059, opened 2023-03-23
JSON retrieval function across all provider_api_scripts tests

### LINKED ISSUE #712: Standardize JSON retrieval function across all tests
## Description

Most of our provider tests use stored JSON blobs as testing artifacts. Each test has its own implementation of a `_get_resource_json` (or similar) function, which just loads & deserializes the JSON blob by name.

We should create a shared utility function for this, and replace all bespoke functions within each test with that function.

The only complication here is that we'd like to address each resource by name rather than full path (e.g. `metadata.json` rather than `../resources/<provider>/metadata.json`). This could be mitigated by assuming that the provider's name is the folder name for the resource. We'll just need to make sure that is consistent for each of the providers.

## Additional context

Inspired by @obulat's comment here: https://github.com/WordPress/openverse-catalog/pull/701#discussion_r959513517

## Implementation

Wbrown633:
## Fixes

Fixes #712 by @AetherUnbound 

## Description

Replace multiple bespoke `load_json` type functions throughout the `api_provider_scripts` test directory with a single function using path lib. Function takes the folder name and the resource name. Then in each test file this function is imported and set up to always refer to the folder specific to that test file. 

Making a draft PR as this is my first code contribution. Not sure that the design of this solution is quite right or where this function should live. Adding an entire new file for one utility didn't feel efficient but I didn't know where else it belonged. Appreciate any feedback. 

## Testing Instructions

Confirm all tests are still passing.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


tests/dags/providers/provider_api_scripts/resources/JsonLoad.py
AetherUnbound, line 1: I appreciate that you followed the naming convention for this folder (based on `FakeSource.py`)! However, we typically try to avoid camel case module names in python, would you mind renaming this to `json_load.py`? It also looks like we're not using `FakeSource.py` anywhere 🤔 do you mind deleting it as part of this PR?

tests/dags/providers/provider_api_scripts/resources/JsonLoad.py
```diff
@@ -0,0 +1,8 @@
+import json
+from pathlib import Path
+
+
+def get_resource_json(folder_name, json_name):
+
+    RESOURCES = Path(__file__).parent / folder_name
+    return json.loads((RESOURCES / json_name).read_text())
```
AetherUnbound, line 8: While the `RESOURCES` variable was caps-case for many of the now-removed functions, we typically reserve usage of caps-case to global variables. This can be lower case 🙂 
```suggestion
    resources = Path(__file__).parent / folder_name
    return json.loads((resources / json_name).read_text())
```

tests/dags/providers/provider_api_scripts/resources/JsonLoad.py
```diff
@@ -0,0 +1,8 @@
+import json
+from pathlib import Path
+
+
+def get_resource_json(folder_name, json_name):
```
AetherUnbound, line 5: One thing I'm seeing in a lot of the files is that we're still retaining a `_get_resource_json` function which merely calls this function with the resource folder. This might be a bit tricky, but one way we could improve this would be to create a higher-order function which _generates_ another function. This could look something like:

```python
import json
from pathlib import Path
from typing import Callable


def make_resource_json_func(folder_name: str) -> Callable[[str], Path]:
    resources = Path(__file__).parent / folder_name

    def get_resource_json(resource_name: str) -> Path:
        return json.loads((resources / resource_name).read_text())

    return get_resource_json
```

Then, in the actual test functions, we can replace the `_get_resource_json` declaration with something like this:

```python
_get_resource_json = make_resource_json_func("brooklynmuseum")
```

That would make `_get_resource_json` within a given testing module behave _exactly_ as it had before without any additional test changes, _and_ we get to remove the `_get_resource_json` declarations in each of the modules too! If that change makes sense (happy to describe it a bit more too), do you think you'd be willing to make that for this PR?

AetherUnbound (commented):
Congrats on the first-time code contribution!! It is fantastic to see so much consolidation happening here 🥳 

I have some suggestions for improvements that I think will help reduce duplicated code even further. Let us know if there is any part of that suggestion that doesn't make sense or could use more explanation. I'm really excited to see this one merged! 😄

tests/dags/providers/provider_api_scripts/resources/JsonLoad.py
Wbrown633: I need a mind blown reaction lol! Yes this makes perfect sense and is the type of solution I was reaching for when I was implementing. I'll work on that this weekend. 

Are type hints encouraged for this project? I usually use them but I didn't notice them in the files I was working on so I avoided them.

obulat: Yes, @Wbrown633, we encourage the use of type hints!
You can see a good example in `openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py`

AetherUnbound: Fantastic! I'm glad that made sense, feel free to ask us if you have any questions 😄 

Type hints are definitely encouraged, and that's the nice thing about Python - while much of this code hasn't been touched in a bit, we can add type hints to the code we _do_ add/modify and don't need to have 100% coverage if we start adding it 🙂 in other words, yes please!

tests/dags/providers/provider_api_scripts/resources/json_load.py
```diff
@@ -0,0 +1,31 @@
...
+
+def make_resource_json_func(folder_name: str) -> Callable[[str], dict]:
+    """
+    Given the name of a folder inside the resources folder returns
+    a function that takes a json file name within the given folder
+    and returns the contents of that json file as a dictionary.
+
+    Args:
+        folder_name (str): name of a folder inside the resources folder
+        (tests/dags/providers/provider_api_scripts/resources/[folder_name]).
```
AetherUnbound, line 14: Love this!

tests/dags/providers/provider_api_scripts/test_brooklyn_museum.py
```diff
@@ -55,6 +52,12 @@ def test_get_data_from_response(resource_name, expected):
     assert actual == expected
 
 
+def test_docstring_example():
+    json_func_ex = make_resource_json_func("brooklynmuseum")
+    json_dict = json_func_ex("cc_license_info.json")
+    assert json_dict["public_name"] == "Creative Commons-BY"
+
+
```
AetherUnbound, line 60: Is this still necessary? Should we perhaps have a separate test file specifically for testing the `make_resourse_json_func` function?

AetherUnbound (commented):
This is excellent, what a great simplification! It's really surprising to see how many times we were reusing the same set of steps 😅 I have one note, but beyond that this is almost there!

tests/dags/providers/provider_api_scripts/test_brooklyn_museum.py
Wbrown633: Yeah splitting out makes sense. I really only added this test because I found myself writing out the logic for a test as an example in the docstring, and then having an entire test that lives in a doc string and not the actual test suite didn't feel right.

stacimc (approved):
Looks good to me, and all tests pass locally 👍 Thanks @Wbrown633!

PR #1060, opened 2023-03-24
Update PhyloPic DAG to use API v2

### LINKED ISSUE #998: Update Phylopic to use v2 API
## Description

Our Phylopic DAG began failing last night: https://airflow.openverse.engineering/log?execution_date=2023-02-12T00%3A00%3A00%2B00%3A00&task_id=ingest_data.pull_image_data&dag_id=phylopic_workflow&map_index=-1

It appears the API URL we were trying to use now 404s: http://phylopic.org/api/a/image/list/modified/2023-02-12/2023-02-13

While I can't find any specific release information, the API now [appears to be on v2](http://api-docs.phylopic.org/v2/). This is confirmed a bit by [the recent tagging and merging of the v2.0.0 branch into a production branch](https://github.com/keesey/phylopic/commit/cb5a77896f6c39f2efc37f684db094478ad253fd) on the phylopic repo.

I suspect we'll need to disable this DAG until we have an opportunity to update the URLs.

krysal:
## Fixes

Fixes WordPress/openverse#902 🚨
Fixes #998 

## Description

This PR updates PhyloPic to use v2 of the API, and with this, it's now converted to a non-dated DAG since v2 doesn't support a filter by modified date _yet_. The good news is that getting the fields is much easier and should run faster since we can get more items per request.


AetherUnbound:
Thanks for jumping on this @krysal!! We actually got a response back from Mike Keesey in the upstream ticket I filed, looks like we might have some new endpoints available: https://github.com/keesey/phylopic/issues/5#issuecomment-1478078566

I'll be responding to their comment today 😄

krysal:
I read it! While that gets shipped, this update will enable us to reactivate PhyloPic after a run :)

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -23,202 +23,114 @@
 logger = logging.getLogger(__name__)
 
 
+def save_response_content(response: dict, filename: str) -> None:
```
obulat, line 26: Was this function used for testing? Seems that it's not used anywhere

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -23,202 +23,114 @@
...
+            "embed_items": "true",
+        }
 
-    @staticmethod
-    def _get_response_data(response_json) -> dict | list | None:
-        """Intermediate method for pulling out results from a Phylopic API request."""
-        if response_json and response_json.get("success") is True:
-            return response_json.get("result")
+    def get_should_continue(self, response_json):
+        return self.current_page < self.total_pages
```
obulat, line 71:
```suggestion
        return self.current_page <= self.total_pages
```
This DAG ran really well until the last page, when the last requested page returned 404:
```
[2023-03-27, 15:45:24 +03] {requester.py:78} WARNING - Unable to request URL: https://api.phylopic.org/images?build=196&page=146&embed_items=true  Status code: 404
[2023-03-27, 15:45:24 +03] {requester.py:148} WARNING - Bad response_json:  None
[2023-03-27, 15:45:24 +03] {requester.py:149} WARNING - Retrying:
_get_response_json(
    https://api.phylopic.org/images,
    {'build': 196, 'page': 146, 'embed_items': 'true'},
    retries=-1)
[2023-03-27, 15:45:24 +03] {requester.py:134} ERROR - No retries remaining.  Failure.
[2023-03-27, 15:45:24 +03] {media.py:230} INFO - Writing 74 lines from buffer to disk.
[2023-03-27, 15:45:24 +03] {provider_data_ingester.py:501} INFO - Committed 6974 records
[2023-03-27, 15:45:24 +03] {taskinstance.py:1768} ERROR - Task failed with exception
providers.provider_api_scripts.provider_data_ingester.IngestionError: Retries exceeded
query_params: {"build": 196, "page": 146, "embed_items": "true"}
```

The `initial_request.json`(tests/dags/providers/provider_api_scripts/resources/phylopic/initial_request.json) seems to suggest that we should use `<=` or even `==` here, because even thought the totalPages is 145, the lastPage is 144.
```
  "lastPage": {
    "href": "/images?build=194&page=144"
  },
  "totalPages": 145
```

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -23,202 +23,114 @@
...
-        foreign_url = self._image_url(uid)
-
-        (
-            creator,
-            meta_data["credit_line"],
-            meta_data["pub_date"],
-        ) = self._get_creator_details(result)
+        title = data.get("self", {}).get("title")
+        creator = data.get("contributor", {}).get("title")
+        creator_url = self.host + data.get("contributor", {}).get("href")
```
obulat, line 108: This could return an invalid url if `contributor` is `None`
```suggestion
        creator_path = data.get("contributor", {}).get("href")
        creator_url = self.host + creator_path if creator_path else None
```

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -23,202 +23,114 @@
...
             "width": width,
             "height": height,
-            "creator": creator,
-            "title": title,
-            "meta_data": meta_data,
+            # TODO: Evaluate whether to include upstream thumbnails.
+            # Sizes available: 192x192, 128x128, 64x64.
+            # "thumbnail": thumbnail,
+            # TODO: Evaluate whether to include nodes' titles as tags.
+            # "tags": tags,
```
obulat, line 125: I think we could re-use the Taxonomy code that @rwidom wrote for iNaturalist here. The titles in phylopic are the Latin names, right?
Not for this PR to implement, but maybe a TODO comment we could add here?

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -23,202 +23,114 @@
...
+            "title": title,
+            "creator": creator,
+            "creator_url": creator_url,
             "width": width,
             "height": height,
-            "creator": creator,
-            "title": title,
-            "meta_data": meta_data,
+            # TODO: Evaluate whether to include upstream thumbnails.
+            # Sizes available: 192x192, 128x128, 64x64.
```
obulat, line 122: I tried the other images we get in the `sample_record.json`: while the SVGs are all crisp on any resolution, the raster image of an acceptable size (https://images.phylopic.org/images/5b1e88b5-159d-495d-b8cb-04f9e28d2f02/raster/512x480.png) looks much worse. 

I wonder if we should handle SVGs differently: they can be basically any size, their width and height don't mean much beyond the aspect ratio (which of course is very important for the frontend rendering), and their thumbnails will probably be larger in size and worse in quality than the original image. Not for this PR specifically, of course.

obulat (changes requested):
Thank you for the quick fix of a ❗ critical issue, @krysal!

I will approve after the `creator_url` and last page are fixed.

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
krysal: I'm not sure the code is reusable in this case because iNaturalist is processed in SQL but the logic could be similar.

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
krysal: Yeah, the raster images are an option but look really blurry, not sure if it's worth including those for thumbnails.

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
obulat: Oh wow, I haven't seen the iNaturalist code before! It's awesome (and I'm afraid I wouldn't be able to use something similar here). No TODO then :) 
But it would be nice to open an issue for adding Vernacular names to Phylopic metadata, to make the images more discoverable.

AetherUnbound:
@krysal it seems the old parameters have been re-added to the new API 😮 https://github.com/keesey/phylopic/issues/5. How would you like to proceed? I'm happy to review this if you'd like to get it in and follow up with making this a dated DAG once more, or we could pivot to just updating the dated portions of the existing DAG.

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -25,200 +23,112 @@
...
+
+    def ingest_records(self):
+        self._get_initial_query_params()
+        super().ingest_records()
+
+    def _get_initial_query_params(self) -> None:
+        """Get the required `build` param from the API and set the total pages."""
+        resp = self.get_response_json(query_params={})
+        if not resp:
+            raise Exception("No response from Phylopic API.")
```
AetherUnbound, line 44: Maybe this might be a more specific exception?
```suggestion
            raise ValueError("No response from Phylopic API.")
```

AetherUnbound (approved):
Thanks again for jumping on this so fast! Code-wise, I just have a few non-blocking comments.

On the data front though, it looks like this will give us an entirely new set of Phylopic data (rather than updating existing values). I was able to test this by:
1. Download production phylopic data: `\copy (SELECT * FROM image WHERE provider='phylopic') to '/tmp/phylopic_v1_prod.csv' DELIMITER ',' CSV HEADER`. This returned **4,156 rows**
2. Start the catalog stack (I had to rebuild the postgres image with a newer version of pgcli, PR coming out for that in a little bit)
3. Copy the file into the postgres image: `docker cp /tmp/phylopic_v1_prod.csv openverse-catalog-postgres-1:/tmp/phylopic_v1_prod.csv`
4. Load the file into the table using `j db-shell`: `\copy image from '/tmp/phylopic_v1_prod.csv' WITH (FORMAT CSV, HEADER)`
5. Run the full DAG, which processed **6,974 rows**.
6. Ran a count using `j db-shell`:

```sql
openledger> select count(*) from image;
+-------+
| count |
|-------|
| 10521 |
+-------+
SELECT 1
Time: 0.040s
```

All that to say that we will be losing the old data, which I think is fine because it's not working anyway 😅 We could explicitly delete the existing Phylopic data from the catalog prior to running this DAG though, that might be the best bet to ensure we don't have records hanging around that can't be accessed.

openverse_catalog/dags/providers/provider_api_scripts/phylopic.py
```diff
@@ -25,200 +23,112 @@
...
-        license_url = result.get("licenseURL")
-
-        img_url, width, height = self._get_image_info(result, uid)
-
-        if img_url is None:
+        data = data.get("_links", {})
+        license_url = data.get("license", {}).get("href")
+        img_url = data.get("sourceFile", {}).get("href")
+        foreign_url = data.get("self", {}).get("href")
+        if not license_url or not img_url or not foreign_url:
```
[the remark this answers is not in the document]
krysal: I find the slightly longer form clearer in this case 😄

krysal:
Good find! Unfortunately, it seems some olders rows have `foreign_identifier` in the form of the direct URL that no longer works :/ but seems it's possible to save them extracting the UUID from the URL. I think that will require an extra DAG to run a one time pass through PhyloPic to fix those values. What do you think? I can work on that, so we don't start PhyloPic DAG until that is done.

AetherUnbound:
Honestly, it's only 4k records 😅 I think it would be okay to move them all, especially since they _were_ changed on the upstream site (and since this precludes our ability to have Phylopic displaying on the API I think). But I'm fine either way!

PR #1066, opened 2023-03-24
Add User Agent string as default header

krysal:
## Fixes
Fixes #869 by @AetherUnbound


AetherUnbound:
We're going to close this for now while we work on the monorepo migration effort in #1094 - we'll re-open when that work is complete!

PR #1074, opened 2023-03-29
Create DAG to fix PhyloPic's `foreign_identifier` column

### LINKED ISSUE #1073: Update PhyloPic's `foreign_identifier` field
## Description

While reviewing https://github.com/WordPress/openverse-catalog/pull/1060#pullrequestreview-1359810476, @AetherUnbound discovered that most of the PhyloPic links would break if we run the DAG for the API v2, due to the change in the field selected as `foreign_identifier`. We could avoid this given this field is present in the old URL used too.

## Reproduction
See https://github.com/WordPress/openverse-catalog/pull/1060#pullrequestreview-1359810476

krysal:
## Fixes
Fixes #1073 by @krysal 
Related to #1060

## Description
This PR adds a one-time-run DAG to update this field for Phylopic images, which extract the uuid from the old URL and updates the `foreign_identifier` with that value, so the v2 of the provider DAG can update the rows instead of creating new ones.

## Test
This can actually be tested with production data following the steps @AetherUnbound used in https://github.com/WordPress/openverse-catalog/pull/1060#pullrequestreview-1359810476.

1. Upload production PhyloPic data. If you have permission, you can follow the steps cited in the previous (review) link or download the data [here](https://gist.github.com/krysal/0721e8109338957d5fce0ea1904c5818/raw/bfe88278ba7d0804f9ff1b1533c3639bd8f01975/phylopic_v1_prod.csv).
2. Checkout this branch and run the **update_phylopic_foreign_identifier** DAG
3. Check in the DB that images have been updated. Some images may still have thumbnails because they were skipped.
4. (optional) Try looking for a random image in the browser concatenating `https://www.phylopic.org/images/<foreign_identifier>`

### DIFF
--- a/openverse_catalog/dags/maintenance/fix_phylopic_foreign_identifier.py
+++ b/openverse_catalog/dags/maintenance/fix_phylopic_foreign_identifier.py
@@ -83,7 +83,7 @@ def update_foreign_identifiers(task: AbstractOperator) -> dict[str, int]:
     return counter
 
 
-def final_report(counter: dict[str, int]) -> None:
+def final_report(counter) -> None:
     message = f"{DAG_ID} DAG run completed. Update statistics:\n{counter}."
     send_message(message, dag_id=DAG_ID)
 
--- a/openverse_catalog/dags/maintenance/fix_phylopic_foreign_identifier.py
+++ b/openverse_catalog/dags/maintenance/fix_phylopic_foreign_identifier.py
@@ -4,8 +4,10 @@ One-time run DAG to fix the foreign identifier for PhyloPic images.
 In order to prevent broken links, we need to update the foreign identifier from using
 the image URL to using the foreign image UUID.
 """
+
 import logging
 import re
+from datetime import timedelta
 from textwrap import dedent
 
 from airflow.decorators import dag
@@ -91,7 +93,7 @@ def final_report(counter: dict[str, int]) -> None:
     default_args={
         **DAG_DEFAULT_ARGS,
         "retries": 0,
-        # "execution_timeout": timedelta(hours=5),
+        "execution_timeout": timedelta(hours=5),
     },
     schedule=None,
     catchup=False,
--- a/openverse_catalog/dags/maintenance/fix_phylopic_foreign_identifier.py
+++ b/openverse_catalog/dags/maintenance/fix_phylopic_foreign_identifier.py
@@ -35,6 +35,7 @@ def update_foreign_identifiers(task: AbstractOperator) -> dict[str, int]:
     uuid_pattern = (
         "[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[089ab][0-9a-f]{3}-[0-9a-f]{12}"
     )
+    prog = re.compile(uuid_pattern)
     counter = {
         "updated": 0,
         "skipped": 0,
@@ -49,7 +50,7 @@ def update_foreign_identifiers(task: AbstractOperator) -> dict[str, int]:
             counter["skipped"] += 1
             continue
 
-        uuid = re.search(uuid_pattern, foreign_identifier).group()
+        uuid = prog.search(foreign_identifier).group()
         if uuid is None:
             counter["failed_none"] += 1
             continue
--- a/openverse_catalog/dags/maintenance/fix_phylopic_foreign_identifier.py
+++ b/openverse_catalog/dags/maintenance/fix_phylopic_foreign_identifier.py
@@ -1,0 +1,116 @@
+"""
+One-time run DAG to fix the foreign identifier for PhyloPic images.
+
+In order to prevent broken links, we need to update the foreign identifier from using
+the image URL to using the foreign image UUID.
+"""
+import logging
+import re
+from textwrap import dedent
+
+from airflow.decorators import dag
+from airflow.models.abstractoperator import AbstractOperator
+from airflow.operators.python import PythonOperator
+from psycopg2.errors import UniqueViolation
+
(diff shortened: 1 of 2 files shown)


openverse_catalog/dags/maintenance/fix_phylopic_foreign_identifier.py
```diff
@@ -0,0 +1,116 @@
...
+        postgres_conn_id=POSTGRES_CONN_ID,
+        default_statement_timeout=PostgresHook.get_execution_timeout(task),
+    )
+    phylopic_rows = pg.get_records(
+        "SELECT foreign_identifier, identifier FROM image WHERE provider = 'phylopic';"
+    )
+
+    uuid_pattern = (
+        "[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[089ab][0-9a-f]{3}-[0-9a-f]{12}"
+    )
```
AetherUnbound, line 37: I think we could compile this pattern which might make the repeated regex a bit faster/more efficient: https://docs.python.org/3/library/re.html#re.compile

openverse_catalog/dags/maintenance/fix_phylopic_foreign_identifier.py
```diff
@@ -0,0 +1,116 @@
...
+            logger.error(
+                f"Failed to update foreign identifier for identifier {identifier}"
+                f" with error: {e}"
+            )
+
+    return counter
+
+
+def final_report(counter: dict[str, int]) -> None:
+    message = f"{DAG_ID} DAG run completed. Update statistics:\n{counter}."
```
AetherUnbound, line 84: Since this will be sent to Slack, it might be better to either explicitly print the statistics with keys & values _or_ to wrap `counter` in backticks:

```suggestion
    counter_message = "\n".join([f"{key.split('_', ' ').title()}: {value}" for key, value in counter.items()])
    message = f"{DAG_ID} DAG run completed. Update statistics:\n{counter_message}."
```

openverse_catalog/dags/maintenance/fix_phylopic_foreign_identifier.py
```diff
@@ -0,0 +1,116 @@
...
+    send_message(message, dag_id=DAG_ID)
+
+
+@dag(
+    dag_id=DAG_ID,
+    default_args={
+        **DAG_DEFAULT_ARGS,
+        "retries": 0,
+        # "execution_timeout": timedelta(hours=5),
```
AetherUnbound, line 93: Should this be set for prod?

openverse_catalog/dags/maintenance/fix_phylopic_foreign_identifier.py
```diff
@@ -0,0 +1,116 @@
...
+    schedule=None,
+    catchup=False,
+    doc_md=__doc__,
+    tags=["data_normalization"],
+)
+def update_phylopic():
+    update = PythonOperator(
+        task_id="update_foreign_identifiers",
+        python_callable=update_foreign_identifiers,
+    )
```
AetherUnbound, line 104: If `update_foreign_identifiers` and `final_report` are both set up as [`@tasks`](https://airflow.apache.org/docs/apache-airflow/stable/tutorial/taskflow.html#tasks), the XCOM can be passed a bit easier! Something like this:

```python
final_report(update_foreign_identifiers())
```

AetherUnbound (approved):
This is fantastic! Thanks for taking the time to write this so we could maintain continuity across API versions 🙌🏼 I tried the testing steps, and then ran the DAG from #1060 and confirmed that the new total count is 7k rather than 10k 🥳 

```sql
-- Before 

openledger> select count(*) from image;
+-------+
| count |
|-------|
| 4156  |
+-------+
SELECT 1
Time: 0.016s

-- After

openledger> select count(*) from image;
+-------+
| count |
|-------|
| 7055  |
+-------+
SELECT 1
Time: 0.012s
```

openverse_catalog/dags/maintenance/fix_phylopic_foreign_identifier.py
krysal: I tried that but even though the data type says it's a `dict`, for some reason at that point `counter` is a `str` 🤷🏻‍♀️ I'll better remove that type hint to avoid confusion.

openverse_catalog/dags/maintenance/fix_phylopic_foreign_identifier.py
krysal: I'm not familiar with this syntax but looks interesting. Where should this go?

stacimc (approved):
Lovely! Tested great for me -- very cool testing instructions :)

openverse_catalog/dags/maintenance/fix_phylopic_foreign_identifier.py
AetherUnbound: Oh, it's probably because we're missing this arg:
https://github.com/WordPress/openverse-catalog/blob/3f2ab0967991a45bb82038e05024d7b82fe06ae2/openverse_catalog/dags/database/terminate_long_queries_workflow.py#L75

openverse_catalog/dags/maintenance/fix_phylopic_foreign_identifier.py
AetherUnbound: We don't have a ton of DAGs that use the TaskFlow API yet, in fact I think this may be one of the few examples we have: https://github.com/WordPress/openverse-catalog/pull/608/files

But it's certainly something I'd like to start using more!

PR #1086, opened 2023-04-03
Check required fields in media store

### LINKED ISSUE #1019: Metropolitan receives records with `None` url
## Description

Several ingestion days during a recent run of `metropolitan_museum_reingestion_workflow` have raised the following error:

```
  File "/opt/airflow/openverse_catalog/dags/providers/provider_api_scripts/provider_data_ingester.py", line 423, in process_batch
    store.add_item(**record)
  File "/opt/airflow/openverse_catalog/dags/common/storage/image.py", line 145, in add_item
    image = self._get_image(**image_data)
  File "/opt/airflow/openverse_catalog/dags/common/storage/image.py", line 152, in _get_image
    image_metadata = self.clean_media_metadata(**kwargs)
  File "/opt/airflow/openverse_catalog/dags/common/storage/media.py", line 142, in clean_media_metadata
    media_data["filetype"] = self._validate_filetype(
  File "/opt/airflow/openverse_catalog/dags/common/storage/media.py", line 312, in _validate_filetype
    filetype = extract_filetype(url, self.media_type)
  File "/opt/airflow/openverse_catalog/dags/common/extensions.py", line 9, in extract_filetype
    possible_filetype = url.split(".")[-1]
AttributeError: 'NoneType' object has no attribute 'split'
```

It looks like somehow we're attempting to process a record with a null `image_url`. We should add a check for this.

## Reproduction

Unfortunately in each of the production instances, the error happens > 5 hrs into the DagRun. Metropolitan only has a single batch which just gets a list of object IDs, and the bulk of the work is done in `get_record_data`, which makes a request for each ID. This means the current logging setup doesn't allow us a way to quickly reproduce the issue, other than doing a full run for the problematic dates. Use DAG config:

```
{
"initial_query_params":{"metadataDate":"2022-06-18"}
}
```

stacimc:
## Fixes

Fixes #1019 by @stacimc 

## Description

This PR:

* raises a `ValueError` in the MediaStore's `clean_media_metadata` if any of the required fields `foreign_landing_url`, `url`, and `foreign_identifier` are missing.
* Adds some new error handling for Metropolitan:
  * Returns early if no image urls (primary or additional) can be pulled
  * Handles missing `isPublicDomain` key (in addition to explicit `False`)
  * Returns early if missing foreign landing url

## Testing Instructions

View the new tests.

The Metropolitan issue is hard to reproduce organically. You can manually update the metropolitan DAG code to add a `None` image [here](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py#L93):

```
        image_list = [main_image] + other_images + [None]
```

Then run the DAG and ensure that you do not see any errors (because no records with `None` url are passed to the MediaStore).

You can test that the error handling in the media store is working by hard-coding `None` for one of the required fields in the results [here](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py#L110). Then you should see the task fail when you run the DAG, with an error message like:

```
Record missing required field: `foreign_landing_url`
```

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/common/storage/media.py
+++ b/openverse_catalog/dags/common/storage/media.py
@@ -111,13 +111,24 @@ class MediaStore(metaclass=abc.ABCMeta):
         - add `provider`,
         - add default `category`, if available.
 
-        Returns None if license is invalid
+        Returns None if license is invalid, or if missing required `foreign_identifier`,
+        `foreign_landing_url`, or `url`.
         """
         if media_data["license_info"].license is None or not is_valid_license_info(
             media_data["license_info"]
         ):
             logger.debug("Discarding media due to invalid license")
             return None
+
+        for field in [
+            "foreign_identifier",
+            "foreign_landing_url",
+            f"{self.media_type}_url",
+        ]:
+            if media_data.get(field) is None:
+                logger.debug(f"Discarding media due to missing {field}")
+                return None
+
         for field in [
             f"{self.media_type}_url",
             "foreign_landing_url",
--- a/openverse_catalog/dags/common/storage/media.py
+++ b/openverse_catalog/dags/common/storage/media.py
@@ -111,8 +111,8 @@ class MediaStore(metaclass=abc.ABCMeta):
         - add `provider`,
         - add default `category`, if available.
 
-        Returns None if license is invalid, or if missing required `foreign_identifier`,
-        `foreign_landing_url`, or `url`.
+        Returns None if license is invalid. Raises an error if missing required
+        `foreign_identifier`, `foreign_landing_url`, or `url`.
         """
         if media_data["license_info"].license is None or not is_valid_license_info(
             media_data["license_info"]
@@ -126,8 +126,7 @@ class MediaStore(metaclass=abc.ABCMeta):
             f"{self.media_type}_url",
         ]:
             if media_data.get(field) is None:
-                logger.debug(f"Discarding media due to missing {field}")
-                return None
+                raise ValueError(f"Record missing required field: `{field}`")
 
         for field in [
             f"{self.media_type}_url",
--- a/openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
@@ -82,15 +82,21 @@ class MetMuseumDataIngester(ProviderDataIngester):
             object_endpoint, self.retries
         )
 
-        if object_json.get("isPublicDomain") is False:
+        if not object_json.get("isPublicDomain"):
             self.non_cc0_objects += 1
             if self.non_cc0_objects % self.batch_limit == 0:
                 logger.info(f"Retrieved {self.non_cc0_objects} non-CC0 objects.")
             return None
 
+        if (foreign_landing_url := object_json.get("objectURL")) is None:
+            return None
+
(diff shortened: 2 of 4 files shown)


openverse_catalog/dags/common/storage/media.py
```diff
@@ -111,13 +111,24 @@ def clean_media_metadata(self, **media_data) -> dict | None:
...
             return None
+
+        for field in [
+            "foreign_identifier",
+            "foreign_landing_url",
+            f"{self.media_type}_url",
+        ]:
+            if media_data.get(field) is None:
+                logger.debug(f"Discarding media due to missing {field}")
+                return None
```
krysal, line 130: I can't believe this wasn't checked before 😱 Well noticed. However, I wonder whether just ignoring items when missing fields at this point, in the media store class, is the right approach. I am slightly inclined to raise an exception here and have each provider DAG handle theses cases explicitly. Sometimes, if a field is not found another from the source can be used in replacement, e.g. this is common in the `creator_url`. These errors have also helped me realize some gaps in the logic in a few cases!

What do you think?

krysal (commented):
Following my previous comment, shouldn't we also add the field validations in the Metropolitan DAG? It could skip processing for more than one image, cutting times a bit.

openverse_catalog/dags/common/storage/media.py
stacimc: > Sometimes, if a field is not found another from the source can be used in replacement, e.g. this is common in the creator_url

That's a really great point -- raising an exception here might help us spot these opportunities in some cases, and even in cases where the record should ultimately be discarded, it'll be an easy fix in the provider class. I'll update!

openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
```diff
@@ -82,15 +82,21 @@ def get_record_data(self, object_id):
...
                 logger.info(f"Retrieved {self.non_cc0_objects} non-CC0 objects.")
             return None
 
+        if (foreign_landing_url := object_json.get("objectURL")) is None:
+            return None
+
         main_image = object_json.get("primaryImage")
         other_images = object_json.get("additionalImages", [])
-        image_list = [main_image] + other_images
+        image_list = [img for img in ([main_image] + other_images) if img is not None]
```
AetherUnbound, line 96: Nit, this could be:
```suggestion
        image_list = [img for img in [main_image, *other_images] if img is not None]
```

AetherUnbound (changes requested):
I like what y'all have proposed & discussed! Given how rapidly we are able to iterate on ingestion DAGs, I like the idea of failing sooner and addressing issues on a per-provider basis. That, in its own way, can serve as a data health check - if the data changes and our assumptions no longer hold, we want to know about it.

I have one logging change which is blocking, but otherwise this looks great!

AetherUnbound (approved):
LGTM! 😄

zackkrida (approved):
Looks great!

PR #1093, opened 2023-04-06
Retire image expiration dag

AshwanthramKL:
Solves Issue: #999


openverse_catalog/dags/common/loader/sql.py
```diff
@@ -372,55 +372,3 @@ def _delete_malformed_row_in_file(tsv_file_name, line_number):
                 write_obj.write(line)
```
obulat, line 374:
```suggestion

```

obulat (approved):
Thank you for your contribution, @AshwanthramKL!

PR #1055, opened 2023-03-20
Do not show a load completion message if all steps failed

rsubra13:
### Description:

Fixing https://github.com/WordPress/openverse-catalog/issues/465

Idea is not to send a slack notification when all steps failed. This can be found by checking the `duration` and `record_counts_by_media_type`. Added extra tests, added some good to have formatting/performance changes on the go as well. 

> Tests

Screenshot 2023-03-20 at 4 10 52 PM

PR #1094, opened 2023-04-07
Restructure codebase to be more monorepo-friendly

dhruvkb:
This PR prepares the `WordPress/openverse-catalog` repository to be merged into the `WordPress/openverse` repository.

Making it a PR helps verify that everything continues to work (like tests and CI) and also make the changes easier to review.


dhruvkb:
For some reason, instead of running the tests, the CI is just printing a help message: https://github.com/WordPress/openverse-catalog/actions/runs/4635136220/jobs/8201959249#step:11:26

docker-compose.yml
```diff
@@ -1,31 +1,136 @@
...
+  restart: on-failure
+  depends_on:
+    - postgres
+    - s3
+  image: openverse_catalog
+  env_file:
+    - .env
+    - catalog/.env
+  build:
+    context: ./catalog/
```
AetherUnbound, line 17: All of the other images are under `docker/<service>/`, should this be as well? Or is that in a separate folder so the `catalog/docker-compose.yml` file will still work? And if it's the latter, can we add a comment somewhere about that difference?

catalog/docker-compose.yml
AetherUnbound, line 1: Ah, I thought this had gone away! So we'll need to update the deployment to point to this file instead, right?

https://github.com/WordPress/openverse-infrastructure/blob/main/modules/services/catalog-airflow/init.tpl#L74-L76

catalog/justfile
```diff
@@ -0,0 +1,131 @@
...
+#########
+# Tests #
+#########
+
+# Run a command in a test container under `SERVICE`
+_mount-test command: up-deps
+    just ../run \
+        --rm \
+        -e AIRFLOW_VAR_INGESTION_LIMIT=1000000 \
+        --volume {{ justfile_directory() }}/../docker:/opt/airflow/docker/ \
```
AetherUnbound, line 95: This is awesome! I love that you were able to set this up so test are mounted automatically in dev 💯

justfile
```diff
@@ -1,145 +1,149 @@
...
-      if [ $? -ne 0 ]; then
-          printf "\n\n\e[31m!! Changes found in DAG documentation, please run 'just generate-dag-docs' locally and commit difference !!\n\n"
-          exit 1
-      fi
-    fi
-
 # Generate files for a new provider
 add-provider provider_name endpoint +media_types="image":
-    python3 openverse_catalog/templates/create_provider_ingester.py "{{ provider_name }}" "{{ endpoint }}" -m {{ media_types }}
+    python3 catalog/templates/create_provider_ingester.py "{{ provider_name }}" "{{ endpoint }}" -m {{ media_types }}
```
AetherUnbound, line 149: This is a catalog-specific recipe, I think it should probably be under `catalog/justfile`?

catalog/pytest.ini
AetherUnbound, line 1: It looks like this file isn't getting picked up appropriately when running tests. I was tipped off by the fact that we were 1) seeing a flaky test report and 2) not automatically using `pytest-xdist` to run the tests multithreaded. It looks like if `pytest` is run from the `catalog/` directory within the container, then it picks up the `pytest.ini` file correctly. Although running it that way gives me a few other warnings:

```
../../../home/airflow/.local/lib/python3.10/site-packages/_pytest/cacheprovider.py:432
  /home/airflow/.local/lib/python3.10/site-packages/_pytest/cacheprovider.py:432: PytestCacheWarning: could not create cache path /opt/airflow/catalog/.pytest_cache/v/cache/nodeids
    config.cache.set("cache/nodeids", sorted(self.cached_nodeids))

../../../home/airflow/.local/lib/python3.10/site-packages/_pytest/stepwise.py:56
  /home/airflow/.local/lib/python3.10/site-packages/_pytest/stepwise.py:56: PytestCacheWarning: could not create cache path /opt/airflow/catalog/.pytest_cache/v/cache/stepwise
    session.config.cache.set(STEPWISE_CACHE_DIR, [])

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

justfile
AetherUnbound, line 149: Additionally, since the tests themselves have moved, I think this script will need to be updated to generate files in `catalog/tests` rather than `tests/`.

catalog/pytest.ini
AetherUnbound: Follow up, when I try to run _all_ tests (`j catalog/test-session` -> `cd catalog` -> `pytest --extended`) I get a number of test failures, I think based on some modules being moved.

AetherUnbound (changes requested):
@dhruvkb this is a _hugely_ impressive effort. I am so amazed and impressed that you were able to conform this to the standard our other projects have while maintaining existing functionality!

For other folks testing this repo who have already used the old setup, I was able to use `just env` and then `just recreate` to get everything up and running. I even ran a few DAGs and things looked good!

I have some notes about testing though, it looks like some of the config isn't being set up quite right, we'll need to figure that out before this can be merged.

docker-compose.yml
dhruvkb: `Dockerfile`s for our core services like API, frontend and catalog are kept with the code folders, and those for services these depend on (like Postgres etc) are kept inside the `docker/` folder.

The `docker-compose.yml` file in the root of the repo is the development one (includes all services grouped by profiles) and each stack that uses Compose for deployment also has a `docker-compose.yml` file inside the codebase (like the `ingestion_server/` in the monorepo).

catalog/docker-compose.yml
dhruvkb: Compared to the ingestion_server and indexer_workers, the change to the infrastructure for the catalog will be a bit more contrived. The `init.tpl` for those services doesn't seem to be cloning the repo and so it didn't matter that much when we moved things around.

justfile
dhruvkb: Yes, I missed this because I didn't know what it is. I was moving them only after understanding each's purpose. I'll move it inside `catalog/` too.

catalog/pytest.ini
dhruvkb: Could you look into those and update this PR? I had a fairly hard time understanding what was failing and why, which was made worse because they were failing locally but not on CI (which is supposed to be the other way round!).

AetherUnbound: You bet, I can do that 🙂

justfile
stacimc: 👍 This should be fixed by updating [this line](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/templates/create_provider_ingester.py#L92) in `create_provider_ingester.py`:

```
tests_path = project_path / "tests"
```

stacimc (commented):
This is tremendous, @dhruvkb ✨ I recreated the catalog locally and ran through the `just` commands and a bunch of DAGs, including various categories of provider scripts and our data refreshes. I'll keep playing with it, but I am not seeing any additional issues other than what has been mentioned. Incredible work! 👏

AetherUnbound:
@dhruvkb I believe I've addressed the issues with the tests! We should now see the `--extended` flag respected on CI as well, which we can check by making sure not tests are skipped. I want to point out, specifically, the change I made in c833356b772662d24e03655e07bf264ba32474d5 - after _lots_ of debugging, I noticed that the quotes around `bash -c 'pytest {{ args {{'` were actually disappearing when the `docker-compose run` command was finally called. Here's an example before the change:

```
j catalog/test --extended
just _mount-test "bash -c \'pytest --extended\'"
env COMPOSE_PROFILES="catalog_dependencies" just ../up 
env COMPOSE_PROFILES="catalog_dependencies" docker-compose -f docker-compose.yml up -d
Recreating openverse-catalog_s3_1 ... 
Recreating openverse-catalog_s3_1 ... done
Recreating openverse-catalog_load_to_s3_1 ... done
just ../run --rm -e AIRFLOW_VAR_INGESTION_LIMIT=1000000 -w /opt/airflow/catalog --volume /home/aether/git-a8c/openverse-catalog/catalog/../docker:/opt/airflow/docker/ scheduler bash -c \'pytest --extended\'
just dc run -u airflow  --rm -e AIRFLOW_VAR_INGESTION_LIMIT=1000000 -w /opt/airflow/catalog --volume /home/aether/git-a8c/openverse-catalog/catalog/../docker:/opt/airflow/docker/ scheduler bash -c 'pytest --extended'
env COMPOSE_PROFILES="api,ingestion_server,frontend,catalog" docker-compose -f docker-compose.yml run -u airflow --rm -e AIRFLOW_VAR_INGESTION_LIMIT=1000000 -w /opt/airflow/catalog --volume /home/aether/git-a8c/openverse-catalog/catalog/../docker:/opt/airflow/docker/ scheduler bash -c pytest --extended
```

Note that the last line shows as `bash -c pytest --extended`, meaning `--extended` (and any other args) get handed to _bash_ rather than the `pytest` command. Here's what that same command looks like after my change:

```
j catalog/test --extended
just _mount-test "bash -c \'pytest --extended\'"
env COMPOSE_PROFILES="catalog_dependencies" just ../up 
env COMPOSE_PROFILES="catalog_dependencies" docker-compose -f docker-compose.yml up -d
openverse-catalog_postgres_1 is up-to-date
openverse-catalog_s3_1 is up-to-date
Starting openverse-catalog_load_to_s3_1 ... done
just ../run --rm -e AIRFLOW_VAR_INGESTION_LIMIT=1000000 -w /opt/airflow/catalog --volume /home/aether/git-a8c/openverse-catalog/catalog/../docker:/opt/airflow/docker/ scheduler bash -c \'pytest --extended\'
just dc run -u airflow  "--rm -e AIRFLOW_VAR_INGESTION_LIMIT=1000000 -w /opt/airflow/catalog --volume /home/aether/git-a8c/openverse-catalog/catalog/../docker:/opt/airflow/docker/ scheduler bash -c 'pytest --extended'"
env COMPOSE_PROFILES="api,ingestion_server,frontend,catalog" docker-compose -f docker-compose.yml run -u airflow --rm -e AIRFLOW_VAR_INGESTION_LIMIT=1000000 -w /opt/airflow/catalog --volume /home/aether/git-a8c/openverse-catalog/catalog/../docker:/opt/airflow/docker/ scheduler bash -c 'pytest --extended'
```

The quotes around `'pytest --extended'` now ensure that all extra CLI args handed at the very top make their way down to pytest. Very odd behavior! I want to take a look at the cache warnings too, so I might push one more commit.

AetherUnbound:
Okay I think I've got it rectified 😄

dhruvkb:
Thank you @AetherUnbound for fixing nearly every shortcoming left. Also both you and @stacimc for the super-fast turnaround on the reviews. 

Should we merge this into `main` or only consider this branch as the base for the monorepo merge and leave `main` unaffected? I'm inclined towards the latter so that our prod and infra can work unchanged while we move the code around.

AetherUnbound:
I'm just realizing that we have a few community PRs (#999, #1055, #1059) that we should probably merge before we merge this one 😅 I'd be okay merging this into `main` after we merge those and update this PR! What we could do is disable the dag sync cron job on production, at least until we update the infrastructure repo and deploy a new version of the catalog. We're also waiting on the image data refresh right now which currently has a 3 week timeout, so there's not really a rush to deploy a new version - thus it might make more sense to be working off of `main` while we're at it. What do you think?

dhruvkb:
I agree with all your points.

- We should merge the community PRs before this, then update this PR with those changes.
- If merging to `main` doesn't cause any breakages, I'm all for it. It'll probably take one week to fix all the usages with references to the monorepo so if there is no need to deploy in that time, we can merge.

AetherUnbound:
Excellent! I will shut the dag-sync cron off once we've merged all the community PRs 🙂

AetherUnbound (approved):
Approving so this is not blocked when time comes to merge, beyond that it's ready to go!

stacimc (approved):
All of my local testing has gone great! Excited for this 🚀

justfile
```diff
@@ -1,145 +1,145 @@
...
+# Meaning of Just prefixes:
+# @ - Quiet recipes (https://github.com/casey/just#quiet-recipes)
+# _ - Private recipes (https://github.com/casey/just#private-recipes)
 
+# The monorepo uses `PROD` instead of `IS_PROD`.
 IS_PROD := env_var_or_default("IS_PROD", "")
-DOCKER_FILES := "--file=docker-compose.yml" + (
-    if IS_PROD != "true" {" --file=docker-compose.override.yml"} else {""}
-)
-SERVICE := env_var_or_default("SERVICE", "scheduler")
```
AetherUnbound, line 11: It looks like we're getting rid of the `SERVICE` default variable for some recipes (namely `exec` and `run`). Do you mind updating any references to `just exec` or `just run` in the `README.md` to include the `scheduler` service? E.g. `just run scheduler airflow dags list`

AetherUnbound (approved):
I tested all of the commands and a couple of DAGs after the rebase - everything looks good! I only have one comment but after that this can be merged 🥳

justfile
dhruvkb: Updated in 24a858e.

PR #1098, opened 2023-04-10
Refactor LicenseInfo

### LINKED ISSUE #1097: LicenseInfo should only hold a valid license information
## Problem

#66 aimed to validate the license information in the provider API scripts so that the `MediaStore` only receives the items with valid licenses. However, the implementation allowed for `None` values to be in `LicenseInfo`, which made the validation more difficult than necessary. To see if the item's `license_info` is valid, you need to see if its `.license` property is `None` or not.

## Description

We can refactor the `get_license_info` function to return a `LicenseInfo` tuple for a valid license, or `None`. Then, in the provider API scripts, we can throw out the items with `None` `LicenseInfo`. Since `LicenseInfo` is a required parameter in `add_item`, and it will only contain valid license information, we will not have to validate the license again in the `MediaStore`.

obulat:
## Fixes
Fixes #1097 by @obulat 

## Description
This PR makes `LicenseInfo` only hold a valid license information, instead of allowing it to have `None` values. Now `get_license_info` returns a valid `LicenseInfo` if data is valid, otherwise, it returns None.

This means that we pass a valid `LicenseInfo` to the `MediaStore`, and no longer need to call `is_valid_license_info` in the `MediaStore` to validate it.

### DIFF
--- a/openverse_catalog/dags/common/licenses/__init__.py
+++ b/openverse_catalog/dags/common/licenses/__init__.py
@@ -2,7 +2,6 @@ from .licenses import (  # noqa: F401
     LicenseInfo,
     get_license_info,
     get_license_info_from_license_pair,
-    is_valid_license_info,
 )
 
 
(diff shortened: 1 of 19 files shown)


AetherUnbound:
We're going to close this for now while we work on the monorepo migration effort - we'll re-open when that work is complete!

PR #1102, opened 2023-04-11
Handle empty string urls for Metropolitan

### LINKED ISSUE #1019: Metropolitan receives records with `None` url

stacimc:
## Fixes

Fixes #1019 by @stacimc 

## Description

Previously we added handling to the Metropolitan DAG for when we receive urls that are `None`. It turns out we also sometimes receive empty strings.

This PR discards records with empty string urls within Metropolitan. We should consider also handling this in the media store; currently the `validate_url_string` method used by the media store returns `None` when it cannot validate a url. We _could_ instead raise an error here. I did not do so in this PR because I did not want to make changes that apply across all DAGs before the catalog code freeze, and the fix will still need to be made in Metropolitan regardless. If folks think this change would be a good idea, I can create an issue for it.

## Testing Instructions

This ordinarily takes a very long time to reproduce, but good news, I logged the id of a 'bad' record locally. You can update  `get_batch_data` to return just the bad id. Change [this line](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py#L74) to:

```
return [4594,]
```

Run the DAG on the `main` branch and you should see the error:

```
[2023-04-11, 21:19:45 UTC] {taskinstance.py:1776} ERROR - Task failed with exception
provider_data_ingester.IngestionError: 'NoneType' object has no attribute 'split'
```

With the same change, run the DAG on this branch and the DAG should discard that record and pass.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
+++ b/openverse_catalog/dags/providers/provider_api_scripts/metropolitan_museum.py
@@ -88,12 +88,12 @@ class MetMuseumDataIngester(ProviderDataIngester):
                 logger.info(f"Retrieved {self.non_cc0_objects} non-CC0 objects.")
             return None
 
-        if (foreign_landing_url := object_json.get("objectURL")) is None:
+        if not (foreign_landing_url := object_json.get("objectURL")):
             return None
 
         main_image = object_json.get("primaryImage")
         other_images = object_json.get("additionalImages", [])
-        image_list = [img for img in [main_image, *other_images] if img is not None]
+        image_list = [img for img in [main_image, *other_images] if img]
         if not image_list:
             logger.info(f"No images detected for {object_id=}.")
             return None
--- a/tests/dags/providers/provider_api_scripts/test_metropolitan_museum.py
+++ b/tests/dags/providers/provider_api_scripts/test_metropolitan_museum.py
@@ -196,9 +196,21 @@ def test_get_record_data_with_non_ok():
             id="missing_foreign_landing_url",
         ),
         pytest.param(
+            json.loads('{"isPublicDomain": false, "objectURL": ""}'),
+            None,
+            id="empty_string_for_foreign_landing_url",
+        ),
+        pytest.param(
             json.loads('{"isPublicDomain": true, "objectURL": "test.com"}'),
             None,
             id="missing_images",
+        ),
+        pytest.param(
+            json.loads(
+                '{"isPublicDomain": true, "objectURL": "test.com", "primaryImage": ""}'
+            ),
+            None,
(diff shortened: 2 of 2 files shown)


AetherUnbound (approved):
Change looks good, and I used the instructions provided to verify that this discards the record!

obulat (approved):
Changes look good and test well. I agree with the scope of this PR, and the idea of improving validation after the monorepo move.

I was thinking about validation recently, and think that we can split the validation step from the string preparation step. In the `add_item`, we could use a dataclass with `post_init` (or `attrs`) to only add the media where all properties are valid (no `None`s, no overflowing length if it's not allowed), and then in the saving step we would truncate or stringify in the way that is described in the `Column` classes.

